PR #26, opened 2017-03-06
Changes circle to build with architect

JosephSalisbury:


JosephSalisbury:
Note: not adding deployment support yet, as we only want this on aws (not leaseweb)

PR #29, opened 2017-03-06
Changes circle to fetch architect unauthed if token not present

JosephSalisbury:
We usually fetch architect's latest binary authenticated, to
avoid any issues with rate limiting.
However, we don't pass secrets to forked builds, for security
reasons, so forked builds can't fetch the architect binary at all.

This changeset makes the fetch fall back to unauthenticated if the
token is not present - this means that forked builds work, but may
be subjected to GitHub rate limiting.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,11 @@ machine:
 dependencies:
   override:
     - |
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
@nhlfr can you try this in your forked repo? i think it should work

nhlfr:
It should work, let's merge it.

PR #30, opened 2017-03-07
Remove RELEASE_TOKEN usage from CircleCI config

nhlfr:
Both aws-operator and architect are open source projects, so there is no need to use it. Also, it doesn't seem to work when pull requests are created from forks.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,7 @@ machine:
 dependencies:
   override:
     - |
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
My thinking for keeping it with a conditional was so that master builds can still authenticate. Not certain if it's _necessary_.

PR #38, opened 2017-03-13
create: make instance creation more robust

iaguis:
* Wait until the instance profile exists before continuing
* Retry instance creation because Roles take some time to propagate to
instance profiles


nhlfr:
Sorry for merging it myself, but I urgently need this.

PR #39, opened 2017-03-13
service: rework roles/policies/instance profiles

iaguis:
We now name roles, policies and instance profiles in a per-cluster
basis.

Also, we treat already-existing roles or instance profiles as errors.
Reusing them doesn't really work. In any case, they should've been
removed when we deleted the cluster (for now we need to do it manually
since the delete operator is not implemented yet).

Also, use same name for instance profile and role so we can delete both when deleting the role from the aws console:

"You cannot use the console to delete an instance profile, except when
it has the exact same name as the role and you delete it as part of the
process of deleting a role as described in the preceding procedure."

http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_manage_delete.html#roles-managingrole-deleting-console

### DIFF
--- a/service/create/policy.go
+++ b/service/create/policy.go
@@ -11,7 +11,7 @@ import (
 const (
 	RoleName                 = "EC2-K8S-Role"
 	PolicyName               = "EC2-K8S-Policy"
-	ProfileName              = "EC2-DecryptTLSCerts"
+	ProfileName              = "EC2-K8S-Role"
 	AssumeRolePolicyDocument = `{
 		"Version": "2012-10-17",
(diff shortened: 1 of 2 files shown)


iaguis:
> Sorry for blocking this, but cannot we add cluster names here? I think that multiple clusters running in the same region shouldn't use the same roles/policies.

Yes, next change is gonna be changing the name of these to `watheverTemplate` and modify the code to add clusterID before creating them.

nhlfr:
needs rebase

iaguis:
I'll wait for #42

PR #63, opened 2017-03-17
rebuild bindata after downgrading to go 1.7

asymmetric:
go-bindata built on 1.7 and 1.8 produces different outputs. since we use
1.7 across the board, downgrade and rebuild bindata.


asymmetric:
This is unnecessary, since we [will remove `bindata.go`](https://github.com/giantswarm/aws-operator/pull/58).

PR #77, opened 2017-03-22
Use organisation parameter in architect

nhlfr:


nhlfr:
This PR fixes errors like the following https://circleci.com/gh/kinvolk/aws-operator/3?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

iaguis:
To explain it a bit more, we need this to build forks on CircleCI because they expect the giantswarm organization in go packages and not, for example, kinvolk.

JosephSalisbury:
I'll have a think about how we can sort this without editing the circle.yml, but let's go with this for now

PR #89, opened 2017-04-04
Updates configmap for incluster deployment

JosephSalisbury:


JosephSalisbury:
I'll do the aws keys via secret (manually), then we should be good

JosephSalisbury:
Yeah for sure, I'll follow up

PR #90, opened 2017-04-04
create/resources: Use VPC for network separation of clusters

### LINKED ISSUE #27: Set up VPC network for the cluster
We need connectivity for our cluster.

As discussed OOB, we will create a VPC per cluster, which will substitute the flannel part on the KVM operator.

We need to implement this on the operator.

nhlfr:
In order to separate Kubernetes cluster in AWS, we will run each cluster in the separate VPC, with the two submets:
- public - for master nodes
- private - for worker nodes

We also set up security group in order to have control on which ports we want to expose in each machine.

Fixes #27

### DIFF
--- a/resources/aws/instance.go
+++ b/resources/aws/instance.go
@@ -11,12 +11,6 @@ import (
 	awsutil "github.com/giantswarm/aws-operator/client/aws"
 )
 
-const (
-	// EC2 instance tag keys.
-	tagKeyName    string = "Name"
-	tagKeyCluster string = "Cluster"
-)
-
 type EC2StateCode int
 
 const (
@@ -42,6 +36,8 @@ type Instance struct {
 	SmallCloudconfig       string
 	IamInstanceProfileName string
 	PlacementAZ            string
+	SecurityGroupID        string
+	SubnetID               string
 	id                     string
 	AWSEntity
 }
@@ -118,6 +114,10 @@ func (i *Instance) CreateOrFail() error {
 			Placement: &ec2.Placement{
 				AvailabilityZone: aws.String(i.PlacementAZ),
 			},
+			SecurityGroupIds: []*string{
+				aws.String(i.SecurityGroupID),
+			},
+			SubnetId: aws.String(i.SubnetID),
 		})
 		if err != nil {
 			return microerror.MaskAny(err)
@@ -154,6 +154,14 @@ func (i *Instance) CreateOrFail() error {
 
 func (i *Instance) Delete() error {
 	if _, err := i.Clients.EC2.TerminateInstances(&ec2.TerminateInstancesInput{
+		InstanceIds: []*string{
+			aws.String(i.id),
+		},
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	if err := i.Clients.EC2.WaitUntilInstanceTerminated(&ec2.DescribeInstancesInput{
 		InstanceIds: []*string{
 			aws.String(i.id),
 		},
--- a/resources/aws/routetable.go
+++ b/resources/aws/routetable.go
@@ -1,0 +1,114 @@
+package aws
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+type RouteTable struct {
+	Name  string
+	VpcID string
+	id    string
+	AWSEntity
+}
+
+func (r RouteTable) findExisting() (*ec2.RouteTable, error) {
+	routeTables, err := r.Clients.EC2.DescribeRouteTables(&ec2.DescribeRouteTablesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name: aws.String(fmt.Sprintf("tag:%s", tagKeyName)),
+				Values: []*string{
+					aws.String(r.Name),
+				},
+			},
+		},
+	})
+	if err != nil {
+		return nil, microerror.MaskAny(err)
+	}
+
+	if len(routeTables.RouteTables) < 1 {
+		return nil, microerror.MaskAny(routeTableFindError)
+	}
+
+	return routeTables.RouteTables[0], nil
+}
+
+func (r *RouteTable) checkIfExists() (bool, error) {
+	routeTable, err := r.findExisting()
+	if err != nil {
+		if strings.Contains(err.Error(), routeTableFindError.Error()) {
+			return false, nil
+		}
+		return false, microerror.MaskAny(err)
+	}
+
+	r.id = *routeTable.RouteTableId
+
+	return true, nil
+}
+
+func (r *RouteTable) CreateIfNotExists() (bool, error) {
+	exists, err := r.checkIfExists()
+	if err != nil {
+		return false, microerror.MaskAny(err)
+	}
+
+	if exists {
+		return false, nil
+	}
+
+	if err := r.CreateOrFail(); err != nil {
+		return false, microerror.MaskAny(err)
+	}
+
+	return true, nil
+}
+
+func (r *RouteTable) CreateOrFail() error {
+	routeTable, err := r.Clients.EC2.CreateRouteTable(&ec2.CreateRouteTableInput{
+		VpcId: aws.String(r.VpcID),
+	})
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	if _, err := r.Clients.EC2.CreateTags(&ec2.CreateTagsInput{
+		Resources: []*string{routeTable.RouteTable.RouteTableId},
+		Tags: []*ec2.Tag{
+			{
+				Key:   aws.String(tagKeyName),
+				Value: aws.String(r.Name),
+			},
+		},
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	r.id = *routeTable.RouteTable.RouteTableId
+
+	return nil
+}
+
+func (r *RouteTable) Delete() error {
+	routeTable, err := r.findExisting()
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	if _, err := r.Clients.EC2.DeleteRouteTable(&ec2.DeleteRouteTableInput{
+		RouteTableId: routeTable.RouteTableId,
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
+
+func (r RouteTable) ID() string {
+	return r.id
+}
--- a/resources/aws/subnet.go
+++ b/resources/aws/subnet.go
@@ -1,0 +1,141 @@
+package aws
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+
+	awsclient "github.com/giantswarm/aws-operator/client/aws"
+)
+
+type Subnet struct {
+	AvailabilityZone string
+	CidrBlock        string
+	Name             string
+	VpcID            string
+	id               string
+	AWSEntity
+}
+
+func (s Subnet) findExisting() (*ec2.Subnet, error) {
+	subnets, err := s.Clients.EC2.DescribeSubnets(&ec2.DescribeSubnetsInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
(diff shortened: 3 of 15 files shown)


resources/aws/vpc.go
```diff
@@ -0,0 +1,117 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+	// "github.com/giantswarm/aws-operator/resources"
```
JosephSalisbury: Why is this commented?

resources/aws/vpc.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+const (
+	cidrBlock         = "10.0.0.0/16"
```
JosephSalisbury: Where does this come from?

service/create/service.go
```diff
@@ -42,6 +42,11 @@ const (
 	// Prefixes used for machine names.
 	prefixMaster string = "master"
 	prefixWorker string = "worker"
+	// SSH port to open in security group
+	portSSH int = 22
+	// Subnet CIDRs
+	privateSubnetCidr = "10.0.0.0/19"
```
JosephSalisbury: These are in two places, can we refactor them into one constant?

service/create/service.go
```diff
@@ -42,6 +42,11 @@ const (
...
+	publicSubnetCidr  = "10.0.128.0/20"
```
JosephSalisbury: This is also replicated, I can feel this being an issue :D

resources/aws/routetable.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+type RouteTable struct {
+	Name  string
+	VpcID string
+	id    string
```
JosephSalisbury: Should this be `ID`? I always forget golang foo around this

resources/aws/subnet.go
```diff
@@ -0,0 +1,116 @@
...
+	microerror "github.com/giantswarm/microkit/error"
+
+	awsclient "github.com/giantswarm/aws-operator/client/aws"
+)
+
+type Subnet struct {
+	AvailabilityZone string
+	CidrBlock        string
+	VpcID            string
+	id               string
```
JosephSalisbury: Same here - should this be `ID`?

JosephSalisbury: As, this is `id`, and the method is `ID`, right?

resources/aws/routetable.go
nhlfr: I want to keep that field private. The only places that modifiy the ID, should be `CreateOrFail` and `CreateIfNotExists` methods, then we should be able to only get it through the interface method.

resources/aws/subnet.go
nhlfr: Right :)

resources/aws/instance.go
```diff
@@ -42,6 +42,8 @@ type Instance struct {
 	SmallCloudconfig       string
 	IamInstanceProfileName string
 	PlacementAZ            string
+	SecurityGroupID        string
```
asymmetric: Do we want to support multiple Security Groups at this stage?

resources/aws/routetable.go
asymmetric: If it's an exported field then it should be ID, but in this case I think @nhlfr wanted to make this field private.

resources/aws/subnet.go
```diff
@@ -0,0 +1,116 @@
...
+	AWSEntity
+}
+
+func (s *Subnet) checkIfExists() (bool, error) {
+	subnets, err := s.Clients.EC2.DescribeSubnets(&ec2.DescribeSubnetsInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name: aws.String("availabilityZone"),
```
asymmetric: Should we have this as a constant in the same place where `tagKeyName` is defined?

resources/aws/subnet.go
```diff
@@ -0,0 +1,116 @@
...
+				Values: []*string{
+					aws.String(s.AvailabilityZone),
+				},
+			},
+			&ec2.Filter{
+				Name: aws.String("cidrBlock"),
```
asymmetric: Same, should be a constant.

resources/aws/subnet.go
```diff
@@ -0,0 +1,116 @@
...
+				Values: []*string{
+					aws.String(s.CidrBlock),
+				},
+			},
+			&ec2.Filter{
+				Name: aws.String("vpc-id"),
```
asymmetric: Ditto

resources/aws/subnet.go
```diff
@@ -0,0 +1,116 @@
...
+
+type Subnet struct {
+	AvailabilityZone string
+	CidrBlock        string
+	VpcID            string
+	id               string
+	AWSEntity
+}
+
+func (s *Subnet) checkIfExists() (bool, error) {
```
asymmetric: I sense a duplication in the force :) But this being Go...

service/create/service.go
```diff
@@ -278,13 +283,111 @@ func (s *Service) Boot() {
...
+						return
+					}
+					if vpcCreated {
+						s.logger.Log("info", fmt.Sprintf("created vpc for cluster '%s'", cluster.Name))
+					} else {
+						s.logger.Log("indo", fmt.Sprintf("vpc for cluster '%s' already exists, reusing", cluster.Name))
+					}
+
+					// Create security group
+					var securityGroup resources.ResourceWithID
```
asymmetric: What's the advantage of not declaring this as a `SecurityGroup`?

resources/aws/subnet.go
nhlfr: Exaclty, we cannot "template" that function in any way, so unfortunatelly, every `checkIfExists` and `CreateIfNotExists` functions will look very similar.

resources/aws/vpc.go
```diff
@@ -0,0 +1,117 @@
...
+		return microerror.MaskAny(err)
+	}
+
+	v.id = vpcID
+
+	return nil
+}
+
+func (v *VPC) Delete() error {
+	return microerror.MaskAny(notImplementedMethodError)
```
asymmetric: Why not?

nhlfr:
@asymmetric what's the name of your cluster? Looks like name overlap for me.

asymmetric:
lorenzo :)

service/create/service.go
```diff
@@ -278,13 +283,111 @@ func (s *Service) Boot() {
...
+					}
+					routeTableCreated, err := routeTable.CreateIfNotExists()
+					if err != nil {
+						s.logger.Log("error", fmt.Sprintf("could not create route table: %s", errgo.Details(err)))
+						return
+					}
+					if routeTableCreated {
+						s.logger.Log("info", "created route table")
+					} else {
+						s.logger.Log("info", "route rable already exists, reusing")
```
asymmetric: Typo

service/create/subnetnames.go
```diff
@@ -0,0 +1,11 @@
+package create
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/awstpr"
+)
+
+func (s *Service) subnetName(cluster awstpr.CustomObject, prefix string) string {
```
asymmetric: I think this could be a function.

resources/aws/vpc.go
nhlfr: For now we (unfortunately) hardcode it. I've just added TODO comment to move this to TPR later.

nhlfr:
@asymmetric @JosephSalisbury fixed the most of your comments. However, I didn't debug the @asymmetric's error yet, but feel free to comment the code at the same time.

resources/aws/vpc.go
asymmetric: Wouldn't it be better to have an issue instead? That seems to be the common practice so far.

resources/aws/routetable.go
```diff
@@ -0,0 +1,112 @@
...
+	routeTable, err := r.findExisting()
+	if err != nil {
+		if err == routeTableFindError {
+			return false, nil
+		}
+		return false, microerror.MaskAny(err)
+	}
+
+	r.id = *routeTable.RouteTableId
+	return true, nil
```
asymmetric: Whitespace before return plz :bowing_man:

resources/aws/routetable.go
```diff
@@ -0,0 +1,112 @@
...
+		return microerror.MaskAny(err)
+	}
+
+	if _, err := r.Clients.EC2.DeleteRouteTable(&ec2.DeleteRouteTableInput{
+		RouteTableId: routeTable.RouteTableId,
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return microerror.MaskAny(notImplementedMethodError)
```
asymmetric: Shouldn't we return `nil` here?

service/create/service.go
```diff
@@ -42,6 +42,16 @@ const (
 	// Prefixes used for machine names.
 	prefixMaster string = "master"
 	prefixWorker string = "worker"
+	// Sufixes used for subnets
```
JosephSalisbury: suffix, not sufix

resources/aws/instance.go
```diff
@@ -161,6 +161,14 @@ func (i *Instance) Delete() error {
 		return microerror.MaskAny(err)
 	}
 
+	if err := i.Clients.EC2.WaitUntilInstanceTerminated(&ec2.DescribeInstancesInput{
```
asymmetric: This can take arbitrarily long (sometimes up to 30 secs), are we sure we need to wait? What's the advantage?

I think it would be better to instead not re-use instances in the *Shutting Down* state.

service/create/service.go
```diff
@@ -382,10 +494,50 @@ func (s *Service) Boot() {
...
+						s.logger.Log("error", fmt.Sprintf("could not delete security group: %s", errgo.Details(err)))
+					}
+
+					// Delete VPC
+					var vpc resources.ResourceWithID
+					vpc = &awsresources.VPC{
+						Name:      cluster.Name,
+						AWSEntity: awsresources.AWSEntity{Clients: clients},
+					}
+					if err := vpc.Delete(); err != nil {
```
asymmetric: I think the `vpc.id` will be `nil` here, causing `Delete()` to fail.

asymmetric: I see, you call 'findExisting' in `Delete()'. All good.

resources/aws/subnet.go
```diff
@@ -0,0 +1,141 @@
...
+type Subnet struct {
+	AvailabilityZone string
+	CidrBlock        string
+	Name             string
+	VpcID            string
+	id               string
+	AWSEntity
+}
+
+func (s *Subnet) findExisting() (*ec2.Subnet, error) {
```
asymmetric: This doesn't need to be have a pointer receiver. WDYT, should we pay attention to these things?

resources/aws/instance.go
nhlfr: Yes, we need to wait - without that we cannot remove subnets and VPC-s.

nhlfr: I can put that to goroutines in the future.

resources/aws/subnet.go
```diff
@@ -0,0 +1,141 @@
...
+	if _, err := s.Clients.EC2.DeleteSubnet(&ec2.DeleteSubnetInput{
+		SubnetId: subnet.SubnetId,
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
+
+func (s *Subnet) ID() string {
```
asymmetric: I would remove the pointer here.

resources/aws/vpc.go
```diff
@@ -0,0 +1,124 @@
...
+	if _, err := v.Clients.EC2.DeleteVpc(&ec2.DeleteVpcInput{
+		VpcId: vpc.VpcId,
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
+
+func (v *VPC) ID() string {
```
asymmetric: Here too, I'd remove the pointer.

resources/aws/subnet.go
nhlfr: I cannot, it will not implement the interface then.

resources/aws/vpc.go
nhlfr: ditto

nhlfr:
@asymmetric @JosephSalisbury ptala

PR #93, opened 2017-04-04
Make cluster name more unique

asymmetric:
We use the following schema for bucket names:

```
$AMAZONID-g8s-$CUSTOMERID
```


JosephSalisbury:
If we put the customer ID in as well, that should make the space pretty big?

asymmetric:
@JosephSalisbury it might just be that I misunderstood what a customer ID looks like. If it's a simple name like foobar, then it's not unique enough. If OTOH it's something more UUID-y, then of course yes :)

iaguis:
Bucket names are globally unique. A common pattern is using `$AWS_ACCOUNT_ID-WHATEVER`, maybe we can do that here too?

JosephSalisbury:
Ah, didn't know they needed to be globally unique. Prefixing with some big uuid (aws account id) makes sense.

JosephSalisbury:
Random thought - we can't guarantee that anything we choose is going to be unique globally - do we need to have some kind of retry?

asymmetric:
Created [this issue](https://github.com/giantswarm/aws-operator/issues/94) to track that @JosephSalisbury .

JosephSalisbury:
Okay cool, let's just go with what works for now then, we can do perfect later

asymmetric:
@rossf7 I thought #94 was more for handling the retrying, and here is where we discuss the prefix.

rossf7:
@asymmetric Sorry I posted this in the wrong issue. Here is my 2 cents on the prefix.

I believe the globally unique name is because each S3 bucket gets a DNS entry for serving files over HTTP. I'd prefer to see a longer prefix to reduce the chance of collisions.

The other approach I've seen used is to include a random section in the bucket name.

iaguis:
How about `$AWS_USER_ID-g8s-$RANDOM_SECTION-`?

asymmetric:
@iaguis sounds OK to me. I'd use a hash of the content instead of pure random data though.

iaguis:
You don't know the content when you create it, do you?

asymmetric:
That is a good point indeed.

rossf7:
@iaguis I like that and it should reduce the chance of collision a lot. Do we want to agree the length of the random section?

I'd say not too long. Is this displayed to the user in the UI or is it just for the bucket name?

asymmetric:
It's displayed in the CLI output atm. I think 8 chars should be OK for now. WDYT?

rossf7:
I think 8 chars is a good length for the random part.

asymmetric:
On second thought, there are issues with using a random value: the operator is stateless and might crash at any time. That means we cannot store the random value somewhere on cluster creation and expect it to be there on deletion.

Me and @nhlfr discussed this and we thought we could use a hash of the customerID instead. This respects the property of *1 customer = 1 bucket*, which is what we want. And if other parts of the TPR change (i.e. a cluster is added/removed), the bucket name doesn't change, which is also what we want.

The other possibility is that we don't append a hash/random value at all. I think this isn't so bad for now, because either way, we're going to have to deal with collisions.

WDYT? @JosephSalisbury @rossf7

rossf7:
Yes I see the problem with a random value.

If hashing the customerID is easy and helps with collisions lets do it. Otherwise I'd say include the customerID without a hash. We can come back to it when we deal with collisions. @JosephSalisbury do you agree?

JosephSalisbury:
I think this the actual decision here isn't _super_ important, as we're gonna have to deal with collisions eventually. No strong opinion.

asymmetric:
I'll go with the easier route then :)

asymmetric:
PTALA.

rossf7:
LGTM so actually approving this time! @asymmetric thanks for taking care of this.

PR #95, opened 2017-04-05
examples: Update cluster example

nhlfr:
awstpr and clutertpr added more detailed fields for configuring Calico, Kubernetes and the Docker images used for deployment.


examples/cluster.yml
```diff
@@ -10,10 +10,35 @@ spec:
...
+
+    calico:
+      subnet: "192.168.0.0"
+      cidr: "24"
+      mtu: "1500"
+
+    kubernetes:
+      api:
+        domain: "k8s.cluster.giantswarm.io"
+        insecurePort: "8080"
```
JosephSalisbury: I get json errors here, I think it needs to be just a plain int, as well as the line below

nhlfr: Good point, fixed that. I will also change MTU to int in `clustertpr` for better consistency.

examples/cluster.yml
```diff
@@ -10,10 +10,35 @@ spec:
...
+      imageNamespace: "giantswarm"
+
+    etcd:
+      domain: "k8s.cluster.giantswarm.io"
+      prefix: "example-cluster"
+
+    calico:
+      subnet: "192.168.0.0"
+      cidr: "24"
+      mtu: "1500"
```
asymmetric: Should this be unquoted because of https://github.com/giantswarm/clustertpr/pull/8?

nhlfr: done

PR #97, opened 2017-04-05
resources: Add SecurityGroup resource

nhlfr:
Security groups are needed both for VPCs and ELBs. Both of them will use this resource.

Ref #27
Ref #69

### DIFF
--- a/resources/aws/securitygroup.go
+++ b/resources/aws/securitygroup.go
@@ -1,0 +1,69 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+type SecurityGroup struct {
+	Description string
+	GroupName   string
+	VpcID       string
+	PortsToOpen []int
+	id          string
+	AWSEntity
+}
+
+func (s *SecurityGroup) CreateIfNotExists() (bool, error) {
+	return false, microerror.MaskAny(notImplementedMethodError)
+}
+
+func (s *SecurityGroup) openPort(port int) error {
+	if _, err := s.Clients.EC2.AuthorizeSecurityGroupIngress(&ec2.AuthorizeSecurityGroupIngressInput{
+		CidrIp:     aws.String("0.0.0.0/0"),
+		GroupId:    aws.String(s.ID()),
+		IpProtocol: aws.String("tcp"),
+		FromPort:   aws.Int64(0),
+		ToPort:     aws.Int64(int64(port)),
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
+
+func (s *SecurityGroup) CreateOrFail() error {
+	securityGroup, err := s.Clients.EC2.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
+		Description: aws.String(s.Description),
+		GroupName:   aws.String(s.GroupName),
+		VpcId:       aws.String(s.VpcID),
+	})
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	s.id = *securityGroup.GroupId
+
+	for _, port := range s.PortsToOpen {
+		if err := s.openPort(port); err != nil {
+			return microerror.MaskAny(err)
+		}
+	}
+
+	return nil
+}
+
+func (s *SecurityGroup) Delete() error {
+	if _, err := s.Clients.EC2.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+		GroupId: aws.String(s.ID()),
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
+
+func (s *SecurityGroup) ID() string {
+	return s.id
+}
--- a/resources/resources.go
+++ b/resources/resources.go
@@ -6,6 +6,11 @@ type Resource interface {
 	Delete() error
 }
 
+type ResourceWithID interface {
+	ID() string
+	Resource
+}
+
 type NamedResource interface {
 	Name() string
 	Resource


resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+var (
+	portsToOpen []int = []int{
+		// SSH
+		22,
+		// Insecure apiserver
+		80,
+		// apiserver
+		443,
```
asymmetric: We were using 6443 in #70, and that's what I see [elsewhere too](https://github.com/giantswarm/kubernetesd/blob/491a7c6cc47706dacf3b54d9027fa297b7bb4c94/kubernetes/configmap.yaml#L45). Which is it?

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+		// Insecure apiserver
+		8080,
+	}
+)
+
+type SecurityGroup struct {
+	Description string
+	GroupName   string
+	VpcID       string
+	name        string
```
asymmetric: Can't we call this `id`, since it contains the `groupID`?

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (s *SecurityGroup) CreateIfNotExists() (bool, error) {
+	return false, microerror.MaskAny(notImplementedMethodError)
+}
+
+func (s *SecurityGroup) openPort(port int) error {
+	if _, err := s.Clients.EC2.AuthorizeSecurityGroupIngress(&ec2.AuthorizeSecurityGroupIngressInput{
+		CidrIp:     aws.String("0.0.0.0/0"),
+		GroupId:    aws.String(s.Name()),
```
asymmetric: I think this should be `name()`

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+		IpProtocol: aws.String("tcp"),
```
asymmetric: I would prefer if this was capitalized.

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+	AWSEntity
+}
+
+func (s *SecurityGroup) CreateIfNotExists() (bool, error) {
+	return false, microerror.MaskAny(notImplementedMethodError)
+}
+
+func (s *SecurityGroup) openPort(port int) error {
+	if _, err := s.Clients.EC2.AuthorizeSecurityGroupIngress(&ec2.AuthorizeSecurityGroupIngressInput{
+		CidrIp:     aws.String("0.0.0.0/0"),
```
asymmetric: I think what you're doing here is allowing inbound connections on those ports from **any** IP. Do we want to be so permissive?

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,77 @@
...
+		if err := s.openPort(port); err != nil {
+			return microerror.MaskAny(err)
+		}
+	}
+
+	return nil
+}
+
+func (s *SecurityGroup) Delete() error {
+	return microerror.MaskAny(notImplementedMethodError)
```
asymmetric: Should we implement this instead? :)

resources/aws/securitygroup.go
nhlfr: Using 6443 was a temporary solution for #70 which IMO should be disregarded, but I wasn't aware that kubernetesd is using that too. Will add it.

resources/aws/securitygroup.go
nhlfr: Then I need to create another interface, but OK, can do.

asymmetric: It's not the most important thing

resources/aws/securitygroup.go
nhlfr: Yes. What alse do you suggest? Maybe we could make it configurable, but IMO it's too early for that.

resources/aws/securitygroup.go
nhlfr: No, `name` is a private field, `Name` is a public method.

resources/aws/securitygroup.go
nhlfr: But it isn't and you can't do anything about that https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AuthorizeSecurityGroupIngress.html

resources/aws/securitygroup.go
nhlfr: Sure, can do

resources/aws/securitygroup.go
asymmetric: @JosephSalisbury WDYT?

resources/aws/securitygroup.go
asymmetric: :cry:

resources/aws/securitygroup.go
asymmetric: I think if we leave it like this then we should definitely add an issue to remember to change it in the future.

resources/aws/securitygroup.go
nhlfr: Done (it means, I included both 443 and 6443, just in case)

resources/aws/securitygroup.go
nhlfr: Sure, I can create it

resources/aws/securitygroup.go
nhlfr: done

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,85 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+var (
+	portsToOpen []int = []int{
```
asymmetric: I get a linter warning here saying you can omit the `[]int` on the LHS.

nhlfr: done

resources/aws/securitygroup.go
JosephSalisbury: We have these ports in the TPR, no? Does it make sense to use those?

resources/aws/securitygroup.go
JosephSalisbury: My gut feeling is that this is too open, but I'm not certain what would improve the situation. Issue is fine with me.

resources/aws/securitygroup.go
nhlfr: Not all of them, but probably everything except SSH. I can use what is available in TPR.

nhlfr: done

resources/aws/securitygroup.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	sshPort = 22
+)
+
+type SecurityGroup struct {
+	Description string
+	GroupName   string
+	VpcID       string
+	Cluster     clustertpr.Cluster
```
asymmetric: I'm not so sure this type needs to have access to the whole TPR. But I understand that this could be a compromise to get us running. In that case though, we should address this later and have an issue to track it.

nhlfr: Where do you want to create that list of ports, in which function and module, and based on what input?

asymmetric: * Where you create the struct (in our case, `service.go`)
* Based on the TPR

I'm just advocating for separating concerns.

asymmetric: But yeah, it's not the most pressing concern, so you're also free to disregard :)

nhlfr: You convinced me, done

asymmetric:
Please also create the issue wrt IP addresses.

PR #100, opened 2017-04-06
resources: Add CreateIfNotExists method for SecurityGroup

nhlfr:
Ref #27
Ref #69

### DIFF
--- a/resources/aws/errors.go
+++ b/resources/aws/errors.go
@@ -8,4 +8,6 @@ var (
 	notImplementedMethodError = errgo.New("not implemented")
 
 	noBucketInBucketObjectError = errgo.New("Object needs to belong to some bucket")
+
+	securityGroupCreateAndFindError = errgo.New("Couldn't create security group, but couldn't find the existing one either")
 )
--- a/resources/aws/securitygroup.go
+++ b/resources/aws/securitygroup.go
@@ -1,9 +1,13 @@
 package aws
 
 import (
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	microerror "github.com/giantswarm/microkit/error"
+
+	awsclient "github.com/giantswarm/aws-operator/client/aws"
 )
 
 type SecurityGroup struct {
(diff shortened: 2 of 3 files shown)


resources/aws/errors.go
```diff
@@ -8,4 +8,6 @@ var (
 	notImplementedMethodError = errgo.New("not implemented")
 
 	noBucketInBucketObjectError = errgo.New("Object needs to belong to some bucket")
+
+	securityGroupCreateAndFindError = errgo.New("Couldn't create security group, but couldn't also find the existing one")
```
asymmetric: "...but couldn't find the existing one either"

resources/aws/securitygroup.go
```diff
@@ -15,8 +19,47 @@ type SecurityGroup struct {
...
+					aws.String(s.GroupName),
+				},
+			},
+		},
+	})
+	if err != nil {
+		return "", microerror.MaskAny(err)
+	}
+
+	for _, securityGroup := range securityGroups.SecurityGroups {
```
asymmetric: I would check if `securityGroups` is empty and return an error, to be consistent with other places in the code: we end with the happy path, and return errors in `if` clauses.

I think it also makes the code more explicit. WDYT?

resources/aws/securitygroup.go
```diff
@@ -15,8 +19,47 @@ type SecurityGroup struct {
...
+			id, err := s.existingSecurityGroupID()
+			if err != nil {
+				return false, microerror.MaskAny(err)
+			}
+			s.id = id
+			return false, nil
+		}
+		return false, microerror.MaskAny(err)
+	}
+	return true, nil
```
asymmetric: I think it would be nice to have newlines before returns here and on line 60, for consistency.

resources/aws/errors.go
nhlfr: done

resources/aws/securitygroup.go
nhlfr: done

resources/aws/securitygroup.go
nhlfr: done

PR #107, opened 2017-04-13
resources: Add alias for KMS keys

### LINKED ISSUE #106: Give KMS keys the name
We need to assing the `Name` tag to KMS keys - that will simplify debugging problems with cluster deployments.

nhlfr:
Fixes #106

### DIFF
--- a/resources/aws/kms.go
+++ b/resources/aws/kms.go
@@ -3,13 +3,19 @@ package aws
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
 	microerror "github.com/giantswarm/microkit/error"
 )
 
 type KMSKey struct {
-	arn string
+	Name string
+	arn  string
 	AWSEntity
+}
+
+func (kk KMSKey) fullAlias() string {
+	return fmt.Sprintf("alias/%s", kk.Name)
 }
 
 func (kk *KMSKey) CreateIfNotExists() (bool, error) {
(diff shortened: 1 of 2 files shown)


resources/aws/kms.go
```diff
@@ -21,14 +27,31 @@ func (kk *KMSKey) CreateOrFail() error {
...
+
 	kk.arn = *key.KeyMetadata.Arn
+
 	return nil
 }
 
-func (kms *KMSKey) Delete() error {
-	return microerror.MaskAny(notImplementedMethodError)
+func (kk *KMSKey) Delete() error {
+	// AWS API doesn't allow to delete the KMS key itself, but we remove the alias and leave the key nameless.
```
rossf7: We can't delete the key because of the KMS API restriction. But I think there is still a cleanup problem because there is a default limit of 1000 keys per region. 

http://docs.aws.amazon.com/kms/latest/developerguide/limits.html

Can you raise an issue for the cleanup task and add to the comment?

resources/aws/kms.go
```diff
@@ -21,14 +27,31 @@ func (kk *KMSKey) CreateOrFail() error {
...
 	}
+
+	if _, err := kk.Clients.KMS.CreateAlias(&kms.CreateAliasInput{
+		// Alias names need to start from "alias/" prefix.
+		AliasName:   aws.String(kk.fullAlias()),
+		TargetKeyId: key.KeyMetadata.Arn,
+	}); err != nil {
+		return microerror.MaskAny(err)
+	}
+
```
rossf7: Nitpick but can you get rid of the empty line

resources/aws/kms.go
asymmetric: Can't we use [`ScheduleKeyDeletion`](https://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.ScheduleKeyDeletion) at least, so the key will eventually be deleted?

resources/aws/kms.go
```diff
@@ -21,14 +27,31 @@ func (kk *KMSKey) CreateOrFail() error {
...
+	return nil
 }
 
-func (kms *KMSKey) Arn() string {
-	return kms.arn
+func (kk *KMSKey) Arn() string {
+	return kk.arn
```
asymmetric: Please no pointer receiver here <3

resources/aws/kms.go
nhlfr: I see no reason for that. I like to keep blocks like

```
something, err := Something()
if err != nil {
    return err
}
```
separated by empty lines. IMO it increases readability, it doesn't mix error handling with the other more important things.

nhlfr:
@rossf7 PTAL again

resources/aws/kms.go
rossf7: Cool that makes sense to me.

PR #115, opened 2017-04-17
service: remove unused fsKeyPairProvider

kopiczko:


service/create/keypair.go
```diff
@@ -2,7 +2,6 @@ package create
 
 import (
```
asymmetric: Do we even use this module at all?

kopiczko:
@nhlfr updated.

PR #116, opened 2017-04-17
aws: get rid of *FindError

kopiczko:
All `*FindError`s were used only internally. I think they added only a clutter as they weren't in fact returned outside the package.

### DIFF
--- a/resources/aws/gateway.go
+++ b/resources/aws/gateway.go
@@ -2,7 +2,6 @@ package aws
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -16,8 +15,8 @@ type Gateway struct {
 	AWSEntity
 }
 
-func (g Gateway) findExisting() (*ec2.InternetGateway, error) {
-	gateways, err := g.Clients.EC2.DescribeInternetGateways(&ec2.DescribeInternetGatewaysInput{
+func (g Gateway) list() ([]*ec2.InternetGateway, error) {
+	out, err := g.Clients.EC2.DescribeInternetGateways(&ec2.DescribeInternetGatewaysInput{
 		Filters: []*ec2.Filter{
 			&ec2.Filter{
 				Name: aws.String(fmt.Sprintf("tag:%s", tagKeyName)),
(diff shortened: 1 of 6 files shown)


resources/aws/securitygroup.go
```diff
@@ -35,27 +35,17 @@ func (s SecurityGroup) findExisting() (*ec2.SecurityGroup, error) {
...
-			if err != nil {
-				return false, microerror.MaskAny(err)
+			securityGroups, err := s.list()
+			if err == nil && len(securityGroups) > 0 {
+				s.id = *securityGroups[0].GroupId
 			}
-			s.id = *securityGroup.GroupId
-
-			return false, nil
+			return false, microerror.MaskAny(err)
```
asymmetric: I'm not sure if this makes the intent clearer. Yes, `microerror.MaskAny(nil)` returns `nil`, but I find the previous implementation more explicit.

resources/aws/gateway.go
```diff
@@ -103,15 +88,18 @@ func (g *Gateway) CreateOrFail() error {
...
+	gateways, err := g.list()
 	if err != nil {
 		return microerror.MaskAny(err)
 	}
 
-	if _, err := g.Clients.EC2.DeleteInternetGateway(&ec2.DeleteInternetGatewayInput{
-		InternetGatewayId: gateway.InternetGatewayId,
-	}); err != nil {
-		return microerror.MaskAny(err)
+	for _, gateway := range gateways {
```
asymmetric: I'm not sure whether we want this `Delete()` to delete more than one gateway. Logically, the operation seems more to mean "delete *this* gateway". Ditto for the others.

resources/aws/subnet.go
```diff
@@ -20,39 +19,26 @@ type Subnet struct {
 	AWSEntity
 }
 
-func (s Subnet) findExisting() (*ec2.Subnet, error) {
-	subnets, err := s.Clients.EC2.DescribeSubnets(&ec2.DescribeSubnetsInput{
+func (s *Subnet) list() ([]*ec2.Subnet, error) {
```
asymmetric: This doesn't need to be a pointer receiver.

resources/aws/gateway.go
nhlfr: There will be always one gateway, assuming that someone didn't create manually some gateway with the same name - and I'm also 99% sure that creating manually such a conflicting gateway is just impossible. I think that it's a matter of whether we prefer to
- explicitly check if list is empty
- iterate over it
Which is a stylistical dilemma which I have no strong feelings about, so feel free to argue (and of course reach some consensus) if you do.

But after all, the list should always contain 0 or 1 element.

resources/aws/gateway.go
```diff
@@ -103,15 +88,18 @@ func (g *Gateway) CreateOrFail() error {
...
+		_, err := g.Clients.EC2.DeleteInternetGateway(&ec2.DeleteInternetGatewayInput{
+			InternetGatewayId: gateway.InternetGatewayId,
+		})
+		if err != nil {
+			return microerror.MaskAny(err)
+		}
 	}
```
nhlfr: But IMO the more important problem which I care about more is dealing with situations when the list is empty. Here, in the `Delete` method. I think that we should return an error to inform user about that in logs. That was my intention when I intoduced the "coudnt' find..." errors - to just log that. To simplify debugging.

asymmetric: Agree for informing the user about the resource not existing (which we are not doing atm). It would be nice to do it with a type assertion instead of a string comparison though. Cfr #112.

resources/aws/subnet.go
kopiczko: This was by mistake, however I think it's good practice to make all receivers consistent. Otherwise an instance will fulfil different interfaces depending on passing it by value or as a pointer. WDYT?

resources/aws/gateway.go
kopiczko: I agree with  @nhlfr and actually start to like previous design more. I didn't realize that `Delete` may return these errors. I'm closing the PR.

One thing, which is still disturbing though is that `findExisting` methods may possibly return indeterministic resource if filters are incorrect, or something else creates a resource that matches filters.
Perhaps there should be a stronger assertion like `len(gateways.InternetGateways) == 1`? I guess in that case  human intervention is required.

nhlfr: Agree on the stronger assertion, that would be a nice thing to implement.

Issue #118, opened 2017-04-19
Create separate Security Groups for masters/workers


asymmetric:

JosephSalisbury:
@rossf7 this is sorted now?

rossf7:
Yes this is deployed.

PR #128, opened 2017-04-20
[WIP] Create separate security groups for masters/workers

rossf7:
Towards giantswarm/giantswarm#1381  and implements #118 WIP

Separate master and workers security groups. Needed for guest cluster ingress.


service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+		Description: groupName,
+		GroupName:   groupName,
+		VpcID:       input.VpcID,
+		PortsToOpen: input.PortsToOpen,
+		AWSEntity:   awsresources.AWSEntity{Clients: input.Clients},
+	}
+	securityGroupCreated, err := securityGroup.CreateIfNotExists()
+	if err != nil {
+		s.logger.Log("error", fmt.Sprintf("could not create security group '%s': %s", groupName, errgo.Details(err)))
+		return nil, err
```
asymmetric: We should mask the error here

service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+		Description: groupName,
+		GroupName:   groupName,
+		AWSEntity:   awsresources.AWSEntity{Clients: input.Clients},
+	}
+	if err := securityGroup.Delete(); err != nil {
+		s.logger.Log("error", fmt.Sprintf("could not delete security group '%s': %s", groupName, errgo.Details(err)))
+		return err
+	} else {
+		s.logger.Log("info", "deleted security group '%s'", groupName)
+		return nil
```
asymmetric: Maybe it's just me but I'd rather have this return at the end of the function, outside the conditional.

service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func securityGroupName(input securityGroupInput) string {
+	return fmt.Sprintf("%s-%s", input.Cluster.Name, input.SecurityGroupType)
```
asymmetric: Maybe we should check if those fields are not set, and return an error. Otherwise we might encounter weird bugs at runtime.

service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+	awsutil "github.com/giantswarm/aws-operator/client/aws"
+	"github.com/giantswarm/aws-operator/resources"
+	awsresources "github.com/giantswarm/aws-operator/resources/aws"
+)
+
+type securityGroupInput struct {
+	Clients           awsutil.Clients
+	Cluster           awstpr.CustomObject
+	PortsToOpen       []int
+	SecurityGroupType string
```
asymmetric: I think we don't use this field, do we?

service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/juju/errgo"
+
+	awsutil "github.com/giantswarm/aws-operator/client/aws"
+	"github.com/giantswarm/aws-operator/resources"
+	awsresources "github.com/giantswarm/aws-operator/resources/aws"
+)
+
+type securityGroupInput struct {
+	Clients           awsutil.Clients
+	Cluster           awstpr.CustomObject
```
asymmetric: Do we need to pass the whole TPO?

service/create/security_group.go
```diff
@@ -0,0 +1,66 @@
...
+	VpcID             string
+}
+
+func (s *Service) createSecurityGroup(input securityGroupInput) (resources.ResourceWithID, error) {
+	groupName := securityGroupName(input)
+
+	securityGroup := &awsresources.SecurityGroup{
+		Description: groupName,
+		GroupName:   groupName,
+		VpcID:       input.VpcID,
```
asymmetric: I tend to capitalize all acronyms, so `VPCID`. Does it look weird?

service/create/service.go
```diff
@@ -514,17 +522,24 @@ func (s *Service) Boot() {
...
+						Cluster:           cluster,
+						SecurityGroupType: prefixMaster,
 					}
-					if err := securityGroup.Delete(); err != nil {
-						s.logger.Log("error", fmt.Sprintf("could not delete security group: %s", errgo.Details(err)))
-					} else {
-						s.logger.Log("info", "deleted security group")
+					if err := s.deleteSecurityGroup(mastersSGInput); err != nil {
+						s.logger.Log("error", "could not delete masters security group")
+					}
```
asymmetric: We should log the successful cases.

asymmetric:
BTW, sorry for reviewing a WIP ;)

rossf7:
@asymmetric No worries :) I'll make sure the deps and these other issues get resolved.

service/create/security_group.go
rossf7: New PR has a smaller securityGroupInput without the whole TPO.

service/create/security_group.go
rossf7: I also prefer VPCID its a shame AWS use VpcID. I've gone will all caps.

service/create/security_group.go
rossf7: Field is no longer required.

service/create/security_group.go
rossf7: New function just takes in the cluster name and a constant. I agree on the error checking but if cluster name is not set it needs to be detected earlier.

rossf7:
@asymmetric closing this for giantswarm/aws-operator/pull/144 which doesn't have any vendor changes and addresses your comments.

PR #141, opened 2017-04-21
Vendor: Updated clustertpr to fix json format string problems

rossf7:
Towards giantswarm/giantswarm#1388

Fix for JSON deserialising problems.


rossf7:
More changes were brought through because of https://github.com/giantswarm/awstpr/pull/19

Issue #99, opened 2017-04-05
Make security group rules less permissive


nhlfr:
For now our Source/Destination `CidrIp` is `0.0.0.0/0`, it's too open

asymmetric:
Here's what I think the `Source/Dest` rules should look like for each of the SGs we currently have:

## Masters SG

* G8S cluster
* Workers VPC

### Workers SG

* Customers, so `0.0.0.0`

asymmetric:
[List](https://coreos.com/kubernetes/docs/latest/kubernetes-networking.html) of ports and sources.

PR #144, opened 2017-04-24
Create separate master and worker security groups

rossf7:
Towards giantswarm/giantswarm#1371 and implements #118 

Creates separate master and worker security groups which is a requirement for guest cluster ingress controller access. The workers security group opens the secure and insecure node ports for the ingress controller as well as SSH.

The SSH access CIDR is 0.0.0.0/0 which is too open but matches the master SG. #99 is already open to lock this down.

### DIFF
--- a/service/create/security_group.go
+++ b/service/create/security_group.go
@@ -18,7 +18,6 @@ type securityGroupInput struct {
 }
 
 func (s *Service) createSecurityGroup(input securityGroupInput) (resources.ResourceWithID, error) {
-
 	var securityGroup resources.ResourceWithID
 	securityGroup = &awsresources.SecurityGroup{
 		Description: input.GroupName,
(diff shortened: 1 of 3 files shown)


service/create/security_group.go
```diff
@@ -0,0 +1,62 @@
...
+
+type securityGroupInput struct {
+	Clients     awsutil.Clients
+	GroupName   string
+	PortsToOpen []int
+	VPCID       string
+}
+
+func (s *Service) createSecurityGroup(input securityGroupInput) (resources.ResourceWithID, error) {
+
```
JosephSalisbury: nitpick: no newline after function definition

service/create/security_group.go
```diff
@@ -0,0 +1,62 @@
...
+		s.logger.Log("info", fmt.Sprintf("created security group '%s'", input.GroupName))
+	} else {
+		s.logger.Log("info", fmt.Sprintf("security group '%s' already exists, reusing", input.GroupName))
+	}
+
+	return securityGroup, nil
+}
+
+func (s *Service) deleteSecurityGroup(input securityGroupInput) error {
+
```
JosephSalisbury: nitpick: no newline after function definition

Issue #147, opened 2017-04-24
Check bucket is in correct region


JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1388

If the bucket is created in a region, and then we try to create clusters in another region, the bucket is reused, but is in the wrong region.

We should check the bucket is in the correct region before reusing.

JosephSalisbury:
Or alternatively, add the region to the bucket name

PR #148, opened 2017-04-24
Adds region to bucket name

JosephSalisbury:
Towards https://github.com/giantswarm/aws-operator/issues/147

If we create a bucket in a particular region, and then try to create
clusters for the same customer in a different region, we re-use
the previous bucket, which fails as it is in the wrong region.

This changeset adds the region to the bucketname, so the buckets
are correct for each region.


JosephSalisbury:
yeah, no strong feelings. i'll move the region to the end

PR #164, opened 2017-04-26
glide: Update k8scloudconfig version

nhlfr:


nhlfr:
Yeah, we should always use glide-vc

PR #175, opened 2017-04-28
service: place workers in public subnet

asymmetric:


asymmetric:
Towards https://github.com/giantswarm/giantswarm/issues/1388.

PR #196, opened 2017-05-08
Updated examples

rossf7:
This PR adds the missing service account secret. It also updates the example cluster TPO to fix issues launching clusters locally.

### DIFF
--- a/examples/cluster.yml
+++ b/examples/cluster.yml
@@ -1,7 +1,7 @@
 apiVersion: "cluster.giantswarm.io/v1"
 kind: Aws
 metadata:
-  name: test
+  name: example-cluster
 spec:
   cluster:
     cluster:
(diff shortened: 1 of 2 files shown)


examples/cluster.yml
```diff
@@ -29,6 +29,8 @@ spec:
         insecurePort: 8080
         securePort: 443
         clusterIPRange: "192.168.0.0/24"
+      dns:
+        ip: "172.29.0.10"
```
rossf7: If this value isn't set it causes problems with the worker cloudconfig

examples/cluster.yml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "cluster.giantswarm.io/v1"
 kind: Aws
 metadata:
-  name: test
+  name: example-cluster
```
rossf7: Can we change this so it matches the clusterID?

PR #203, opened 2017-05-09
service: open kubelet and calico BGP ports on workers security group

### LINKED ISSUE #202: security group changes for kubectl logs and calico
For both `kubectl logs` and `kubectl exec` to work the master needs to be able to access the kubelet port (default 10250) on the workers.

https://github.com/giantswarm/k8scloudconfig/blob/master/templates.go#L1371

asymmetric:
The kubelet port comes from the TPO, whereas the BGP network port is hardcoded (since it's a low level detail).

Closes #202.

### DIFF
--- a/examples/cluster.yml
+++ b/examples/cluster.yml
@@ -38,6 +38,8 @@ spec:
         insecurePort: 30010
         securePort: 30011
         domain: "ingress.example.aws.gigantic.io"
+      kubelet:
+        port: 10250
       networkSetup:
         docker:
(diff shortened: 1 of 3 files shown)


rossf7:
I may be wrong but I think there's a problem with the CIDR. If its `0.0.0.0/0` then thats too open. It needs to reference the other security group. So kubelet access is just from masters.

asymmetric:
I agree but we have this problem in general (i.e. for the existing rules too), and [this](https://github.com/giantswarm/aws-operator/issues/99) issue tracks that.

PR #210, opened 2017-05-10

rossf7:


asymmetric:
Did you run `make update-vendor`?

rossf7:
@asymmetric yes I did that.

rossf7:
Yes it pulled in some more updates but I think that matches our current vendoring approach. I like that we have `make update-vendor`.

rossf7:
@asymmetric because I was lazy and used the web UI. I'll fix.

rossf7:
Closing. Will do fresh vendor PR so we have clean history and can move on with our lives. 😆

PR #214, opened 2017-05-10
Move TLS assets from k8scloudconfig to cloud config plugin

nhlfr:
Fixes giantswarm/k8scloudconfig#80


service/create/encrypt.go
```diff
@@ -8,10 +8,27 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/giantswarm/certificatetpr"
-	"github.com/giantswarm/k8scloudconfig"
 	microerror "github.com/giantswarm/microkit/error"
 )
 
+type CompactTLSAssets struct {
```
asymmetric: We don't need this, we can import the one from `certificatetpr`.

service/create/cloud_config.go
```diff
@@ -93,17 +76,264 @@ func (c *CloudConfigExtension) Units() ([]cloudconfig.UnitAsset, error) {
...
+
+	files, err := m.renderFiles(masterFilesMeta)
+	if err != nil {
+		return nil, microerror.MaskAny(err)
+	}
+
+	return files, nil
+}
+
+func (m *MasterCloudConfigExtension) Units() ([]cloudconfig.UnitAsset, error) {
```
asymmetric: Couldn't the two `Units` methods be defined on the `CloudConfigExtension` type, since they don't change?

nhlfr: done

PR #224, opened 2017-05-11
service: fix security group ports

asymmetric:
* Allow connections on etcd port from workers (needed since we use Calico)
* Only allow connections on Calico port from workers on master
* The Ingress security group does not need Calico (it's an SG for an ELB, not instances)

### DIFF
--- a/service/create/security_group.go
+++ b/service/create/security_group.go
@@ -19,9 +19,9 @@ type securityGroupInput struct {
 
 type rulesInput struct {
 	Cluster                awstpr.CustomObject
-	Rules                  []awsresources.Rule
-	OwnSecurityGroupID     string
+	Rules                  []awsresources.SecurityGroupRule
 	MastersSecurityGroupID string
+	WorkersSecurityGroupID string
 	IngressSecurityGroupID string
 }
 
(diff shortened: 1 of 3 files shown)


service/create/security_group.go
```diff
@@ -113,14 +117,14 @@ func (ri rulesInput) workerRules() []awsresources.Rule {
 		},
 		{
 			Port:       calicoBGPNetworkPort,
-			SourceCIDR: defaultCIDR,
+			SourceCIDR: ri.MastersSecurityGroupID,
```
xh3b4sd: Why is an ID applied to a CIDR value? I feel like I miss something. IMO this should be assigned as `SecurityGroupID`.

asymmetric: Yep, it's a bug.

service/create/security_group.go
```diff
@@ -71,30 +71,34 @@ func (s *Service) deleteSecurityGroup(input securityGroupInput) error {
...
+func (ri rulesInput) masterRules() []awsresources.SecurityGroupRule {
+	return []awsresources.SecurityGroupRule{
 		{
 			Port:       ri.Cluster.Spec.Cluster.Kubernetes.API.SecurePort,
 			SourceCIDR: defaultCIDR,
 		},
 		{
 			Port:            ri.Cluster.Spec.Cluster.Etcd.Port,
-			SecurityGroupID: ri.OwnSecurityGroupID,
+			SecurityGroupID: ri.MastersSecurityGroupID,
```
rossf7: This rule should be removed. Its not needed and caused problems in my testing. The default SG is that all traffic is open within the SG. The etcd elb and the master node are in the same SG.
screen shot 2017-05-11 at 2 23 59 pm

asymmetric: Didn't we say we need this because we connect to etcd via the LB, and not `localhost`?

asymmetric: Also, I'm not sure if the default SG is applied to the instance.

rossf7: The LB also has a SG and it uses the master SG. But I see your point as the etcd access is via the ELB and not localhost.

OK lets go with this and retest.

PR #231, opened 2017-05-12
Adds quay.io badge

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1087


JosephSalisbury:
@rossf7 we had a chat about image discoverability previously, I think this should help, I'm doing it for all the larger public projects

and yes, i am a fanboy :(

PR #248, opened 2017-05-17
open read only kubelet port for heapster

asymmetric:
once we upgrade to kubernetes 1.6 and heapster 1.3 with apiserver
deployment we can remove this, as heapster will connect to the API server
instead of to the kubelets directly.


puja108:
`heapster > 1.6` meaning `k8s >1.6 and heapster > 1.3 with apiserver deployment`

asymmetric:
Changed the wording, thanks @puja108.

PR #255, opened 2017-05-18
resources: Decrease healthcheck interval

### LINKED ISSUE #193: ELB default healthcheck interval is too damn high
The default interval for ELB healthchecks is 30 seconds and 10 health responses.

nhlfr:
Fixes #193

### DIFF
--- a/resources/aws/elb.go
+++ b/resources/aws/elb.go
@@ -41,6 +41,11 @@ const (
 	proxyProtocolPolicyNameSuffix = "proxy-protocol-policy"
 	// proxyProtocolAttributeName is the name of the ProxyProtocol attribute we set on the policy.
 	proxyProtocolAttributeName = "ProxyProtocol"
+	// Default values for health checks.
+	healthCheckHealthyThreshold   = 10
+	healthCheckInterval           = 5
+	healthCheckTimeout            = 3
+	healthCheckUnhealthyThreshold = 2
 )
 
 func (lb *ELB) CreateIfNotExists() (bool, error) {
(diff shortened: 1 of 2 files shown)


resources/aws/elb.go
```diff
@@ -92,6 +97,19 @@ func (lb *ELB) CreateOrFail() error {
 		return microerror.MaskAny(err)
 	}
 
+	if _, err := lb.Client.ConfigureHealthCheck(&elb.ConfigureHealthCheckInput{
+		HealthCheck: &elb.HealthCheck{
+			HealthyThreshold:   aws.Int64(int64(healthCheckHealthyThreshold)),
+			Interval:           aws.Int64(int64(healthCheckInterval)),
+			Target:             aws.String(fmt.Sprintf("TCP:%d", lb.PortsToOpen[0].PortELB)),
```
asymmetric: What if `PortsToOpen` is empty?

nhlfr: I've added a check for that above, PTALA.

resources/aws/elb.go
```diff
@@ -66,6 +71,9 @@ func (lb *ELB) CreateOrFail() error {
 	if lb.Client == nil {
 		return microerror.MaskAny(clientNotInitializedError)
 	}
+	if len(lb.PortsToOpen) < 1 {
```
asymmetric: I don't know why you like  `< 1` so much, instead of `== 0` :)

But there was already a comment that we should use `== 0` in these cases.

nhlfr: oh man... ok, done

PR #256, opened 2017-05-18
Reuse KMS keys

### LINKED ISSUE #239: Too many KMS master keys created
We're creating far too many KMS master keys. There are now ~20 more KMS keys in eu-central-1 in about 12 hours even though we've only launched 2 or 3 clusters.

I think its a problem with changing to a ListWatch and the KMS key creation logic. This logging looks suspicious. I think we need to check for a key with the alias before creating the key.

```
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:180","info":"creating cluster 'zv2m4'","time":"17-05-17 06:57:36.019"}
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:217","info":"keypair 'zv2m4' already exists, reusing","time":"17-05-17 06:57:36.510"}
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:220","info":"waiting for k8s secrets...","time":"17-05-17 06:57:36.510"}
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:227","info":"created KMS key for cluster 'zv2m4'","time":"17-05-17 06:57:36.552"}
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:237","error":"could not create KMS key: [{/go/src/github.com/giantswarm/aws-operator/resources/aws/kms.go:36: } {AlreadyExistsException: An alias with the name arn:aws:kms:eu-west-1:292127003750:alias/zv2m4 already exists\n\tstatus code: 400, request id: 1c1631a0-3ace-11e7-bfb2-a1e7ce797414}]","time":"17-05-17 06:57:36.708"}
```

asymmetric:
Closes #239 by making sure that we only create a key when we are sure
that we can also create an alias for it.

Previously we would create the key, fail because of an already existing
alias, and then fail to remove the key, since we look for them by alias.

Thanks @rossf7!

### DIFF
--- a/resources/aws/kms.go
+++ b/resources/aws/kms.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/kms"
 	microerror "github.com/giantswarm/microkit/error"
 )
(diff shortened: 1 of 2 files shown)


rossf7:
@asymmetric that's a pleasure :) really happy we tracked this down. Looks good but there is a problem in CI.

rossf7:
@kopiczko Yes this solves the problem of creating too many KMS keys.

Due to the ListWatch the same cluster can get processed multiple times. The CreateIfNotExists means extra keys aren't created. The KMS API is a bit painful but we can look up keys by their alias which is the cluster ID.

PR #261, opened 2017-05-22
DNS records should match format used by API

rossf7:
We need to remove the Route 53 hosted zones for aws.giantswarm.io and aws.gigantic.io. This PR updates the domain names to use the same format as the API.


rossf7:
The longer format is the one we're using but it has the customer name and it shouldn't. Until we have our own AWS installation this is an issue.

I'll need redo the PR to remove the customer name.

PR #262, opened 2017-05-22
use custom decoder

asymmetric:
This reverts #260, which was implicitly decoding to a `[]byte`.

This approach makes more sense, and is used by coreos/prometheus-operator.

We also add debugging around decoding failures.


kopiczko:
@asymmetric: out of curiosity:

> This approach makes more sense

Why?

asymmetric:
@kopiczko good question. Turns out it doesn't, but I thought this change would help us get rid of the decoding errors, which is not the case.

PR #268, opened 2017-05-23
test ugoji fix

asymmetric:
Test whether giantswarm/awstpr#25 works.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,6 +19,6 @@ test:
 
 deployment:
   master:
-    branch: master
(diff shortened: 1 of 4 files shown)


kopiczko:
@asymmetric have you seen https://github.com/coreos/etcd-operator/blob/master/pkg/spec/etcd_cluster_list.go#L33-L62 ?

asymmetric:
@kopiczko see https://github.com/giantswarm/aws-operator/pull/269, that's where development is happening now.

PR #273, opened 2017-05-24
WIP: Restrict security group rules

rossf7:
Towards giantswarm/giantswarm#1497

### DIFF
--- a/service/create/security_group.go
+++ b/service/create/security_group.go
@@ -89,8 +89,8 @@ func (ri rulesInput) masterRules() []awsresources.SecurityGroupRule {
 			SourceCIDR: defaultCIDR,
 		},
 		{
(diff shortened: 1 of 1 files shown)


asymmetric:
Isn't this what we were doing previously? It won't work unless the network setup change, i.e. if the connection still comes with a public IP, then the SG won't apply.

rossf7:
Yes this is to debug it. I agree I think the problem is the public IP for etcd.

The problem is currently the kubelet ports are unsecured and open. This was temporary just to unblock testing.

asymmetric:
Yes, we should have a meeting to discuss network layout. Is fixing this now a priority? I thought we wanted to do it later.

rossf7:
Security group changes will be done later.

PR #276, opened 2017-05-24
Readme: update docker image URL

kopiczko:


asymmetric:
BTW the GitHub diff highlighter fails so hard here.

kopiczko:
@asymmetric not that much as when you have indentation changed, I wish it had `-w` option :)

asymmetric:
It does! Append `?w=1` to the URL. https://github.com/blog/967-github-secrets

asymmetric:
But the problem here is that it's highlighting somthing that **didn't** change!

kopiczko:
OMG thank you ❤️ `?w=1`

PR #284, opened 2017-05-26
Do not create Hosted Zones

asymmetric:
Use Hosted Zone IDs from TPO, as their creation is considered part of the manual setup process.

**Note**: I haven't removed the [logic](https://github.com/giantswarm/aws-operator/blob/master/resources/aws/hosted_zone.go) for HZ creation, as I thought we might need it in the future. OTOH we can easily retrieve it from the history if we do. WDYT?

Fixes https://github.com/giantswarm/giantswarm/issues/1491


kopiczko:
On code removal: I don't have strong opinion.

rossf7:
I think it would be better to remove the hosted zone code. As you say we can get it back from the history if we need it.

asymmetric:
@rossf7 Please feel free to do it if you're free. I'm currently busy with this PR still, as I think I left out some important bits.

rossf7:
@asymmetric Cool, on it :)

rossf7:
@asymmetric LGTM and the kubernetesd change is merged

PR #297, opened 2017-05-30
Updated example TPO and docs

rossf7:
- Update the example TPO to use 1.6 hyperkube
- Sets Kubernetes.Domain to `cluster.local` for kubedns
- Updates cluster IP ranges to match those in kubernetesd

### DIFF
--- a/examples/cluster.yml
+++ b/examples/cluster.yml
@@ -35,7 +35,7 @@ spec:
       domain: "cluster.local"
       hyperkube:
         docker:
-          image: "giantswarm/hyperkube:v1.6.4_coreos.0"
+          image: "quay.io/giantswarm/hyperkube:v1.6.4_coreos.0"
       ingressController:
         insecurePort: 30010
         securePort: 30011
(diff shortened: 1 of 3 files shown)


examples/cluster.yml
```diff
@@ -27,13 +27,15 @@ spec:
...
-        clusterIPRange: "192.168.0.0/24"
+        clusterIPRange: "172.31.0.0/24"
       dns:
-        ip: "172.29.0.10"
+        ip: "172.31.0.10"
+      domain: "cluster.local"
       hyperkube:
         docker:
-          image: "giantswarm/hyperkube:v1.5.2_coreos.0"
+          image: "giantswarm/hyperkube:v1.6.4_coreos.0"
```
asymmetric: Should we also point to quay.io?

rossf7: Thanks good spot. That's fixed

kopiczko:
I'm glad it's up to date <3

PR #298, opened 2017-05-30

rossf7:


rossf7:
Ping @xh3b4sd @puja108 @kopiczko @asymmetric @nhlfr merged cloudconfig with kubedns fix for search domain

PR #308, opened 2017-06-07
Auto Scaling Group for workers

### LINKED ISSUE #258: Use auto scaling groups on AWS

asymmetric:
### Changes

* We don't attach instances to the Ingress ELB anymore. Instead, we attach the ELB to the workers ASG.
* Instance creation/deletion is handled by the ASG, not the `runMachines`/`deleteMachines` methods anymore
* For now, only workers.

**NOTE**: AFAIK, there's no way to name instances created by an ASG, so we lose that kind of visual cue in the UI.

Closes #258.

### DIFF
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -7,15 +7,28 @@ import (
 )
 
 type AutoScalingGroup struct {
-	Client                  *autoscaling.AutoScaling
-	Name                    string
-	MinSize                 int
-	MaxSize                 int
-	AvailabilityZone        string
+	// AvailabilityZone is the AZ the instances will be placed in.
+	AvailabilityZone string
+	// HealthCheckGracePeriod is the time, in seconds, that the instances are
+	// given after boot before the healtchecks start.
+	HealthCheckGracePeriod int
+	// LaunchConfigurationName is the name of the Launch Configuration used for the instances.
 	LaunchConfigurationName string
-	LoadBalancerName        string
-	VPCZoneIdentifier       string
-	HealthCheckGracePeriod  int
+	// LoadBalancerName is the name of the ELB that will front the instances.
+	LoadBalancerName string
+	// MaxSize is the maximum amount of insances that will be created in this ASG.
+	MaxSize int
+	// MinSize is the minimum amount of insances in this ASG. There will never be
+	// less than MinSize instances running.
+	MinSize int
+	// Name is the ASG name.
+	Name string
+	// VPCZoneIdentifier is the Subnet ID of the subnet the instances should be
+	// placed in.
+	VPCZoneIdentifier string
+
+	// Dependencies.
+	Client *autoscaling.AutoScaling
 }
 
 func (asg *AutoScalingGroup) CreateOrFail() error {
@@ -52,7 +65,8 @@ func (asg *AutoScalingGroup) Delete() error {
 
 	params := &autoscaling.DeleteAutoScalingGroupInput{
 		AutoScalingGroupName: aws.String(asg.Name),
-		ForceDelete:          aws.Bool(true),
+		// We force deletion of the ASG even if instances are still terminating.
+		ForceDelete: aws.Bool(true),
 	}
 
 	if _, err := asg.Client.DeleteAutoScalingGroup(params); err != nil {
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -13,6 +13,7 @@ type AutoScalingGroup struct {
 	MaxSize                 int
 	AvailabilityZone        string
 	LaunchConfigurationName string
+	LoadBalancerName        string
 	VPCZoneIdentifier       string
 	HealthCheckGracePeriod  int
 }
@@ -30,8 +31,11 @@ func (asg *AutoScalingGroup) CreateOrFail() error {
 			aws.String(asg.AvailabilityZone),
 		},
 		LaunchConfigurationName: aws.String(asg.LaunchConfigurationName),
-		VPCZoneIdentifier:       aws.String(asg.VPCZoneIdentifier),
-		HealthCheckGracePeriod:  aws.Int64(int64(asg.HealthCheckGracePeriod)),
+		LoadBalancerNames: []*string{
+			aws.String(asg.LoadBalancerName),
+		},
+		VPCZoneIdentifier:      aws.String(asg.VPCZoneIdentifier),
+		HealthCheckGracePeriod: aws.Int64(int64(asg.HealthCheckGracePeriod)),
 	}
 
 	if _, err := asg.Client.CreateAutoScalingGroup(params); err != nil {
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -49,6 +49,18 @@ func (asg *AutoScalingGroup) CreateOrFail() error {
 		},
 		VPCZoneIdentifier:      aws.String(asg.VPCZoneIdentifier),
 		HealthCheckGracePeriod: aws.Int64(int64(asg.HealthCheckGracePeriod)),
+		Tags: []*autoscaling.Tag{
+			{
+				Key:               aws.String(tagKeyName),
+				PropagateAtLaunch: aws.Bool(true),
+				Value:             aws.String(asg.Name),
+			},
+			{
+				Key:               aws.String(tagKeyCluster),
+				PropagateAtLaunch: aws.Bool(true),
+				Value:             aws.String(asg.Name),
+			},
+		},
 	}
 
 	if _, err := asg.Client.CreateAutoScalingGroup(params); err != nil {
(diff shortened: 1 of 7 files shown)


service/create/service.go
```diff
@@ -824,38 +824,19 @@ func (s *Service) onAdd(obj interface{}) {
...
-	if err != nil {
-		s.logger.Log("error", errgo.Details(err))
-		return
-	}
-
 	// If the policy couldn't be created and some instances didn't exist before, that means that the cluster
 	// is inconsistent and most problably its deployment broke in the middle during the previous run of
 	// aws-operator.
-	if (anyMastersCreated || anyWorkersCreated) && (kmsKeyErr != nil || policyErr != nil) {
+	if anyMastersCreated && (kmsKeyErr != nil || policyErr != nil) {
```
asymmetric: @nhlfr do you know if this still works as intended? i.e. now that we don't use `runMachines` for the workers, do we still catch inconsistent states?

service/create/error.go
```diff
@@ -25,7 +25,7 @@ func IsMalformedCloudConfigKey(err error) bool {
 	return errgo.Cause(err) == malformedCloudConfigKeyError
 }
 
-var missingCloudConfigKeyError = errgo.New("missing cloud config key")
+var missingCloudConfigKeyError = errgo.New("missing key in the cloudconfig")
```
rossf7: Nice :)

service/create/launch_configuration.go
```diff
@@ -91,21 +93,23 @@ func (s *Service) createLaunchConfiguration(input launchConfigurationInput) (boo
 		return false, microerror.MaskAny(err)
 	}
 
-	launchConfigName, err := launchConfigurationName(input.cluster, input.prefix)
+	launchConfigName, err := launchConfigurationName(input.cluster, input.prefix, securityGroupID)
```
rossf7: Why do we need to add securityGroupID to the name? 

It's a later change but when we support multiple ASGs with different instance types they will need separate launch configs. But I think they would use the same security group.

service/create/launch_configuration.go
```diff
@@ -115,19 +119,40 @@ func (s *Service) createLaunchConfiguration(input launchConfigurationInput) (boo
...
-
 	return nil
 }
 
-func launchConfigurationName(cluster awstpr.CustomObject, prefix string) (string, error) {
+// launchConfigurationName uses the cluster ID, a prefix and the security group
+// ID to produce a launch configuration name.  LC names are their unique
+// identifiers in AWS.  The reason we need the securityGroupID in the name is
+// that we can only reuse an LC if it has been created for the current SG.
+// Otherwise, the SG might not exist anymore.
```
rossf7: OK that answers my question on the sec group ID ;)

I guess for multiple ASGs we could add the instance type to the name. But that is a later change so we can address it then.

resources/aws/auto_scaling_group.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+type AutoScalingGroup struct {
+	// AvailabilityZone is the AZ the instances will be placed in.
+	AvailabilityZone string
+	// HealthCheckGracePeriod is the time, in seconds, that the instances are
+	// given after boot before the healtchecks start.
```
rossf7: Typo on healthchecks

resources/aws/auto_scaling_group.go
```diff
@@ -0,0 +1,77 @@
...
+	HealthCheckGracePeriod int
+	// LaunchConfigurationName is the name of the Launch Configuration used for the instances.
+	LaunchConfigurationName string
+	// LoadBalancerName is the name of the ELB that will front the instances.
+	LoadBalancerName string
+	// MaxSize is the maximum amount of insances that will be created in this ASG.
```
rossf7: Typo "maximum number of instances"

resources/aws/auto_scaling_group.go
```diff
@@ -0,0 +1,77 @@
...
+	MaxSize int
+	// MinSize is the minimum amount of insances in this ASG. There will never be
```
rossf7: Same as above

service/create/service.go
```diff
@@ -896,12 +877,50 @@ func (s *Service) onAdd(obj interface{}) {
...
+	asg := awsresources.AutoScalingGroup{
+		Client:                  clients.AutoScaling,
+		Name:                    cluster.Name,
+		MinSize:                 asgSize,
+		MaxSize:                 asgSize,
+		AvailabilityZone:        cluster.Spec.AWS.AZ,
+		LaunchConfigurationName: workersLCName,
+		LoadBalancerName:        ingressLB.Name,
+		VPCZoneIdentifier:       publicSubnetID,
+		HealthCheckGracePeriod:  10,
```
rossf7: A comment that this is seconds would be good

service/create/launch_configuration.go
asymmetric: There's a comment in the function's definition that explains that.

Basically, it happened to me that leftover LCs would match, but they had been created with different SGs (which didn't exist anymore).

Admittedly, this might have been a temporary situation. WDYT?

resources/aws/auto_scaling_group.go
asymmetric: It's one of the hardest words to type... ARRRGH!!

resources/aws/auto_scaling_group.go
asymmetric: Second hardest :angry:

service/create/service.go
asymmetric: Good catch - I arbitrarily set on 10 seconds, but am open to suggestions.

service/create/service.go
nhlfr: Oh, I'm not sure whether that fully works now. That piece of code was written when we thought that we cannot reuse KMS keys and now we are doing that without any problems.

But the another thing - why we are touching the code here? It is needed for ASG?

service/create/launch_configuration.go
rossf7: It worries me that the SG IDs are changing. Any idea what caused that?

Although I think we'll need to update the launch config as part of updating worker nodes. So if this makes the launch config more robust maybe it's a good thing.

service/create/service.go
asymmetric: > why we are touching the code here? It is needed for ASG?

We don't use `runInstances` anymore for workers, so `anyWorkersCreated` isn't initialized anymore.

> Oh, I'm not sure whether that fully works now. 

I've created https://github.com/giantswarm/aws-operator/issues/310 to track that.

service/create/launch_configuration.go
asymmetric: The SG ID changes every time, since I was deleting the SG as part of cluster deletion.

asymmetric:
PTALA.

service/create/service.go
```diff
@@ -896,12 +880,50 @@ func (s *Service) onAdd(obj interface{}) {
...
+		return
+	}
+
+	if lcCreated {
+		s.logger.Log("info", fmt.Sprintf("created worker launch config"))
+	} else {
+		s.logger.Log("info", fmt.Sprintf("launch config %s already exists, reusing", cluster.Name))
+	}
+
+	workersLCName, err := launchConfigurationName(cluster, "worker", workersSecurityGroupID)
```
kopiczko: I see above there is `prefixWorker` variable or constant.

It doesn't look good for me anyway. `createLaunchConfiguration` generates LC name, then this function recreates it. Why not just pass the name with `launchConfigurationInput`, and use `lcInput.name`?

service/create/service.go
```diff
@@ -1003,31 +1025,28 @@ func (s *Service) onDelete(obj interface{}) {
...
 	}
 
-	// Delete worker launch configuration.
-	workerLCName, err := launchConfigurationName(cluster, prefixWorker)
-	if err != nil {
+	// Delete workers launch configuration.
+	lcInput := launchConfigurationInput{
+		clients: clients,
+		cluster: cluster,
+		prefix:  "worker",
```
kopiczko: Re my comment above: here would be only: `name: launchConfigurationName(cluster, "worker", workersSecurityGroupID)`

service/create/launch_configuration.go
```diff
@@ -91,21 +93,23 @@ func (s *Service) createLaunchConfiguration(input launchConfigurationInput) (boo
...
-		KeyName:                input.keypairName,
-		SecurityGroupID:        securityGroupID,
-		SmallCloudConfig:       smallCloudconfig,
+		IamInstanceProfileName:   input.instanceProfileName,
+		ImageID:                  imageID,
+		InstanceType:             instanceType,
+		KeyName:                  input.keypairName,
+		SecurityGroupID:          securityGroupID,
+		SmallCloudConfig:         smallCloudconfig,
+		AssociatePublicIpAddress: publicIP,
```
asymmetric: To my credit, this is code Ross wrote that I just moved around ;)

rossf7: Yes these inputs have a lot of fields so alphabetical order makes sense. If we're doing that we should do it everywhere so please raise an issue to tidy this up.

service/create/service.go
rossf7: Agreed this should use the constant please add to the issue. No strong opinion on the refactoring.

PR #309, opened 2017-06-07
Lowers resource requests and limits in line with usage

JosephSalisbury:


asymmetric:
@JosephSalisbury For my interest: you observed the actual memory/CPU usage of the operator, and these are more realistic values?

JosephSalisbury:
Yeah, but I don't think we really need this right now. Closing.

JosephSalisbury:
Me being too ocd.

PR #321, opened 2017-06-14
support increasing workers ASG size

asymmetric:
Allows setting Minimum, Maximum and Desired ASG sizes to a higher value.

This means that we can go from a fixed-size cluster, to a **bigger** fixed-size cluster. There's no elasticity in the size.

Closes https://github.com/giantswarm/giantswarm/issues/1507.

### DIFF
--- a/service/create/service.go
+++ b/service/create/service.go
@@ -180,6 +180,7 @@ func (s *Service) Boot() {
 			tpr.ResyncPeriod,
 			cache.ResourceEventHandlerFuncs{
 				AddFunc:    s.onAdd,
+				UpdateFunc: s.onUpdate,
 				DeleteFunc: s.onDelete,
 			},
 		)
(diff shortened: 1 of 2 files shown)


service/create/service.go
```diff
@@ -1299,3 +1296,48 @@ func (s *Service) onDelete(obj interface{}) {
...
+
+	oldSize := len(oldCluster.Spec.AWS.Workers)
+	newSize := len(cluster.Spec.AWS.Workers)
+
+	if oldSize == newSize {
+		// We get update events for all sorts of changes. We are currently only
+		// interested in changes to one property, so we ignore all the others.
+		return
+	} else if oldSize > newSize {
+		s.logger.Log("clusterID", cluster.Spec.Cluster.Cluster.ID, "error", "cluster size reduction not supported")
```
rossf7: Nitpick but I think this should be logged as info rather than error. As its just not supported by the operator yet.

PR #327, opened 2017-06-21
allow getting objects from whole bucket

asymmetric:
since we use a bucket per install, it doesn't make sense to add
additional restrictions based on sub-directories.

the instances in the cluster will be able to access all objects in the
cluster's bucket.


asymmetric:
We are using subdirs. What I did here is allow access to everything, including all subdirs.

rossf7:
Ack, I'm too fried this week. Thanks for clarifying!

PR #331, opened 2017-06-22
Replace tabs with spaces

rossf7:
Replace tabs with spaces for cloudconfig


rossf7:
The templates were edited for the storage change and I think that replaced the tabs with spaces. I checked k8scloudconfig itself and I couldn't see any tabs.

PR #333, opened 2017-06-23
update k8scloudconfig and fix indents

rossf7:


service/create/templates.go
```diff
@@ -120,17 +120,17 @@ write_files:
...
+    metadata:
+      name: gp2
+      annotations:
+        storageclass.beta.kubernetes.io/is-default-class: "true"
+      labels:
+        kubernetes.io/cluster-service: "true"
+        addonmanager.kubernetes.io/mode: EnsureExists
+    provisioner: kubernetes.io/aws-ebs
+    parameters:
+    type: gp2
```
puja108: type needs to be further in under paramters

PR #340, opened 2017-07-12
WIP: Allow all traffic between masters and workers for Calico

### LINKED ISSUE #336: Allow all traffic between masters and workers for Calico
we now only allow few specifics ports from cidr `0.0.0.0/0`, so that means we block all other traffic even internally

This breaks calico.

We need to add rules which allow all traffic(that's how it's called on web) from either both security groups (master and worker) or from `10.0.0.0/8`

rossf7:
Fixes #336


rossf7:
Replaced by #348

PR #341, opened 2017-07-13
Add wildcard domain to cluster template

rossf7:


examples/local/cluster.tmpl.yaml
```diff
@@ -40,6 +40,7 @@ spec:
         insecurePort: 30010
         securePort: 30011
         domain: "ingress.${CLUSTER_NAME}.${COMMON_DOMAIN_INGRESS}"
+        wildcardDomain: "*.${CLUSTER_NAME}.${COMMON_DOMAIN_INGRESS}"
```
kopiczko: For my interest: why we need both `domain` and `wildcardDomain`?

rossf7: We create a wildcard record to make it easier to test the ingress. Both records are in `clustertpr` which makes the domain names a bit more flexible.

PR #348, opened 2017-07-21
Allow all traffic between masters and workers for Calico

### LINKED ISSUE #336: Allow all traffic between masters and workers for Calico

rossf7:
Fixes #336 

This PR allows all traffic between the masters and workers security groups. This is needed so Calico can route traffic between the nodes. 

It also deletes any security group rules that reference other groups. This is needed because you can't delete a security group that is referenced in another security group.

### DIFF
--- a/resources/aws/security_group.go
+++ b/resources/aws/security_group.go
@@ -120,6 +120,48 @@ func (s *SecurityGroup) createRule(rule SecurityGroupRule) error {
 	return nil
 }
 
+// deleteRule creates a security group rule.
+// SourceCIDR always takes precedence over SecurityGroupID.
+func (s *SecurityGroup) deleteRule(rule SecurityGroupRule) error {
+	groupID, err := s.GetID()
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	var params *ec2.RevokeSecurityGroupEgressInput
+	if rule.SourceCIDR != "" {
+		params = &ec2.RevokeSecurityGroupEgressInput{
+			CidrIp:     aws.String(rule.SourceCIDR),
+			GroupId:    aws.String(groupID),
+			IpProtocol: aws.String(rule.Protocol),
+			FromPort:   aws.Int64(int64(rule.Port)),
+			ToPort:     aws.Int64(int64(rule.Port)),
+		}
+	} else {
+		params = &ec2.RevokeSecurityGroupEgressInput{
+			GroupId: aws.String(groupID),
+			IpPermissions: []*ec2.IpPermission{
+				&ec2.IpPermission{
+					FromPort:   aws.Int64(int64(rule.Port)),
+					ToPort:     aws.Int64(int64(rule.Port)),
+					IpProtocol: aws.String(rule.Protocol),
+					UserIdGroupPairs: []*ec2.UserIdGroupPair{
+						{
+							GroupId: aws.String(rule.SecurityGroupID),
+						},
+					},
+				},
+			},
+		}
+	}
+
+	if _, err := s.Clients.EC2.RevokeSecurityGroupEgress(params); err != nil {
+		return microerror.MaskAny(err)
+	}
+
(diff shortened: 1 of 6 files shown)


resources/aws/security_group.go
```diff
@@ -88,7 +93,7 @@ func (s *SecurityGroup) createRule(rule SecurityGroupRule) error {
 		params = &ec2.AuthorizeSecurityGroupIngressInput{
 			CidrIp:     aws.String(rule.SourceCIDR),
 			GroupId:    aws.String(groupID),
-			IpProtocol: aws.String("tcp"),
+			IpProtocol: aws.String(rule.Protocol),
```
rossf7: Protocol is now configurable to support "allow all" as well as TCP.

resources/aws/security_group_rules.go
```diff
@@ -0,0 +1,67 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+// SecurityGroupRules allows AWS security group rules to be deleted. Any rules
```
rossf7: This resource just deletes any rules that reference other sec groups.

resources/aws/security_group_rules.go
```diff
@@ -0,0 +1,67 @@
...
+func (s SecurityGroupRules) Delete() error {
+	securityGroup, err := s.findExisting()
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	var params *ec2.RevokeSecurityGroupIngressInput
+	params = &ec2.RevokeSecurityGroupIngressInput{
+		GroupId:       securityGroup.GroupId,
+		IpPermissions: securityGroup.IpPermissions,
```
rossf7: IpPermissions contains the rules that reference other groups.

service/create/security_group.go
```diff
@@ -26,14 +26,18 @@ type rulesInput struct {
...
+	httpPort             = 80
+	httpsPort            = 443
 	// This port is required in our current kubernetes/heapster setup, but will become unnecessary
 	// once we upgrade to kubernetes 1.6 and heapster 1.3 with apiserver deployment.
 	readOnlyKubeletPort = 10255
-	httpPort            = 80
-	httpsPort           = 443
 	sshPort             = 22
 
+	allowAllProtocols = "-1"
```
rossf7: This is really ugly but it's what the AWS API wants :(

service/create/service.go
```diff
@@ -1182,6 +1182,32 @@ func (s *Service) deleteFunc(obj interface{}) {
 		s.logger.Log("info", "deleted public subnet")
 	}
 
+	// Before the security groups can be deleted any rules referencing other
```
rossf7: Here is where any rules referencing other groups are deleted before the groups are deleted.

service/create/security_group.go
```diff
@@ -26,14 +26,18 @@ type rulesInput struct {
...
+	tcpProtocol       = "tcp"
```
kopiczko: nitpick: I think it should be named `allowTCPProtocol`, or `allowAllProtocols` should be named `allProtocols`.

service/create/security_group.go
```diff
@@ -26,14 +26,18 @@ type rulesInput struct {
 }
 
 const (
+	allowAllPorts        = -1
 	calicoBGPNetworkPort = 179
+	httpPort             = 80
+	httpsPort            = 443
```
kopiczko: This naming also should be consistent. See my other comment about `allow` prefix.

PR #350, opened 2017-07-24
Creating security group rules should be idempotent

### LINKED ISSUE #323: Creating security group rules should be idempotent
We had a problem in an installation where the addFunc was called multiple times because the operator pod was restarted. Creating the security group is idempotent but creating rules isn't and caused the cluster creation to fail.

```
{"caller":"github.com/giantswarm/aws-operator/service/create/service.go:696","error":"could not create rules for security group '23p6g-master': [{/go/src/github.com/giantswarm/aws-operator/resources/aws/security_group.go:140: } {/go/src/github.com/giantswarm/aws-operator/resources/aws/security_group.go:114: } {InvalidPermission.Duplicate: the specified rule \"peer: 0.0.0.0/0, TCP, from port: 443, to port: 443, ALLOW\" already exists\n\tstatus code: 400, request id: 0cd9b832-186a-4ab7-9d0b-1ee345494c41}]","time":"17-06-20 10:38:39.057"}
```

rossf7:
Fixes #323 

Currently when a cluster is processed multiple times the addFunc errors if a security group rule already exists.

This change checks if the rule exists before creating it. It does this by getting the security group and adding a filter for the rule config. I used this approach because you can't get an individual rule from the EC2 API.

### DIFF
--- a/resources/aws/consts.go
+++ b/resources/aws/consts.go
@@ -10,4 +10,10 @@ const (
 	subnetDescription      string = "description"
 	subnetGroupName        string = "group-name"
 	subnetVpcID            string = "vpc-id"
+	// Security Group IP Permission keys
+	ipPermissionCIDR     string = "ip-permission.cidr"
+	ipPermissionFromPort string = "ip-permission.from-port"
+	ipPermissionGroupID  string = "ip-permission.group-id"
+	ipPermissionProtocol string = "ip-permission.protocol"
+	ipPermissionToPort   string = "ip-permission.to-port"
 )
--- a/resources/aws/security_group.go
+++ b/resources/aws/security_group.go
@@ -33,21 +33,13 @@ type SecurityGroupRule struct {
 }
 
 func (s SecurityGroup) findExisting() (*ec2.SecurityGroup, error) {
+	filters, err := s.getGroupFilters()
+	if err != nil {
+		return nil, microerror.MaskAny(err)
+	}
+
 	securityGroups, err := s.Clients.EC2.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
-		Filters: []*ec2.Filter{
-			&ec2.Filter{
-				Name: aws.String(subnetDescription),
-				Values: []*string{
-					aws.String(s.Description),
-				},
-			},
-			&ec2.Filter{
-				Name: aws.String(subnetGroupName),
-				Values: []*string{
-					aws.String(s.GroupName),
-				},
-			},
-		},
+		Filters: filters,
 	})
 	if err != nil {
 		return nil, microerror.MaskAny(err)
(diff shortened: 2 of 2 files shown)


resources/aws/consts.go
```diff
@@ -10,4 +10,10 @@ const (
 	subnetDescription      string = "description"
 	subnetGroupName        string = "group-name"
 	subnetVpcID            string = "vpc-id"
+	// Security Group IP Permission keys
```
rossf7: These keys are used in the filters when checking if the rule exists

resources/aws/security_group.go
```diff
@@ -61,6 +62,94 @@ func (s SecurityGroup) findExisting() (*ec2.SecurityGroup, error) {
...
+			},
+			&ec2.Filter{
+				Name: aws.String(ipPermissionToPort),
+				Values: []*string{
+					aws.String(strconv.Itoa(rule.Port)),
+				},
+			},
+		}
+	} else {
+		filters = []*ec2.Filter{
```
rossf7: We need to use a separate set of filters if the rule references another group

resources/aws/security_group.go
```diff
@@ -82,10 +171,18 @@ func (s *SecurityGroup) CreateIfNotExists() (bool, error) {
...
 // SourceCIDR always takes precedence over SecurityGroupID.
-func (s *SecurityGroup) createRule(rule SecurityGroupRule) error {
+func (s *SecurityGroup) createRuleIfNotExists(rule SecurityGroupRule) (bool, error) {
 	groupID, err := s.GetID()
 	if err != nil {
-		return microerror.MaskAny(err)
+		return false, microerror.MaskAny(err)
+	}
+
+	existingGroup, err := s.findGroupWithRule(rule)
```
rossf7: Here is the new logic to check if the rule exists before creating it.

resources/aws/security_group.go
kopiczko: Can you elaborate? Is this AWS specific or the operator specific?

Can we have common subset of filters and append `SourceCIDR` or `SecurityGroupID` in the `if`s?

kopiczko:
@xh3b4sd I personally find `aws-operator` structure easier to follow than for example `ingress-operator`. I still don't understand by heart the framework we have in `operatorkit` and every time I look at the code I have to adjust me thinking to it. From my POV this is a problem with adapting it.

rossf7:
@xh3b4sd Thanks for raising this. I think we do need to standardise our operators and using the reconciler in the each makes sense.

@kopiczko I also liked your point from last week about the updateFunc should also reconcile the cluster to the desired state. I didn't get a chance to reply on that.

My current problem is how to apply the reconciler in aws-operator and that this will be a big change. 

My preferred approach would be to tackle it as part of updating worker nodes. Maybe we can use the reconciler in the update func to get it working. We can then move the bulk of the logic in the addFunc to also use it.

rossf7:
@kopiczko The filters are needed to query the EC2 API to check if the rule exists. This is because there isn't an endpoint for individual rules. 

Different filters are needed for CIDR rules and those referencing other groups. I'll see if I can remove the duplication.

resources/aws/security_group.go
```diff
@@ -181,3 +256,29 @@ func (s SecurityGroup) GetID() (string, error) {
 
 	return *securityGroup.GroupId, nil
 }
+
+func (s SecurityGroup) getGroupFilters() ([]*ec2.Filter, error) {
```
kopiczko: I think there isn't a reason to move that to a separate method, but I'm fine with that.

rossf7: It's because the filters are used in both the `findExisting` and `findGroupWithRule` methods. I'm going to go with this.

PR #354, opened 2017-07-25
update clustertpr & awstpr

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1666


glide.yaml
```diff
@@ -16,6 +16,10 @@ import:
   - logger
   - server
   - transaction
+- package: github.com/giantswarm/operatorkit
+  verson: master
+- package: google.golang.org/grpc
+  version: v1.2.1
```
kopiczko: I added operatorkit explicitly. I also had to pin grpc.

Issue #363, opened 2017-07-28
Additional permissions needed for instances to pull from ECR


marians:
The worker instances need the following permissions to fetch credentials and images from EC2 container registries (ECR):

- ecr:GetAuthorizationToken
- ecr:BatchCheckLayerAvailability
- ecr:GetDownloadUrlForLayer
- ecr:GetRepositoryPolicy
- ecr:DescribeRepositories
- ecr:ListImages
- ecr:BatchGetImage

See [Kubernetes docs](https://kubernetes.io/docs/concepts/containers/images/#using-aws-ec2-container-registry).

PR #364, opened 2017-07-31
Add IAM policy allowing authenticated pull from ECR

hobti01:
Add an IAM policy on the master and workers to allow image pulling from ECR as implemented in https://github.com/kubernetes/kubernetes/pull/19447 and https://github.com/kubernetes/kubernetes/pull/24369

As noted in https://github.com/kubernetes/kubernetes/pull/19447#issuecomment-174026000 it might be possible to reduce the permission set, but this is the policy implemented upstream.

### DIFF
--- a/resources/aws/policy.go
+++ b/resources/aws/policy.go
@@ -54,6 +54,19 @@ const (
 				"Effect": "Allow",
 				"Action": "s3:GetObject",
 				"Resource": "arn:aws:s3:::%s/*"
+			},
+			{
+				"Effect": "Allow",
+				"Action": [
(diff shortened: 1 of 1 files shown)


rossf7:
We may want to make this configurable for clusters that are not using ECR. But this can be done as a later change.

puja108:
I'd say having it in without using ECR is not a problem, configuration would be rather if only certain clusters should have access to ECR and others not, but yeah, I'll create an issue for that for later.

puja108:
closes https://github.com/giantswarm/aws-operator/issues/363

PR #376, opened 2017-08-17
Refactor updating the ASG Stack

rossf7:
Towards giantswarm/giantswarm#1726

This PR refactors the ASG stack and adds processASGStack which will be called from both the addFunc and the updateFunc. It handles whether to create or update the stack.

For the stack update it checks if an updatable parameter has changed. The Cloud Formation API is only called if there is an update to apply.

### DIFF
--- a/resources/aws/asg_stack.go
+++ b/resources/aws/asg_stack.go
@@ -19,6 +19,8 @@ const (
 	defaultTimeout = 5
 	// imageIDParam is the Cloud Formation parameter name for the ASG image ID.
 	imageIDParam = "ImageID"
+
+	stackDoesNotExistError = "Stack with id %s does not exist"
 )
 
 // ASGStack represents a CloudFormation stack for an Auto Scaling Group.
@@ -198,11 +200,31 @@ func (s *ASGStack) Delete() error {
 	return nil
 }
 
+// CheckIfExists checks if there is an autoscaling group stack in Cloud Formation
+// with the provided name.
+func (s *ASGStack) CheckIfExists() (bool, error) {
+	_, err := s.findExisting()
+	if IsNotFound(err) {
+		return false, nil
+	} else if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	return true, nil
+}
+
 func (s *ASGStack) findExisting() (*cloudformation.Stack, error) {
 	stacks, err := s.Client.DescribeStacks(&cloudformation.DescribeStacksInput{
 		StackName: aws.String(s.Name),
 	})
 	if err != nil {
+		underlying := microerror.Cause(err)
+		if awserr, ok := underlying.(awserr.Error); ok {
+			if awserr.Message() == fmt.Sprintf(stackDoesNotExistError, s.Name) {
+				return nil, microerror.Mask(notFoundError)
+			}
+		}
+
 		return nil, microerror.Mask(err)
 	}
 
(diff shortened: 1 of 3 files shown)


resources/aws/asg_stack.go
```diff
@@ -19,6 +20,8 @@ const (
 	defaultTimeout = 5
 	// imageIDParam is the Cloud Formation parameter name for the ASG image ID.
 	imageIDParam = "ImageID"
+
+	stackDoesNotExistError = "Stack with id %s does not exist"
```
rossf7: The Cloud Formation API returns a validation error when getting a stack that doesn't exist. So string comparison is needed :(

resources/aws/asg_stack.go
```diff
@@ -149,29 +153,32 @@ func (s *ASGStack) Update() error {
 		asgMinSizeParam: strconv.Itoa(s.ASGMinSize),
 		imageIDParam:    s.ImageID,
 	}
-	params := []*cloudformation.Parameter{}
 
-	for _, param := range currentStack.Parameters {
-		if value, ok := updateableParams[*param.ParameterKey]; ok {
-			param.ParameterValue = aws.String(value)
-		}
+	if hasStackChanged(currentStack.Parameters, updateableParams) {
```
rossf7: Now the stack is only updated if an updatable parameter has changed.

resources/aws/asg_stack.go
```diff
@@ -194,11 +201,31 @@ func (s *ASGStack) Delete() error {
 	return nil
 }
 
+// CheckIfExists checks if there is an autoscaling group stack in Cloud Formation
```
rossf7: CheckIfExists is used to decide whether a stack create or update is needed.

resources/aws/asg_stack.go
```diff
@@ -194,11 +201,31 @@ func (s *ASGStack) Delete() error {
...
+}
+
 func (s *ASGStack) findExisting() (*cloudformation.Stack, error) {
 	stacks, err := s.Client.DescribeStacks(&cloudformation.DescribeStacksInput{
 		StackName: aws.String(s.Name),
 	})
 	if err != nil {
+		underlying := microerror.Cause(err)
+		if awserr, ok := underlying.(awserr.Error); ok {
+			if awserr.Message() == fmt.Sprintf(stackDoesNotExistError, s.Name) {
```
rossf7: This is the error handling with the string comparison. Unfortunately I can't find a cleaner way to do it.

resources/aws/asg_stack.go
```diff
@@ -210,3 +237,18 @@ func (s *ASGStack) findExisting() (*cloudformation.Stack, error) {
 
 	return stacks.Stacks[0], nil
 }
+
+func hasStackChanged(params []*cloudformation.Parameter, paramUpdates map[string]string) bool {
```
rossf7: Here the updatable parameters are checked to see if there are any changes.

service/create/asg_stack.go
```diff
@@ -68,6 +67,38 @@ const (
 	defaultEBSVolumeType = "gp2"
 )
 
+func (s *Service) processASGStack(input asgStackInput) (bool, error) {
```
rossf7: processASGStack can be called from the addFunc and the updateFunc. It handles the check for whether the stack exists.

resources/aws/asg_stack.go
kopiczko: I think this is fine, but would be good to have an integration test, just to make sure error message didn't changed.

PR #382, opened 2017-08-25
fix setting transactionResponder dependency

kopiczko:


kopiczko:
Yes, that's why I wanted to decouple transactions from the server...
If we can decide we can get rid of them we should do it.

kopiczko:
Also your idea about making them operator-style means decoupling as well. I like that.

PR #390, opened 2017-08-31
Correctly sets peer owner id from host clients

JosephSalisbury:

If the peer owner id is not set, it defaults to the current account, which is incorrect in cases where host and guest vpcs are in different accounts. This changeset fetches the account ID from IAM, and sets it as the vpc peer owner account id.

### DIFF
--- a/resources/aws/peering.go
+++ b/resources/aws/peering.go
@@ -1,14 +1,19 @@
 package aws
 
 import (
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/giantswarm/microerror"
 )
 
 const (
 	RequesterVpcFilterName = "requester-vpc-info.vpc-id"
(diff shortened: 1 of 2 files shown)


resources/aws/peering.go
```diff
@@ -79,10 +84,19 @@ func (v *VPCPeeringConnection) CreateIfNotExists() (bool, error) {
 }
 
 func (v *VPCPeeringConnection) CreateOrFail() error {
+	resp, err := v.HostClients.IAM.GetUser(&iam.GetUserInput{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	userArn := *resp.User.Arn
+	peerOwnerId := strings.Split(userArn, ":")[accountIDPosition]
```
rossf7: I think it would be good to have a helper function and basic tests for getting the owner ID.

JosephSalisbury: I was cheeky, nicked this from https://github.com/giantswarm/aws-operator/blob/master/client/aws/aws.go#L65

Considering time pressure, I'll break this out into a boyscouting issue. See https://github.com/giantswarm/aws-operator/issues/391

PR #395, opened 2017-09-05
Update local setup

rossf7:
This PR  adds support for setting the VPC Peer ID in the cluster template. It also bumps the k8s version to match installations.


examples/cluster.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       domain: "cluster.local"
       hyperkube:
         docker:
-          image: "quay.io/giantswarm/hyperkube:v1.6.4_coreos.0"
+          image: "quay.io/giantswarm/hyperkube:v1.7.3_coreos.0"
```
rossf7: Bump versions in the example cluster TPO.

examples/cluster.yaml
```diff
@@ -72,9 +72,9 @@ spec:
       publicSubnetCidr: "10.0.128.0/20"
 
     masters:
-    - imageid: "ami-d60ad6b9"
-      instancetype: "t2.medium"
+    - imageid: "ami-33df775c"
+      instancetype: "m3.medium"
```
rossf7: Minor but t2.medium no longer works because of the EBS volume we mount for docker.

examples/local/cluster.tmpl.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       domain: "cluster.local"
       hyperkube:
         docker:
-          image: "quay.io/giantswarm/hyperkube:v1.6.4_coreos.0"
+          image: "quay.io/giantswarm/hyperkube:v1.7.3_coreos.0"
```
rossf7: Bump the k8s version in the template used for local setup.

rossf7:
@MarcelMue I've merged as it's a safe change. If you hit any issues running the local setup please update the guide.

PR #398, opened 2017-09-07
Fix configmap not having correct extension and readme inconsistency

MarcelMue:
The configmap failed after yesterdays changes from `yml` to `yaml`.

Fixed readme referring to `COMMON_DOMAIN_INTERNAL` everywhere now.

### DIFF
--- a/examples/local/README.md
+++ b/examples/local/README.md
@@ -21,7 +21,6 @@ guide, all placeholders must be replaced with sensible values.
 
 - *CLUSTER_NAME* - Cluster's name.
 - *COMMON_DOMAIN* - Cluster's etcd and API common domain.
-- *COMMON_DOMAIN_INTERNAL* - Ingress common domain.
 - *ID_RSA_PUB* - SSH public key to be installed on nodes.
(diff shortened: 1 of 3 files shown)


examples/local/README.md
```diff
@@ -21,7 +21,7 @@ guide, all placeholders must be replaced with sensible values.
 
 - *CLUSTER_NAME* - Cluster's name.
 - *COMMON_DOMAIN* - Cluster's etcd and API common domain.
-- *COMMON_DOMAIN_INGRESS* - Ingress common domain.
+- *COMMON_DOMAIN_INTERNAL* - Ingress common domain.
```
rossf7: This is great, thanks for correcting but `COMMON_DOMAIN_INTERNAL` is no longer needed.

This is because in early installs we had etcd and apiserver in `*.aws.giantswarm.io` and ingress in `*.aws.gigantic.io`. Now we use `gigantic.io` everywhere so only `COMMON_DOMAIN` is needed.

MarcelMue: Removed all instances of `COMMON_DOMAIN_INTERNAL`, also done in Wiki

PR #399, opened 2017-09-07
fix vpc range in local docs

MarcelMue:
`10.0.0.0/16` is occupied in our test setup, therefor changing to a free CIDR

### DIFF
--- a/examples/local/cluster.tmpl.yaml
+++ b/examples/local/cluster.tmpl.yaml
@@ -67,9 +67,9 @@ spec:
       etcd: "${AWS_API_HOSTED_ZONE}"
       ingress: "${AWS_INGRESS_HOSTED_ZONE}"
     vpc:
-      cidr: "10.0.0.0/16"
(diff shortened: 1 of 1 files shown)


examples/local/cluster.tmpl.yaml
```diff
@@ -67,9 +67,9 @@ spec:
       etcd: "${AWS_API_HOSTED_ZONE}"
       ingress: "${AWS_INGRESS_HOSTED_ZONE}"
     vpc:
-      cidr: "10.0.0.0/16"
-      privateSubnetCidr: "10.0.0.0/19"
-      publicSubnetCidr: "10.0.128.0/20"
+      cidr: "10.1.5.0/24"
```
fgimenez: Will this work if two of us (or CI in two separate branches) try this at the same time? Anyway :+1: for the mask change

MarcelMue: Yes collisions will still happen, but should be easy to solve by changing the third digit of the CIDR.

PR #400, opened 2017-09-07
Add AWS SessionToken support

MarcelMue:
This is needed when we use a account with MFA to create a cluster (eg personal account for test purposes).

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -110,7 +110,7 @@ func New(config Config) (*Service, error) {
 		accessKeySecret := config.Viper.GetString(config.Flag.Service.AWS.HostAccessKey.Secret)
 		sessionToken := config.Viper.GetString(config.Flag.Service.AWS.HostAccessKey.Session)
 
-		if accessKeyID == "" && accessKeySecret == "" && sessionToken == "" {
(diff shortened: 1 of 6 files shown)


MarcelMue:
Only tested locally, should properly deploy it somewhere in case it breaks with other logins.

service/service.go
```diff
@@ -100,22 +100,25 @@ func New(config Config) (*Service, error) {
...
 	}
 
 	var awsHostConfig awsclient.Config
 	{
 		accessKeyID := config.Viper.GetString(config.Flag.Service.AWS.HostAccessKey.ID)
 		accessKeySecret := config.Viper.GetString(config.Flag.Service.AWS.HostAccessKey.Secret)
+		sessionToken := config.Viper.GetString(config.Flag.Service.AWS.HostAccessKey.Session)
 
-		if accessKeyID == "" && accessKeySecret == "" {
+		if accessKeyID == "" && accessKeySecret == "" && sessionToken == "" {
```
rossf7: Please remove the session token check here. When deployed the operator uses an IAM account that doesn't use MFA so it won't have a token.

MarcelMue: Done

PR #405, opened 2017-09-12
Validate AZ

### LINKED ISSUE #76: Validate AZ belongs to Region
The AZ format is "eu-west-1a", i.e. it includes the Region, so we should enforce that in the cluster TPR the AZ matches the Region.

rossf7:
Fixes #76 

Validates the Availability Zone in cluster custom object matches the Region. The cluster is validated before the cluster is launched so we catch this before any resources are created.

### DIFF
--- a/service/create/service.go
+++ b/service/create/service.go
@@ -28,8 +28,6 @@ import (
 )
 
 const (
-	ClusterListAPIEndpoint  string = "/apis/cluster.giantswarm.io/v1/awses"
-	ClusterWatchAPIEndpoint string = "/apis/cluster.giantswarm.io/v1/watch/awses"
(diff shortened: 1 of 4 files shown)


service/create/service.go
```diff
@@ -28,8 +28,6 @@ import (
 )
 
 const (
-	ClusterListAPIEndpoint  string = "/apis/cluster.giantswarm.io/v1/awses"
-	ClusterWatchAPIEndpoint string = "/apis/cluster.giantswarm.io/v1/watch/awses"
```
rossf7: Some boyscouting, these constants weren't removed when we switched the TPR creation to operatorkit.

PR #406, opened 2017-09-12
Route to guest vpc

MarcelMue:
towards [#1799](https://github.com/giantswarm/giantswarm/issues/1799)


resources/aws/subnet.go
```diff
@@ -198,3 +198,46 @@ func (s *Subnet) AddHostVPCRoute(routeTable *RouteTable, vpcPeeringConnection *V
 
 	return nil
 }
+
+func (s *Subnet) AddGuestVPCRoute(routeTable *RouteTable, vpcPeeringConnection *VPCPeeringConnection) error {
```
fgimenez: Maybe having a private method in `Subnet` with the common logic might help, with a flag for creating or deleting, I think the code would remain very understandable, but not sure about the conventions in this sense.

service/create/service.go
```diff
@@ -1281,12 +1310,40 @@ func (s *Service) deleteFunc(obj interface{}) {
...
-		AWSEntity: awsresources.AWSEntity{Clients: clients},
+		AWSEntity: awsresources.AWSEntity{
+			Clients:     clients,
+			HostClients: hostClients,
+		},
+	}
+
+	// Delete Guest VPC Route.
+	for _, privateSubnetSpec := range cluster.Spec.AWS.VPC.PrivateSubnets {
+		privateSubnet := &awsresources.Subnet{
```
fgimenez: Also here, `DefaultSubnetConfig()` and `DefaultRouteTableConfig()` might help, i've seen this more often..

examples/local/cluster.tmpl.yaml
```diff
@@ -68,9 +68,14 @@ spec:
...
       ingress: "${AWS_INGRESS_HOSTED_ZONE}"
     vpc:
-      cidr: "10.1.5.0/24"
-      privateSubnetCidr: "10.1.5.0/25"
-      publicSubnetCidr: "10.1.5.128/25"
+      cidr: "10.1.12.0/24"
+      privateSubnetCidr: "10.1.12.0/25"
+      publicSubnetCidr: "10.1.12.128/25"
+      privateSubnets:
+      - name: gauss-worker_0
```
MarcelMue: Only the `routeTableName` is needed after all.

resources/aws/peering.go
```diff
@@ -145,3 +145,7 @@ func (v *VPCPeeringConnection) GetID() (string, error) {
 
 	return *vpcPeeringConnection.VpcPeeringConnectionId, nil
 }
+
+func (v VPCPeeringConnection) FindExisting() (*ec2.VpcPeeringConnection, error) {
```
MarcelMue: I needed this to get some `VPCPeeringConnection` properties, this was called in the same package before.

resources/aws/route.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Route struct {
+	RouteTableID         string
+	DestinationCidrBlock string
+	VpcID                string
+	AWSEntity
+}
+
+func (r *Route) CreateIfNotExists() (bool, error) {
+	return false, microerror.Mask(notImplementedMethodError)
```
MarcelMue: Routes are bound to the `RouteTableID`, the ID changes with each cluster creation - therefore the route cannot easily be reused. (Could be if the route table and route both weren't deleted? not certain)

resources/aws/subnet.go
```diff
@@ -176,25 +176,3 @@ func (s *Subnet) MakePublic(routeTable *RouteTable) error {
...
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	conn, err := vpcPeeringConnection.findExisting()
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	if _, err := s.Clients.EC2.CreateRoute(&ec2.CreateRouteInput{
```
MarcelMue: Moved this to `resources/aws/route.go` as it isn't related to subnets.

MarcelMue:
This is currently running on `gauss` from my local setup:

- A peering connection with local vpc `marcel-cluster` and peered vpc `gauss`
- There are three route tables attached, 2 matched to the `gauss` vpc, 1 matched to `marcel-cluster` vpc

The destinations and targets for the `gauss` route tables (`gauss_private_0`, `gauss_private_1`):
```
Destination   | Target                | Status | Propagated
10.0.0.0/16   | local                 | Active | No
0.0.0.0/0     | nat-006ffaac803f31249 | Active | No
10.1.12.0/24  | pcx-5020a139          | Active | No
```
And for the `marcel-cluster` route table:
```
Destination | Target       | Status | Propagated
10.1.12.0/24| local        | Active | No
0.0.0.0/0   | igw-f2ed229a | Active | No
10.0.0.0/16 | pcx-5020a139 | Active | No
```

This looks like the correct peering to me.

resources/aws/subnet.go
MarcelMue: I changed this around quite a bit, not sure if I made it better or worse but having this in `resources/aws/subnet.go` did not feel right.

examples/local/cluster.tmpl.yaml
rossf7: Nice that just the route table name is needed as it keeps the TPR simpler.

resources/aws/peering.go
rossf7: Good to move this to the resource.

resources/aws/route.go
rossf7: It's fine if we can't reuse the route but please make sure all the peering routes are deleted. We can't delete the VPC until all sub resources are deleted so it's easy to have orphan resources after cluster deletion.

service/create/service.go
```diff
@@ -794,10 +794,53 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+		DestinationCidrBlock: *conn.AccepterVpcInfo.CidrBlock,
+		VpcID:                *conn.VpcPeeringConnectionId,
+		AWSEntity:            awsresources.AWSEntity{Clients: clients},
+	}
+
+	if err := publicRoute.CreateOrFail(); err != nil {
 		return microerror.Maskf(executionFailedError, fmt.Sprintf("could not add host vpc route: '%#v'", err))
 	}
 
+	for _, privateSubnetSpec := range cluster.Spec.AWS.VPC.PrivateSubnets {
```
rossf7: Removing the spec here and just using the route table name will simplify the code a bit.

service/create/service.go
```diff
@@ -794,10 +794,53 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+
+		privateRouteTable := &awsresources.RouteTable{
+			Name:   privateSubnetSpec.RouteTableName,
+			VpcID:  cluster.Spec.AWS.VPC.PeerID,
+			Client: hostClients.EC2,
+		}
+
+		privateRouteTableID, err := privateRouteTable.GetID()
+		if err != nil {
+			return microerror.Maskf(executionFailedError, fmt.Sprintf("could not get private route table ID: '%#v'", err))
```
rossf7: Might be good if the errors included the route table name so we can see which has failed.

service/create/service.go
```diff
@@ -794,10 +794,53 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+		privateRoute := &awsresources.Route{
+			RouteTableID:         privateRouteTableID,
+			DestinationCidrBlock: *conn.RequesterVpcInfo.CidrBlock,
+			VpcID:                *conn.VpcPeeringConnectionId,
+			AWSEntity:            awsresources.AWSEntity{Clients: hostClients},
+		}
+
+		if err := privateRoute.CreateOrFail(); err != nil {
+			return microerror.Maskf(executionFailedError, fmt.Sprintf("could not add guest vpc route: '%#v'", err))
+		}
```
kopiczko: I'm worried those `CreateOrFail` calls will cause errors after ResyncPeriod and operator pod restart. @rossf7 will know more.

rossf7: Yes creating the routes needs to be idempotent for when the cluster is reprocessed after the resync period.

For some other resources we tag them with the cluster ID and then look for a resource with that tag. That might help here so that `CreateIfNotExists` can be implemented.

MarcelMue: I've changed this and it now uses `CreateIfNotExists`, I tested this by deleting the `aws-operator` deployment and restarting it to see if it reprocesses a cluster without errors.

service/create/service.go
```diff
@@ -794,10 +794,53 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
 		return microerror.Maskf(executionFailedError, fmt.Sprintf("could not make subnet public, '%#v'", err))
 	}
 
-	if err := publicSubnet.AddHostVPCRoute(routeTable, vpcPeeringConection.(*awsresources.VPCPeeringConnection)); err != nil {
+	conn, err := vpcPeeringConection.(*awsresources.VPCPeeringConnection).FindExisting()
```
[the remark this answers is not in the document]
MarcelMue: Done.

examples/local/cluster.tmpl.yaml
```diff
@@ -68,9 +68,14 @@ spec:
...
+        routeTableName: gauss_private_0
+      - name: gauss-worker_1
+        routeTableName: gauss_private_1
```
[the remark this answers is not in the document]
MarcelMue: I changed this and added it to the `README`.

examples/local/cluster.tmpl.yaml
```diff
@@ -68,9 +68,14 @@ spec:
       etcd: "${AWS_API_HOSTED_ZONE}"
       ingress: "${AWS_INGRESS_HOSTED_ZONE}"
     vpc:
-      cidr: "10.1.5.0/24"
-      privateSubnetCidr: "10.1.5.0/25"
-      publicSubnetCidr: "10.1.5.128/25"
+      cidr: "10.1.12.0/24"
+      privateSubnetCidr: "10.1.12.0/25"
+      publicSubnetCidr: "10.1.12.128/25"
```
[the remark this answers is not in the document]
MarcelMue: Just for testing reasons, as my `CIDR` for testing clashes with @fgimenez if we test at the same time otherwise.

service/create/service.go
MarcelMue: Now iterating over the route table name.

resources/aws/route.go
MarcelMue: I did this, but it got fairly ugly. `Route` can not be tagged and it does not have `DescribeRouteInput`. The best I could come up with was fetching the Routes from the `RouteTable` and iterating through them.

glide.yaml
```diff
@@ -25,7 +25,7 @@ import:
 - package: google.golang.org/grpc
   version: v1.2.1
 - package: github.com/giantswarm/awstpr
-  version: master
+  version: add-private-subnets
```
JosephSalisbury: TODO

resources/aws/route.go
```diff
@@ -0,0 +1,88 @@
...
+
+type Route struct {
+	DestinationCidrBlock string
+	VpcID                string
+	RouteTable           RouteTable
+	AWSEntity
+}
+
+func (r Route) findExisting() (*ec2.Route, error) {
+
```
JosephSalisbury: Remove this line

service/create/service.go
```diff
@@ -1049,6 +1093,12 @@ func (s *Service) deleteFunc(obj interface{}) {
 		s.logger.Log("error", fmt.Sprintf("could not retrieve amazon account id: '%#v'", err))
 		return
 	}
+	// Retrieve AWS host cluster client.
```
JosephSalisbury: Don't we already have the clients available here?

service/create/service.go
```diff
@@ -794,9 +793,54 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+	if err != nil {
 		return microerror.Maskf(executionFailedError, fmt.Sprintf("could not add host vpc route: '%#v'", err))
 	}
+	if publicRouteCreated {
+		s.logger.Log("info", fmt.Sprintf("created host vpc route for cluster '%s'", key.ClusterID(cluster)))
+	} else {
+		s.logger.Log("info", fmt.Sprintf("host vpc route for cluster '%s' already exists, reusing", key.ClusterID(cluster)))
+	}
+
+	for _, privateRouteTableName := range cluster.Spec.AWS.VPC.RouteTableNames {
```
JosephSalisbury: This should be refactored out

service/create/service.go
```diff
@@ -1279,12 +1329,40 @@ func (s *Service) deleteFunc(obj interface{}) {
...
+			HostClients: hostClients,
+		},
 	}
+	conn, err := vpcPeeringConection.FindExisting()
+	if err != nil {
+		s.logger.Log("error", fmt.Sprintf("could not find vpc peering connection: '%#v'", err))
+	}
+
+	// Delete Guest VPC Route.
+	for _, privateRouteTableName := range cluster.Spec.AWS.VPC.RouteTableNames {
```
JosephSalisbury: This should be refactored out

PR #408, opened 2017-09-14
Migrate legacy ASG workers to CF stack-managed ASG

fgimenez:
Targeted to #379, builds on top of it and tries to address the shortcomings of that approach when dealing with legacy clusters. The old workers' ASG is detected, the workers are detached from it and attached to the new ASG.

### DIFF
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/microerror"
 )
 
@@ -185,11 +186,15 @@ func (asg *AutoScalingGroup) DetachInstances() ([]*autoscaling.Instance, error) 
 	}
 
 	for _, i := range autoScalingGroup.Instances {
-		_, err = asg.Client.DetachInstances(&autoscaling.DetachInstancesInput{
-			AutoScalingGroupName:           aws.String(asg.Name),
-			InstanceIds:                    []*string{i.InstanceId},
-			ShouldDecrementDesiredCapacity: aws.Bool(false),
-		})
+		operation := func() error {
+			_, err = asg.Client.DetachInstances(&autoscaling.DetachInstancesInput{
+				AutoScalingGroupName:           aws.String(asg.Name),
+				InstanceIds:                    []*string{i.InstanceId},
+				ShouldDecrementDesiredCapacity: aws.Bool(false),
+			})
+			return err
+		}
+		err := backoff.Retry(operation, backoff.NewExponentialBackOff())
 		if err != nil {
 			return []*autoscaling.Instance{}, microerror.Mask(err)
 		}
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -174,18 +174,7 @@ func (asg *AutoScalingGroup) FindLegacy() (bool, error) {
 		return false, microerror.Mask(err)
 	}
 
-	var keyNameFound, keyClusterFound bool
-	for _, td := range a.Tags {
-		if *td.Key == tagKeyName && *td.Value == asg.Name {
-			keyNameFound = true
-		} else if *td.Key == tagKeyCluster && *td.Value == asg.ClusterID {
-			keyClusterFound = true
-		} else if strings.HasPrefix(*td.Key, "aws:cloudformation:") {
-			return false, nil
-		}
-	}
-
-	return keyNameFound && keyClusterFound, nil
+	return isLegacyASG(asg.Name, asg.ClusterID, a.Tags), nil
 }
 
 // DetachInstances detaches instances from this ASG and returns their details
@@ -211,8 +200,15 @@ func (asg *AutoScalingGroup) DetachInstances() ([]*autoscaling.Instance, error) 
 
 // AttachInstances returns details of the instances attached to this ASG
 func (asg *AutoScalingGroup) AttachInstances(instances []*autoscaling.Instance) error {
+	// save current termination policies
+	autoScalingGroup, err := asg.findExisting()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	backTerminationPolicies := autoScalingGroup.TerminationPolicies
+
 	// set termination policy to newest first so that the old instances don't get removed
-	terminationPolicy := "NewestInstance"
+	terminationPolicy := newestFirstTerminationPolicy
 	params := &autoscaling.UpdateAutoScalingGroupInput{
 		AutoScalingGroupName: aws.String(asg.Name),
 		TerminationPolicies:  []*string{&terminationPolicy},
@@ -226,7 +222,7 @@ func (asg *AutoScalingGroup) AttachInstances(instances []*autoscaling.Instance) 
 		ids = append(ids, i.InstanceId)
 	}
 
-	_, err := asg.Client.AttachInstances(&autoscaling.AttachInstancesInput{
+	_, err = asg.Client.AttachInstances(&autoscaling.AttachInstancesInput{
 		AutoScalingGroupName: aws.String(asg.Name),
 		InstanceIds:          ids,
 	})
@@ -234,14 +230,28 @@ func (asg *AutoScalingGroup) AttachInstances(instances []*autoscaling.Instance) 
 		return microerror.Mask(err)
 	}
 
-	// back to default termination policy
-	terminationPolicy = "Default"
+	// back to original termination policies
 	params = &autoscaling.UpdateAutoScalingGroupInput{
 		AutoScalingGroupName: aws.String(asg.Name),
-		TerminationPolicies:  []*string{&terminationPolicy},
+		TerminationPolicies:  backTerminationPolicies,
 	}
 	if _, err := asg.Client.UpdateAutoScalingGroup(params); err != nil {
 		return microerror.Mask(err)
 	}
 	return nil
+}
+
+func isLegacyASG(name, clusterID string, tags []*autoscaling.TagDescription) bool {
+	var keyNameFound, keyClusterFound bool
+	for _, td := range tags {
+		if *td.Key == tagKeyName && *td.Value == name {
+			keyNameFound = true
+		} else if *td.Key == tagKeyCluster && *td.Value == clusterID {
+			keyClusterFound = true
+		} else if strings.HasPrefix(*td.Key, "aws:cloudformation:") {
+			return false
+		}
+	}
+
+	return keyNameFound && keyClusterFound
 }
--- a/resources/aws/auto_scaling_group.go
+++ b/resources/aws/auto_scaling_group.go
@@ -1,12 +1,15 @@
 package aws
 
 import (
+	"fmt"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
 )
 
 type AutoScalingGroup struct {
@@ -179,10 +182,14 @@ func (asg *AutoScalingGroup) FindLegacy() (bool, error) {
 }
 
 // DetachInstances detaches instances from this ASG and returns their details
-func (asg *AutoScalingGroup) DetachInstances() ([]*autoscaling.Instance, error) {
+func (asg *AutoScalingGroup) DetachInstances(logger micrologger.Logger) ([]*autoscaling.Instance, error) {
 	autoScalingGroup, err := asg.findExisting()
 	if err != nil {
 		return []*autoscaling.Instance{}, microerror.Mask(err)
+	}
+
+	notify := func(err error, dur time.Duration) {
+		logger.Log("warning", fmt.Sprintf("retrying 'DetachInstances' due to error (%s)", err.Error()))
 	}
 
 	for _, i := range autoScalingGroup.Instances {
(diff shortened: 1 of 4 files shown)


resources/aws/auto_scaling_group.go
```diff
@@ -161,3 +163,85 @@ func (asg *AutoScalingGroup) findExisting() (*autoscaling.Group, error) {
...
+	a, err := asg.findExisting()
+	if IsNotFound(err) {
+		return false, nil
+	} else if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	var keyNameFound, keyClusterFound bool
+	for _, td := range a.Tags {
+		if *td.Key == tagKeyName && *td.Value == asg.Name {
```
xh3b4sd: We could improve this a bit by separating network related actions and standalone algorithms. I am thinking about having a method that does the name, key and prefix checking. This method can then be unit tested and the code here is more easily understandable. Then we also only have one boolean we deal with here.

resources/aws/auto_scaling_group.go
```diff
@@ -161,3 +163,85 @@ func (asg *AutoScalingGroup) findExisting() (*autoscaling.Group, error) {
...
+		}
+	}
+
+	return autoScalingGroup.Instances, nil
+}
+
+// AttachInstances returns details of the instances attached to this ASG
+func (asg *AutoScalingGroup) AttachInstances(instances []*autoscaling.Instance) error {
+	// set termination policy to newest first so that the old instances don't get removed
+	terminationPolicy := "NewestInstance"
```
xh3b4sd: I know the libs are shite and do not provide most of the magic strings but lets please at least have constants for these things and some useful godoc.

service/create/service.go
```diff
@@ -912,6 +914,28 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+
+	var instances []*autoscaling.Instance
+	if legacyASG {
+		// detach instances and delete legacy ASG to prevent name clash
+		instances, err = asg.DetachInstances()
+		if err != nil {
+			return microerror.Maskf(executionFailedError, fmt.Sprintf("detaching legacy worker from ASG failed: '%#v'", err))
+		}
+		if err := asg.Delete(); err != nil {
+			return microerror.Maskf(executionFailedError, fmt.Sprintf("deletingg legacy ASG failed: '%#v'", err))
```
rossf7: Typo

service/create/service.go
```diff
@@ -942,6 +966,17 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
 
 	s.logger.Log("info", fmt.Sprintf("processed autoscaling group stack '%s'", key.AutoScalingGroupName(cluster, prefixWorker)))
 
+	if legacyASG {
+		cfAsg, err := s.findCFASG(clients.AutoScaling, key.AutoScalingGroupName(cluster, prefixWorker), key.ClusterID(cluster))
+		if err != nil {
+			return microerror.Maskf(executionFailedError, fmt.Sprintf("could find CF ASG: '#%v'", err))
```
rossf7: I think the error should be "could not find".

service/create/service.go
```diff
@@ -1383,3 +1418,24 @@ func (s *Service) updateFunc(oldObj, newObj interface{}) {
 
 	s.logger.Log("info", fmt.Sprintf("processed autoscaling group stack '%s'", key.AutoScalingGroupName(cluster, prefixWorker)))
 }
+
+func (s *Service) findCFASG(client *autoscaling.AutoScaling, name string, clusterID string) (awsresources.AutoScalingGroup, error) {
```
rossf7: This API call should really be in the `resources/aws` package. I'm OK with it since its migration code. If we need it long term it should be moved.

service/create/service.go
xh3b4sd: It is also that `Maskf` does the formatting already. No need to use `fmt` here. <3

service/create/service.go
xh3b4sd: IMO there is lots of room for improvement and the code structure will change anyway as soon as we move to the reconciler framework.

resources/aws/auto_scaling_group.go
kopiczko: I know you were explaining that in the standup but I don't follow.

I understand when you add new instances now, they're terminated immediately.
And then when you change back to default nothing happens. Correct? Why is that desired?

resources/aws/auto_scaling_group.go
```diff
@@ -161,3 +163,85 @@ func (asg *AutoScalingGroup) findExisting() (*autoscaling.Group, error) {
...
+	_, err := asg.Client.AttachInstances(&autoscaling.AttachInstancesInput{
+		AutoScalingGroupName: aws.String(asg.Name),
+		InstanceIds:          ids,
+	})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	// back to default termination policy
+	terminationPolicy = "Default"
```
kopiczko: The "old" termination policy didn't have to be `Default`. I think it should be remembered before switching to `NewestInstance`

service/create/service.go
```diff
@@ -912,6 +914,28 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
+		Client:    clients.AutoScaling,
+		Name:      key.AutoScalingGroupName(cluster, prefixWorker),
+		ClusterID: key.ClusterID(cluster),
+	}
+	legacyASG, err := asg.FindLegacy()
+	if err != nil {
+		return microerror.Maskf(executionFailedError, fmt.Sprintf("search for legacy worker ASG failed: '%#v'", err))
+	}
+
+	var instances []*autoscaling.Instance
```
kopiczko: Can we somehow don't reuse `instances` variable? Otherwise this code is not idempotent. When CF fails after deleting legacy ASG we have orphaned instances. If the code was selecting detached instances when the legacy ASG is gone after executing CF the problem would be solved.

resources/aws/auto_scaling_group.go
```diff
@@ -161,3 +163,85 @@ func (asg *AutoScalingGroup) findExisting() (*autoscaling.Group, error) {
...
+	if err != nil {
+		return []*autoscaling.Instance{}, microerror.Mask(err)
+	}
+
+	for _, i := range autoScalingGroup.Instances {
+		_, err = asg.Client.DetachInstances(&autoscaling.DetachInstancesInput{
+			AutoScalingGroupName:           aws.String(asg.Name),
+			InstanceIds:                    []*string{i.InstanceId},
+			ShouldDecrementDesiredCapacity: aws.Bool(false),
+		})
```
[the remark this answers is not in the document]
xh3b4sd: Note that we get retries for free with the reconciler framework.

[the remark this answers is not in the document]
kopiczko: I'd like to retry this particular line: (199) https://github.com/giantswarm/aws-operator/pull/408/files#diff-1124618ae218f5ff63bf668745931e87R199
Otherwise it's possible that only part of instances.

This is inside for loop. I don't know how the framework could improve situation here.

---

I just noticed `InstanceIds:                    []*string{i.InstanceId},`.  @fgimenez why we can't detach all instances in a single call? That would't solve that problem completely w/o retires.

resources/aws/auto_scaling_group.go
fgimenez: Yes, with the default policy the old instances being migrated can be terminated after being attached to the new ASG. Changing the policy to `NewestInstance` makes that the recently created instances (the ones already attached to the CF-controlled ASG) are terminated after the old ones are attached.

I think other approaches are possible, like protecting the instances being transferred, let me know WDYT.

service/create/service.go
fgimenez: Exactly, I didn't spend too much time organising the code given the upcoming changes.

resources/aws/auto_scaling_group.go
fgimenez: Good point, on it

resources/aws/auto_scaling_group.go
kopiczko: Makes sense. Thanks for explanation!

resources/aws/auto_scaling_group.go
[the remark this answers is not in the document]
fgimenez: Indeed, the AWS API was complaining when I tried removing all of them at once

resources/aws/auto_scaling_group.go
fgimenez: Good point! Done, pls take a look

resources/aws/auto_scaling_group.go
fgimenez: Sure, added to `resources/aws/consts.go`

service/create/service.go
fgimenez: Yep, done

service/create/service.go
fgimenez: Good catch! fixed

resources/aws/auto_scaling_group.go
fgimenez: Done, pls take a look

service/create/service.go
fgimenez: mm not sure, there could be other detached instances not related to any ASG and without any reference to the cluster, how could we identify the ones to attach?

resources/aws/auto_scaling_group.go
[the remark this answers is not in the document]
fgimenez: > I'd like to retry this particular line: (199) https://github.com/giantswarm/aws-operator/pull/408/files#diff-1124618ae218f5ff63bf668745931e87R199

Makes sense, on it

resources/aws/auto_scaling_group_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func TestIsLegacyASG(t *testing.T) {
+	nameOk := "my-asg"
+	clusterOk := "my-cluster"
+	tagNameOK := tagKeyName
+	tagClusterOK := tagKeyCluster
+	tagCF := "aws:cloudformation:stack-id"
+
+	t.Run("name and cluster match, no CF tag", func(t *testing.T) {
```
xh3b4sd: We usually write table driven tests. They are easier to extend, more DRY and more readable. Right now I have to closely look for the `!` before `IsLegacyASG` to see if the expected result is true or false. Would you mind rewriting these tests? Check any of our golang projects and you should find this kind of tests. See for instance https://github.com/giantswarm/ipam/blob/032da5280d2dff234739ac38c9445e4d9cfcf2cd/ipam_test.go#L35.

fgimenez: Sure, i started with them but because of the tags the testCases were not that readable, will push the changes to use them

fgimenez: Done, i've moved also the `Run` call description to the table, to make things more clear, pls take a look

resources/aws/auto_scaling_group_test.go
```diff
@@ -0,0 +1,54 @@
...
+		description string
+		name        string
+		clusterID   string
+		tags        []*autoscaling.TagDescription
+		expected    bool
+	}{
+		{"name and cluster match, no CF tag", nameOk, clusterOk, []*autoscaling.TagDescription{
+			{Key: &tagNameOK, Value: &nameOk},
+			{Key: &tagClusterOK, Value: &clusterOk},
+		}, true},
```
xh3b4sd: I know no tool enforces this but my personal preference would be the explicit way. Not using the property name when initializing a structure is not really readbale. 

```
{
  description: ...,
  name: ...,
  clusterID: ...,
  tags: ...,
  expected: ...,
},
```

fgimenez: Indeed, added, much better :)

resources/aws/auto_scaling_group.go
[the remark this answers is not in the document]
fgimenez: Done with simple `backoff.Retry`, pls take a look

service/create/service.go
rossf7: Cool location is fine since lots of code restructuring is coming.

fgimenez: Thank you Ross, you are completely right about this code belonging to `resources/aws`, probably there are things also in the ASG resource that don't belong there either, let's see how we can structure our logic in the next steps.

resources/aws/auto_scaling_group.go
```diff
@@ -161,3 +164,99 @@ func (asg *AutoScalingGroup) findExisting() (*autoscaling.Group, error) {
...
+	for _, i := range autoScalingGroup.Instances {
+		operation := func() error {
+			_, err = asg.Client.DetachInstances(&autoscaling.DetachInstancesInput{
+				AutoScalingGroupName:           aws.String(asg.Name),
+				InstanceIds:                    []*string{i.InstanceId},
+				ShouldDecrementDesiredCapacity: aws.Bool(false),
+			})
+			return err
+		}
+		err := backoff.Retry(operation, backoff.NewExponentialBackOff())
```
kopiczko: We usually use `RetryNotify` and log retry error. Otherwise on retries it looks like the pod is hanging. In order to achieve that you'll have to pass logger to AutoSaclingGroup.

kopiczko: Example: https://github.com/giantswarm/operatorkit/blob/14cddd8e1d187249ff6a47307036a3c14f7052d7/framework/retryresource/resource.go#L94-L101

fgimenez: Thanks, done. I'll retry the full upgrade test on local example now that all the changes are in place

service/create/service.go
kopiczko: I just thought we could make sure there is a tag on each instance before detaching it. That way we could get rid of the backoff in detaching. WDYT?

fgimenez: Totally, i was thinking the same, adding tags to instances would help a lot. I'm thinking of wrapping all the code on a separate component, there's a lot of things going on currently in the migration. This way we could reuse/adapt it in the future, will ping you when ready.

rossf7: Per instance tags would be very useful. We have node IDs in the cluster TPO but these aren't applied to the instances because the tags are from the autoscaling group.

These would be useful for the API so a user could terminate a specific node.

PR #410, opened 2017-09-14
Add script to setup local

jgsqware:
Add 3 scripts to set env variables, setup on minikube and create a cluster. Also a teardown script to remove everything

### DIFF
--- a/examples/local/env.sh
+++ b/examples/local/env.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+if [ -z ${AWS_ACCESS_KEY_ID} ] ; then
+  echo "AWS_ACCESS_KEY_ID not set"
+  exit 1
+fi
+
+if [ -z ${AWS_SECRET_ACCESS_KEY} ] ; then
+  echo "AWS_SECRET_ACCESS_KEY not set"
+  exit 1
+fi
+
+if [ -z ${AWS_SESSION_TOKEN} ] ; then
+  echo "AWS_SESSION_TOKEN not set"
+  exit 1
+fi
+
+if [ -z ${AWS_AMI} ] ; then
+  echo "AWS_AMI not set"
+  exit 1
+fi
+
+if [ -z ${AWS_API_HOSTED_ZONE} ] ; then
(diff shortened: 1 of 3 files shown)


examples/local/teardown.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -x
+
+. ./env.sh
+
+kubectl delete aws -l clusterID=${CLUSTER_NAME}
+
+kubectl delete -f configmap.yaml
+kubectl delete -f configmap-ssh.yaml
+kubectl delete -f deployment.yaml
```
fgimenez: By this time maybe the cluster hasn't been deleted yet, removing the operator can leave resources behind. It would be nice to be able to wait until the cluster is completely deleted before removing the deployment.

examples/local/up.sh
```diff
@@ -0,0 +1,37 @@
...
+    cd ../..
+    GOOS=linux go build github.com/giantswarm/aws-operator
+    docker build -t quay.io/giantswarm/aws-operator:local-dev .
+)
+
+kubectl apply -f ./configmap.yaml
+kubectl apply -f ./configmap-ssh.yaml
+kubectl apply -f ./deployment.yaml
+
+sleep 5
```
fgimenez: You could actually ask for the TPR to be defined before continuing

examples/local/env.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
fgimenez: A bit of a nitpick but not sure if you are using here any bash-specific feature, could you just use `sh` here?

examples/local/teardown.sh
jgsqware: Completely right, just happened yesterday after pushing my PR :)

rossf7: Yes it's definitely better to wait until all the AWS resources have been deleted. There is the log line for `cluster 'xxxxx' deleted` which would work but sadly I can't think of a better way of detecting this.

jgsqware:
Not useful anymore with new local setup

PR #421, opened 2017-10-04
StorageClass encrypted

jgsqware:

Encrypt StorageClass resource by default

fixes https://github.com/giantswarm/vodafone/issues/254


jgsqware:
@puja108 By default, if no key is specified, AWS will create a new key in KVM and use this key to encrypt.
It's possible to specify a key.

WDYT about this? I think it's better to create a new key for the cluster but I don't know how to handle this

puja108:
I think it's ok to let AWS create it by itself, we don't need to control the key(s)

PR #424, opened 2017-10-04
Encrypt ebs volume at rest

jgsqware:

fixes https://github.com/giantswarm/vodafone/issues/243

### DIFF
--- a/resources/aws/launch_configuration.go
+++ b/resources/aws/launch_configuration.go
@@ -28,6 +28,7 @@ const (
 	// defaultEBSVolumeSize is expressed in GB.
 	defaultEBSVolumeSize = 50
 	defaultEBSVolumeType = "gp2"
+	defaultEBSEncrypted  = true
 )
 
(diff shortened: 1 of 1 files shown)


jgsqware:
@puja108 Need some insight about KMS here

puja108:
Doesn't AWS just use some key by default if we don't give it any special key? Like with the other EBS in storage class, I don't think we need to control keys for these.

jgsqware:
For Worker, it's working, EBS is encrypted because it's linked to the launch configuration, not to a snapshot

For Master, the instance is linked to a snapshot that is not currently encrypted. I'll continue digging on this

puja108:
For master we can leave this out I think, as master is reserved for us and not used for user data which is the use case for their encryption. @othylmann WDYT?

jgsqware:
Ok if good for @othylmann I'll merge it

othylmann:
Yep, should be fine @puja108 @jgsqware

PR #434, opened 2017-10-11
Restrict security group ports

rossf7:

This PR locks down the ports on the master and worker security groups. This change is backwards compatible for existing clusters.

The services below still use the default CIDR of `0.0.0.0/0`. Changing these is not backwards compatible.

* k8s apiserver
* etcd 
* ssh

### DIFF
--- a/service/create/security_group.go
+++ b/service/create/security_group.go
@@ -27,14 +27,10 @@ type rulesInput struct {
 }
 
 const (
-	allPorts             = -1
-	calicoBGPNetworkPort = 179
-	httpPort             = 80
-	httpsPort            = 443
-	// This port is required in our current kubernetes/heapster setup, but will become unnecessary
-	// once we upgrade to kubernetes 1.6 and heapster 1.3 with apiserver deployment.
-	readOnlyKubeletPort = 10255
-	sshPort             = 22
+	allPorts  = -1
+	httpPort  = 80
+	httpsPort = 443
+	sshPort   = 22
 
 	allProtocols = "-1"
 	tcpProtocol  = "tcp"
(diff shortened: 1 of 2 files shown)


service/create/security_group.go
```diff
@@ -96,11 +92,6 @@ func (ri rulesInput) masterRules() []awsresources.SecurityGroupRule {
 			Protocol:   tcpProtocol,
 			SourceCIDR: defaultCIDR,
 		},
-		{
-			Port:       calicoBGPNetworkPort,
-			Protocol:   tcpProtocol,
-			SourceCIDR: defaultCIDR,
```
rossf7: Access is allowed to allow all rule for calico.

service/create/security_group.go
```diff
@@ -119,35 +110,20 @@ func (ri rulesInput) masterRules() []awsresources.SecurityGroupRule {
...
-		},
-		{
-			Port:       readOnlyKubeletPort,
-			Protocol:   tcpProtocol,
-			SourceCIDR: defaultCIDR,
+			Port:            ri.Cluster.Spec.Cluster.Kubernetes.IngressController.InsecurePort,
+			Protocol:        tcpProtocol,
+			SecurityGroupID: ri.IngressSecurityGroupID,
```
rossf7: Restricted to the ingress sec group.

service/create/security_group.go
```diff
@@ -119,35 +110,20 @@ func (ri rulesInput) masterRules() []awsresources.SecurityGroupRule {
...
 		{
 			Port:       sshPort,
 			Protocol:   tcpProtocol,
 			SourceCIDR: defaultCIDR,
 		},
-		{
-			Port:       calicoBGPNetworkPort,
-			Protocol:   tcpProtocol,
-			SourceCIDR: defaultCIDR,
-		},
```
rossf7: Access is allowed from allow all rule for calico.

rossf7:
Successfully tested on `gauss`.

There was a problem with the ingress http port due to a bug when creating sec group rules. The filter wasn't included the port so only the first rule for a group was created. Now fixed.

PR #438, opened 2017-10-13
update example tpo template

fgimenez:


examples/local/aws-resource-lab-chart/values.yaml
```diff
@@ -1,7 +1,7 @@
 clusterName: "test-cluster"
 commonDomain: "aws.gigantic.io"
 sshUser: "test-user"
-sshPublicKey: ssh-rsa XXXXXXXXXXXXXXXXXXXXXXXXX"
+sshPublicKey: "ssh-rsa XXXXXXXXXXXXXXXXXXXXXXXXX"
```
rossf7: Thanks for fixing!

fgimenez:
> It would be great if we could run kubernetesd locally. As keeping the cluster object up to date is a pain.

Indeed, this will be required for e2e tests aiui

PR #447, opened 2017-10-16
Set route table when creating public subnet

rossf7:
Set route table when creating public subnet so it's made public.


rossf7:
The route table wasn't set when creating the public subnet which meant there was no default route to the internet gateway.

PR #452, opened 2017-10-17
Private master IP and use VPC peering for SSH access

rossf7:

This PR changes the master to use a private IP for new clusters. To keep SSH access it uses the VPC peering connection from the host cluster. The SSH security group rules are restricted to the host cluster CIDR.

## Issues

For the VPC peering we don't have routes from the public route table used by the bastion. So SSH works from the host cluster nodes but not the bastions.

Options are

- Create routes for the guest cluster CIDR in the public route tables of the host cluster.
- Add the bastion to both the public and private subnets if that is possible.


service/create/service.go
```diff
@@ -774,8 +774,8 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
 		return microerror.Maskf(executionFailedError, fmt.Sprintf("could not get security group '%s' ID: '%#v'", ingressSGInput.GroupName, err))
 	}
 
-	// In rare cases the host cluster CIDR block is nil. Return error rather
-	// that panic. TODO Fix issue #451 that tracks this.
+	// In rare cases the host cluster CIDR block can be nil. Return error rather
+	// than panic. TODO Fix issue #451 that tracks this.
```
rossf7: Fix typos

service/create/service.go
```diff
@@ -920,18 +920,26 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
 		}
 	}
 
-	publicRoute := &awsresources.Route{
+	hostClusterRoute := &awsresources.Route{
```
rossf7: Renamed to make the route name clearer.

service/create/service.go
```diff
@@ -972,13 +980,18 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
 		bucket:              bucket,
 		securityGroup:       mastersSecurityGroup,
-		subnet:              publicSubnet,
 		instanceProfileName: masterPolicy.GetName(),
 		prefix:              prefixMaster,
 	}
 
-	// An EC2 Keypair is needed for legacy clusters. New clusters provide SSH keys via cloud config.
-	if !key.HasClusterVersion(cluster) {
+	if key.HasClusterVersion(cluster) {
```
rossf7: This removes the public IP from the master.

rossf7:
Going to go with this as the 2 options for the bastion access don't involve code changes.

Adding the public route tables is a just config change and is also my preferred option.

PR #460, opened 2017-10-19
Wait for S3 domain in small cloudconfig

rossf7:

Now that the nodes are in the private subnet there is a delay before they have network access. So we now wait for the S3 URL to be available before fetching the large cloudconfig.

My suspicion is the delay on network access is while the NAT gateway is created.

### DIFF
--- a/service/create/templates.go
+++ b/service/create/templates.go
@@ -24,16 +24,16 @@ const (
 # gzip+base64 file (the final cloudconfig) from AWS S3 and run coreos-cloudinit
 # with it as an argument.
 
-# Wait for S3 domain to be available.
-s3_domain="s3.{{.Region}}.amazonaws.com"
-
-until nslookup $s3_domain; do
-    echo "waiting for $s3_domain to be available"
-    sleep 5
-done
-
 . /etc/environment
 USERDATA_FILE={{.MachineType}}
+
+# Wait for S3 bucket to be available.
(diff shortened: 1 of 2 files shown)


service/create/templates.go
```diff
@@ -27,6 +27,14 @@ const (
 . /etc/environment
 USERDATA_FILE={{.MachineType}}
 
+# Wait for S3 bucket to be available.
+s3_http_uri="https://s3.{{.Region}}.amazonaws.com/{{.S3URI}}/cloudconfig/$USERDATA_FILE"
+
+until [ $(curl --output /dev/null --silent --head --fail -w "%{http_code}" $s3_http_uri) -eq "403" ]; do
+   echo "waiting for $s3_http_uri to be available"
+   sleep 5
+done
```
rossf7: The check waits for a 403 from the S3 bucket to make sure its accessible. Improvements to my bash are very welcome! 😄

fgimenez: LGTM! :) Not sure if the script in which it is included has an overall timeout, if not I would include one here so that we don't wait forever, perhaps with a number of retries like this one  https://github.com/giantswarm/shipyard/blob/master/image/run-checks.sh#L11

rossf7:
@fgimenez Thanks, that's really nice. The script now exits after a number of retry attempts.

PR #463, opened 2017-10-23
migrate to circleci 2.0

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1853

Not sure if checking for `RELEASE_TOKEN` is needed, let me know :)


JosephSalisbury:
I didn't bother checking for the token, because we should break loudly if it's not there.

PR #468, opened 2017-10-26
Add alerts for orphan AWS resources

rossf7:
Towards giantswarm/giantswarm/issues/1953

This is the automated alert for orphan AWS resources. It runs as a new alerter service which is called on boot and starts a background timer that checks every 5 mins.

The main problem is our AWS account having 2 installs. So I'm tagging the VPCs with the installation name. Otherwise the operator would need to call the APIs for both installs and need endpoints, auth tokens etc.

I think having the tag on VPCs is useful for support anyway. We could do the same for Cloud Formation stacks. I don't think we should tag all the resources with the installation.

At the moment the installation name is hardcoded which needs to be fixed.

@JosephSalisbury WDYT? If its OK to tag the VPCs how can I get the installation name from the draughtsman config?


flag/service/aws/aws.go
```diff
@@ -8,4 +8,5 @@ type AWS struct {
 	AccessKey     accesskey.AccessKey
 	HostAccessKey accesskey.AccessKey
 	PubKeyFile    string
+	Region        string
```
rossf7: Region flag is needed by the alerter. For the operator it comes from the custom object.

resources/aws/vpc.go
```diff
@@ -97,6 +98,10 @@ func (v *VPC) CreateOrFail() error {
 				Key:   aws.String(tagKeyName),
 				Value: aws.String(v.Name),
 			},
+			{
+				Key:   aws.String(tagKeyInstallation),
+				Value: aws.String(v.InstallationName),
+			},
```
rossf7: This adds the new installation tag.

resources/aws/vpc.go
```diff
@@ -161,3 +166,32 @@ func (v VPC) GetID() (string, error) {
 
 	return *vpc.VpcId, nil
 }
+
+// List lists the VPCs for this installation.
+func (v VPC) List() ([]VPC, error) {
```
rossf7: List lists the VPCs with the installation tag.

service/alerter/aws.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// ListVpcs lists the VPCs in this installation and returns the clusterIDs
+// associated with them.
+func (s *Service) ListVpcs() ([]string, error) {
+	clusterIDs := []string{}
+
+	vpc := &awsresources.VPC{
+		// TODO Make configurable.
+		InstallationName: "gauss",
```
rossf7: TODO: Need to get this from the draughtsman config.

service/alerter/metric.go
```diff
@@ -0,0 +1,19 @@
...
+	orphanResourcesTotal = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "awsoperator",
+			Subsystem: "resources",
+			Name:      "orphan_resources_total",
+			Help:      "Number of AWS resources not associated with a cluster to be cleaned up.",
+		},
+		[]string{"resource"},
+	)
+)
```
rossf7: This is the new metric. There is a label per AWS resource type.

service/alerter/service.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/tpr"
+	"k8s.io/client-go/kubernetes"
+
+	awsutil "github.com/giantswarm/aws-operator/client/aws"
+	"github.com/giantswarm/aws-operator/service/key"
+)
+
+const (
+	alertIntervalMins = 5
```
rossf7: Alert runs every 5 mins. I went for a constant but this could be a flag.

service/alerter/service.go
```diff
@@ -0,0 +1,199 @@
...
+			select {
+			case <-alertChan:
+				err := s.OrphanResourcesAlert()
+				if err != nil {
+					s.logger.Log("error", fmt.Sprintf("could not execute orphan resources alert: '%#v'", err))
+				}
+			}
+		}
+	}()
+}
```
rossf7: Goroutine starts a ticker that calls the alert.

service/alerter/service.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	for _, resourceName := range resourceNames {
+		if ok, _ := clusters[resourceName]; !ok {
+			orphanResources = append(orphanResources, resourceName)
+		}
+	}
+
+	return orphanResources
+}
```
rossf7: This is the main logic for finding the orphan resources.

service/service.go
```diff
@@ -216,6 +235,10 @@ type Service struct {
 
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
+		// Start alerts to check for orphan resources.
+		s.Alerter.StartAlerts()
```
rossf7: Alerts are started before the operator is booted.

main.go
```diff
@@ -130,6 +130,9 @@ func main() {
 	daemonCommand.PersistentFlags().String(f.Service.AWS.HostAccessKey.ID, "", "ID of the AWS access key for the host cluster account. If empty, guest cluster account is used.")
 	daemonCommand.PersistentFlags().String(f.Service.AWS.HostAccessKey.Secret, "", "Secret of the AWS access key for the host cluster account. If empty, guest cluster account is used.")
 	daemonCommand.PersistentFlags().String(f.Service.AWS.HostAccessKey.Session, "", "Session token of the AWS access key for the host cluster account. If empty, guest cluster token is used.")
+	// TODO Remove hardcoded install name,
```
rossf7: TODO: Remove hardcoding and get from installations config.

resources/aws/vpc.go
```diff
@@ -161,3 +166,32 @@ func (v VPC) GetID() (string, error) {
...
+					aws.String(v.InstallationName),
+				},
+			},
+		},
+	})
+	if err != nil {
+		return []VPC{}, microerror.Mask(err)
+	}
+
+	for _, vpc := range resp.Vpcs {
```
JosephSalisbury: Why do you tag vpcs with the installation name after fetching them?

rossf7: We use the cluster ID as the Name tag for VPCs. So this is getting the cluster IDs rather than tagging with the installation name.

JosephSalisbury: Ah, I get you. Reading is hard.

JosephSalisbury:
If the installation names are not in the draughtsman config, you may need to add them.

examples/local/aws-operator-lab-chart/templates/configmap.yaml
```diff
@@ -10,6 +10,7 @@ data:
           id: "{{.Values.aws.accessKeyId}}"
           secret: "{{.Values.aws.secretAccessKey}}"
           session: "{{.Values.aws.sessionToken}}"
+        region: "{{.Values.aws.region}}"
```
fgimenez: It would be nice to have a default value in `examples/local/aws-operator-lab-chart/values.yaml`

service/alerter/service.go
```diff
@@ -0,0 +1,212 @@
...
+		for {
+			select {
+			case <-alertChan:
+				err := s.OrphanResourcesAlert()
+				if err != nil {
+					s.logger.Log("error", fmt.Sprintf("could not execute orphan resources alert: '%#v'", err))
+				}
+			}
+		}
+	}()
```
fgimenez: Wouldn't this goroutine leak? It would be nice to have a way to signal it to exit cleanly, maybe from `s`?

PR #473, opened 2017-10-30
Adds retries to VPC acceptance

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1943

We've had issues where the vpc was created, but acceptance did not
succeed. We think this is due to eventual consistency inside AWS
networking components, so some retrying should make it more reliable.

### DIFF
--- a/resources/aws/peering.go
+++ b/resources/aws/peering.go
@@ -6,7 +6,10 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/cenkalti/backoff"
+
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
 )
 
 const (
@@ -21,6 +24,8 @@ type VPCPeeringConnection struct {
 	PeerVPCId string // PeerVPCId the ID of the VPC in the host cluster.
 	id        string
 	AWSEntity
+
+	Logger micrologger.Logger
 }
 
 func (v VPCPeeringConnection) findExisting() (*ec2.VpcPeeringConnection, error) {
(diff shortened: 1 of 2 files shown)


resources/aws/peering.go
```diff
@@ -21,6 +24,8 @@ type VPCPeeringConnection struct {
 	PeerVPCId string // PeerVPCId the ID of the VPC in the host cluster.
 	id        string
 	AWSEntity
+
+	Logger micrologger.Logger
```
JosephSalisbury: I also added a logger, so we can get some visibility on the VPC acceptance retries.

service/create/service.go
```diff
@@ -694,6 +694,7 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
 			Clients:     clients,
 			HostClients: hostClients,
 		},
+		Logger: s.logger,
```
JosephSalisbury: The lack of a `New` function is a bit worrying, but I couldn't find any other struct creation locations.

resources/aws/peering.go
```diff
@@ -105,11 +110,19 @@ func (v *VPCPeeringConnection) CreateOrFail() error {
 
 	v.id = *vpcPeeringConnection.VpcPeeringConnection.VpcPeeringConnectionId
 
-	if _, err := v.HostClients.EC2.AcceptVpcPeeringConnection(
-		&ec2.AcceptVpcPeeringConnectionInput{
-			VpcPeeringConnectionId: aws.String(v.id),
-		},
-	); err != nil {
+	acceptOperation := func() error {
```
JosephSalisbury: We've only had issues with the accept call, so I only added the retry here. Adding the retry to create seemed excessive, at least at this point.

JosephSalisbury:
Tested on `gauss`, vpc peering connection was accepted. We don't have a good test case for the postmortem issue, as it's largely dependent on AWS networking internals, so I think we'll have to wait and see if it happens again.

resources/aws/peering.go
teemow: so @fgimenez pointed out to me why we are not adding those kind of comments to the code itself? i think it would be good to explain that to your future self :D

fgimenez: losing the fresh eyes sigh :)

JosephSalisbury: Totally agree with this. Comments on a PR should be comments in the code.

JosephSalisbury: Arguably, this conversation is only a git blame and some clicks away.

PR #474, opened 2017-10-30
Reuse existing ELBs and their DNS resource records

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2063

These changes will prevent the creation of a new ELB when a cluster is reprocessed, we were always creating new ones and updating the DNS entries even when they were already in place. Also, the proxy protocol is only assigned after creating a new ingress ELB, so that old clusters will keep the old configuration (the old ELBs will be reused and, because the ingress controller is not being created, the proxy policy will not be applied).

I've also included changes to the circleci pipeline, recent changes to architect make an empty project fail to build, and currently the in-cluster tests for this repo are empty, this build step will be readded when we have ready the first in-cluster test.

### DIFF
--- a/resources/aws/elb.go
+++ b/resources/aws/elb.go
@@ -56,16 +56,19 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
 		return false, microerror.Mask(clientNotInitializedError)
 	}
 
-	_, err := lb.findExisting()
-	if err == nil {
-		return false, nil
-	}
-
+	lbDescription, err := lb.findExisting()
 	if !strings.Contains(err.Error(), notFoundError.Error()) {
-
+		// error looking for elb
 		return false, err
 	}
 
+	if err == nil {
+		// elb found, initialize dns and hoested zone id
+		lb.setDNSFields(*lbDescription)
+		return false, nil
+	}
+
+	// elb not found, create
 	if err := lb.CreateOrFail(); err != nil {
 		return false, microerror.Mask(err)
 	}
--- a/resources/aws/elb.go
+++ b/resources/aws/elb.go
@@ -57,8 +57,13 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
 	}
 
 	_, err := lb.findExisting()
-	if err != nil && strings.Contains(err.Error(), notFoundError.Error()) {
+	if err == nil {
 		return false, nil
+	}
+
+	if !strings.Contains(err.Error(), notFoundError.Error()) {
+
+		return false, err
 	}
 
 	if err := lb.CreateOrFail(); err != nil {
--- a/resources/aws/elb.go
+++ b/resources/aws/elb.go
@@ -57,15 +57,16 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
 	}
 
 	lbDescription, err := lb.findExisting()
-	if !strings.Contains(err.Error(), notFoundError.Error()) {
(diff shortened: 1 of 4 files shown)


resources/aws/elb.go
```diff
@@ -57,14 +56,21 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
...
-		}
+	lbDescription, err := lb.findExisting()
 
+	if err == nil {
+		// elb found, initialize dns and hoested zone id
+		lb.setDNSFields(*lbDescription)
+		return false, nil
+	}
+
+	if !strings.Contains(err.Error(), notFoundError.Error()) {
```
JosephSalisbury: We can't compare the errors directly?

service/create/load_balancer.go
```diff
@@ -57,6 +57,13 @@ func (s *Service) createLoadBalancer(input LoadBalancerInput) (*awsresources.ELB
 
 	if lbCreated {
 		s.logger.Log("debug", fmt.Sprintf("created ELB '%s'", lb.Name))
+
+		// Assign the ProxyProtocol policy for ingress controller
+		if input.Name == input.Cluster.Spec.Cluster.Kubernetes.IngressController.Domain {
```
JosephSalisbury: I don't follow why this check is needed.

resources/aws/elb.go
```diff
@@ -57,14 +56,21 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
...
 	}
 
-	if err := lb.CreateOrFail(); err != nil {
-		if strings.Contains(err.Error(), awsclient.ELBConfigurationMismatch) {
-			return false, microerror.Mask(err)
-		}
-		if strings.Contains(err.Error(), awsclient.ELBAlreadyExists) {
-			return false, nil
-		}
+	lbDescription, err := lb.findExisting()
```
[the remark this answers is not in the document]
fgimenez: Indeed, very good point, fixed

resources/aws/elb.go
```diff
@@ -57,14 +56,21 @@ func (lb *ELB) CreateIfNotExists() (bool, error) {
...
 
+	if err == nil {
+		// elb found, initialize dns and hoested zone id
```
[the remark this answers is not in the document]
fgimenez: Done

resources/aws/elb.go
fgimenez: The actual error we are getting is `LoadBalancerNotFound: There is no ACTIVE Load Balancer named 'test-cluster-api'\n\tstatus code: 400, request id: aebfcf4d-bd87-11e7-89ff-f593dcba153f`, it includes the request id. Btw, instead of comparing with `notFoundError.Error()` we need to use here the `elb.ErrCodeAccessPointNotFoundException` constant from aws-sdk, it is the `LoadBalancerNotFound` part in the error string above. 

Do you think it would be better to use `strings.HasPrefix` instead of `strings.Contains`?

service/create/load_balancer.go
fgimenez: The proxy protocol policy can only be assigned to the ingress ELB, assigning it to the API ELB gives:
```
$ kubectl get pods
Unable to connect to the server: remote error: tls: record overflow
```

resources/aws/elb.go
JosephSalisbury: I think either is fine, I was just wondering if there was a 'proper' matcher.

PR #475, opened 2017-10-31
remove stale example files and move local to example

fgimenez:
The contents of example/local are moved to example and the stale templates at example/*yaml are removed.


fgimenez:
Thanks @rossf7 I've also updated the reference to the cert-operator lab now that it is merged.

PR #476, opened 2017-10-31
Move runMachines and deleteMachines logic to machine.go

rossf7:
Towards giantswarm/giantswarm/issues/2023

This is some restructuring to make it easier to migrate the create service to the legacy operatorkit resource. No logic is changed is changed it just extracts the logic to machine.go.


service/create/service.go
```diff
@@ -1686,3 +1401,16 @@ func (s *Service) updateFunc(oldObj, newObj interface{}) {
...
+		Body:          strings.NewReader(data),
+		Bucket:        aws.String(s3Bucket),
+		Key:           aws.String(path),
+		ContentLength: aws.Int64(int64(len(data))),
+	}); err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
```
rossf7: This method is moved after the func handlers. Again as prep to make the migration easier.

PR #477, opened 2017-10-31
Extract delete logic to processDelete

rossf7:

This PR is more prep for creating the legacy resource. It extracts the delete logic to a processDelete method. No logic is changed its just restructuring.

### DIFF
--- a/service/create/service.go
+++ b/service/create/service.go
@@ -223,6 +223,41 @@ func (s *Service) addFunc(obj interface{}) {
 	s.logger.Log("info", fmt.Sprintf("cluster '%s' processed", key.ClusterID(cluster)))
 }
 
+func (s *Service) deleteFunc(obj interface{}) {
+	// We can receive an instance of awstpr.CustomObject or cache.DeletedFinalStateUnknown.
+	// We need to assert the type properly and log an error when we cannot do that.
+	// Also, the cache.DeleteFinalStateUnknown object can contain the proper CustomObject,
+	// but doesn't have to.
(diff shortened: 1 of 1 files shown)


service/create/service.go
```diff
@@ -223,6 +223,41 @@ func (s *Service) addFunc(obj interface{}) {
 	s.logger.Log("info", fmt.Sprintf("cluster '%s' processed", key.ClusterID(cluster)))
 }
 
+func (s *Service) deleteFunc(obj interface{}) {
```
rossf7: The deleteFunc now just calls processDelete to delete the resources.

service/create/service.go
```diff
@@ -925,35 +960,9 @@ func (s *Service) processCluster(cluster awstpr.CustomObject) error {
...
-			return
-		}
-		cluster = *clusterPtr
-	}
-
+func (s *Service) processDelete(cluster awstpr.CustomObject) error {
 	if err := validateCluster(cluster); err != nil {
-		s.logger.Log("error", "cluster spec is invalid: '%#v'", err)
-		return
+		return microerror.Maskf(executionFailedError, fmt.Sprintf("cluster spec is invalid: '%#v'", err))
```
rossf7: In this error and the next we want to stop execution as something is wrong. Otherwise we continue deleting resources.

service/create/service.go
```diff
@@ -1356,7 +1364,7 @@ func (s *Service) deleteFunc(obj interface{}) {
 		s.logger.Log("info", "deleted keypair")
 	}
 
-	s.logger.Log("info", fmt.Sprintf("cluster '%s' deleted", key.ClusterID(cluster)))
+	return nil
```
rossf7: This log line moves to the deleteFunc.

PR #478, opened 2017-10-31
Move updateFunc with other handlers

rossf7:
Last one of these restructuring PRs, thanks for bearing with me!

Moves the updateFunc with the addFunc and deleteFunc. These will become ProcessCreateState, etc in the Legacy resource so the diffs are cleaner for the migration.


rossf7:
Yes this was just moving it to make the move to the Legacy resource easier.

PR #479, opened 2017-10-31
e2e test for CloudFormation resource existence

fgimenez:

Updates the migration branch and enables an in-cluster test that checks for the TPR, currently failing. Any feedback is more than welcome, especially in the e2e test (e2e directory).


fgimenez:
Pushed changes for the operator to create the cloudformation resource, the sonobuoy test is passing, this is the execution log https://circleci.com/gh/giantswarm/aws-operator/1243 and this is how the test looks like https://github.com/giantswarm/aws-operator/pull/479/commits/3e60f42286c5374b130ea7f71c46f9a7e52e7300#diff-a22d92e3b0ec4ef4f4f8394565cafa4c

I've added also changes to save the junit test results as circleci artifacts https://1243-80619275-gh.circle-artifacts.com/0/home/circleci/project/.e2e-harness/workdir/plugins/e2e/results.xml, about the junit format, believe it or not, it is the default for upstream e2e tests... https://github.com/kubernetes/kubernetes/blob/master/test/e2e/e2e.go#L370

e2e/k8s/k8s.go
```diff
@@ -0,0 +1,20 @@
+package k8s
+
+import (
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func Client() (*kubernetes.Clientset, error) {
+	// creates the in-cluster config
```
rossf7: Comments nitpick but I'd like these to be sentences with a capital and full stop.

e2e/main.go
```diff
@@ -0,0 +1,29 @@
...
+func main() {
+	// architect requires this
+	if len(os.Args) > 1 {
+		if os.Args[1] == "version" {
+			fmt.Println("0.1.0")
+			return
+		}
+
+		if os.Args[1] == "--help" {
+			fmt.Println("Yet another lookout")
```
rossf7: What is this logging for?

e2e/tests/cloudformation.go
```diff
@@ -0,0 +1,39 @@
...
+		clientset: cs,
+		logger:    logger,
+	}
+	Add(ts.TestCRExists)
+}
+
+func (ts *TestSet) TestCRExists() (string, error) {
+	desc := "cloudformation TPR exists"
+
+	_, err := ts.clientset.ExtensionsV1beta1().ThirdPartyResources().Get("cloudformation.giantswarm.io", metav1.GetOptions{})
```
rossf7: I don't think this TPR is correct. The cloud formation resource will still process awstpr but it will do CRUD operations on the Cloud Formation stack.

service/operator/operator.go
```diff
@@ -79,12 +81,29 @@ func New(config Config) (*Operator, error) {
...
 		c.Version = awstpr.VersionV1
 
-		newTPR, err = tpr.New(c)
+		awsTPR, err = tpr.New(c)
 		if err != nil {
 			return nil, microerror.Maskf(err, "creating TPR util for "+awstpr.Name)
 		}
 	}
 
+	var cfTPR *tpr.TPR
```
rossf7: As above there won't be a Cloud Formation TPR we just need awstpr.

e2e/k8s/k8s.go
fgimenez: Sure, done :)

e2e/tests/cloudformation.go
fgimenez: Ok, sorry missed that from the spec, I've modified it to check for awsptr being defined.

service/operator/operator.go
fgimenez: Yep, fixed

e2e/main.go
fgimenez: For nothing in fact! :) thanks a lot for bringing this up, i just c&p'ed this from my hiring task :) Removed, much cleaner now

fgimenez:
Closing, the in-cluster test is proposed in #482 and the resource changes will be targeted to a new branch.

PR #483, opened 2017-11-03
vendor-k8scc-changes-0311

calvix:
merging latest changes in k8scc
ping @r7vme @JosephSalisbury @kopiczko


calvix:
that is a date of today so 03.11

PR #484, opened 2017-11-06
[WIP] implement GetDesiredState for the cloudformation resource

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2023, https://github.com/giantswarm/giantswarm/issues/1918

This should land after the actual migration to operatorkit happens (AIUI it wouldn't hurt to have it in place, but doesn't make too much sense IMO).


client/aws/aws.go
```diff
@@ -62,6 +63,16 @@ func NewClients(config Config) Clients {
 	return clients
 }
 
+func NewCFClient(config Config) cloudformationiface.CloudFormationAPI {
```
rossf7: NewClients already returns a cloudformation client. Why do we need to use `cloudformation/cloudformationiface`?

service/resource/cloudformation/main_stack.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	awsCF "github.com/aws/aws-sdk-go/service/cloudformation"
+	"github.com/giantswarm/aws-operator/service/key"
+	"github.com/giantswarm/awstpr"
+)
+
+func newMainStack(customObject awstpr.CustomObject) (*awsCF.CreateStackInput, error) {
+	clusterId := key.ClusterID(customObject)
+	stackName := fmt.Sprintf("cluster-main-stack-%s", clusterId)
```
rossf7: I think we should add a helper to the key service for the stack name. Also I think the name should start with the clusterID so that stacks are ordered by cluster in the CLI / console.

service/resource/cloudformation/resource.go
```diff
@@ -0,0 +1,89 @@
...
+// resource by best effort.
+func DefaultConfig() Config {
+	return Config{
+		// Dependencies.
+		AwsClient: nil,
+		Logger:    nil,
+	}
+}
+
+// Resource implements the config map resource.
```
rossf7: Nitpick, resource name should be cloud formation.

service/resource/cloudformation/resource.go
```diff
@@ -0,0 +1,89 @@
...
+		awsClient: config.AwsClient,
+		logger: config.Logger.With(
+			"resource", Name,
+		),
+	}
+
+	return newService, nil
+}
+
+func toCustomObject(v interface{}) (awstpr.CustomObject, error) {
```
rossf7: toCustomObject could be added to the key service. We've done that in newer operators.

client/aws/aws.go
fgimenez: Indeed, this was a leftover, this interface is used as the type of one of the members of the Resource (in order to be able to unit test it with a mock, probably we can use a more ad-hoc interface, let's see), it doesn't belong here, fixed.

service/resource/cloudformation/main_stack.go
fgimenez: Good idea thx, done

service/resource/cloudformation/resource.go
fgimenez: Yep! :) fixed

service/resource/cloudformation/resource.go
fgimenez: Done

service/resource/cloudformation/#main_stack.go#
```diff
@@ -0,0 +1,21 @@
+package cloudformation
```
rossf7: These temp files should be removed.

service/resource/cloudformation/desired_test.go
```diff
@@ -0,0 +1,66 @@
...
+			obj: &awstpr.CustomObject{
+				Spec: awstpr.Spec{
+					Cluster: clustertpr.Spec{
+						Cluster: spec.Cluster{
+							ID: "5xchu",
+						},
+					},
+				},
+			},
+			expectedName: "cluster-main-stack-5xchu",
```
rossf7: Should the test use the new name format?

service/resource/cloudformation/#main_stack.go#
fgimenez: yep, done

service/resource/cloudformation/desired_test.go
fgimenez: Of course, but i'm not sure if we should be testing the name here at all, we have already in place a unit test for the new key method that gives the name, maybe test that the key method is called? We can discuss in the new PR.

fgimenez:
Thx @rossf7 I've added here the current asg template as a string constant, and one method for executing the template using the custom object as data (being optimistic here :D). It is not being used now, we can discuss this approach too when these changes are proposed to the new migration branch.

fgimenez:
Closing in favor of #489

PR #488, opened 2017-11-07
Adds support for creating master service

JosephSalisbury:


service/create/master_service.go
```diff
@@ -0,0 +1,114 @@
...
+	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/pkg/api/v1"
+
+	"github.com/giantswarm/microerror"
+)
+
+type MasterServiceInput struct {
+	Clients   awsutil.Clients
+	ClusterID string
```
JosephSalisbury: It's probably nicer to pass the whole tpr in here

service/create/master_service.go
```diff
@@ -0,0 +1,114 @@
...
+					},
+				},
+			},
+		})
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if len(reservations.Reservations) < 1 {
+			return microerror.Maskf(executionFailedError, "could not find master reservation: %s", masterID)
```
JosephSalisbury: We can use a better error here

JosephSalisbury:
@fgimenez @rossf7 i haven't tested out here, but I want input if you guys think this is the right direction. Considerate it an informal review :D

service/create/master_service.go
```diff
@@ -0,0 +1,116 @@
...
+		reservations, err := input.Clients.EC2.DescribeInstances(&ec2.DescribeInstancesInput{
+			Filters: []*ec2.Filter{
+				{
+					Name: aws.String("instance-id"),
+					Values: []*string{
+						aws.String(masterID),
+					},
+				},
+			},
+		})
```
rossf7: We're getting rid of the old AWS resources but it looks like it supports finding instances. Could you use that?

https://github.com/giantswarm/aws-operator/blob/e796869c17eb3a40322c3197463adf529b8d5cdd/resources/aws/instance.go#L237

service/create/master_service.go
```diff
@@ -0,0 +1,116 @@
...
+		masterInstance := reservations.Reservations[0].Instances[0]
+
+		service := v1.Service{
+			ObjectMeta: apismetav1.ObjectMeta{
+				Name:      fmt.Sprintf("master-%s", masterID),
+				Namespace: key.ClusterID(input.Cluster),
+				Labels: map[string]string{
+					"app":      "master",
+					"cluster":  key.ClusterID(input.Cluster),
+					"customer": key.ClusterID(input.Cluster),
```
rossf7: Shouldn't this be the customer ID?

service/create/master_service.go
JosephSalisbury: Yeah sweet, I didn't look if we had a good way for this already. Will update.

service/create/master_service.go
JosephSalisbury: Lol, yeah, it should be :D Will update.

JosephSalisbury:
Yeah, I assumed that the k8s stuff would turn into separate resources. So like, service resource that fetches the master private IP, reconciles the service.

rossf7:
Yes the service resource can handle that which is nice! 😄 

@fgimenez  FYI I've added to our checklist in giantswarm/giantswarm#2023

JosephSalisbury:
@rossf7 mind having another look? i'm using `FindInstances` now, and have fixed up the customer tagging

PR #492, opened 2017-11-08
Set the size of the workers ASG if it exists

rossf7:

This PR sets the size of the worker autoscaling group if it already exists. This replaces the previous updateFunc logic and means we can still scale workers via the API.

### DIFF
--- a/service/resource/legacy/resource.go
+++ b/service/resource/legacy/resource.go
@@ -936,7 +936,19 @@ func (s *Resource) processCluster(cluster awstpr.CustomObject) error {
 	if asgCreated {
 		s.logger.Log("info", fmt.Sprintf("created auto scaling group '%s' with size %v", asg.Name, key.WorkerCount(cluster)))
 	} else {
-		s.logger.Log("info", fmt.Sprintf("auto scaling group '%s' already exists, reusing", asg.Name))
+		// If the cluster exists set the worker count so the cluster can be scaled.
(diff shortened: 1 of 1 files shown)


service/resource/legacy/resource.go
```diff
@@ -936,7 +893,19 @@ func (s *Resource) processCluster(cluster awstpr.CustomObject) error {
...
+		scaleWorkers := awsresources.AutoScalingGroup{
+			Client:  clients.AutoScaling,
+			Name:    key.AutoScalingGroupName(cluster, prefixWorker),
+			MinSize: key.WorkerCount(cluster),
+			MaxSize: key.WorkerCount(cluster),
+		}
+
+		if err := scaleWorkers.Update(); err != nil {
+			s.logger.Log("error", fmt.Sprintf("%#v", err))
+		}
```
rossf7: Here we update the ASG. The ASG is only affected if the size has changed.

service/resource/legacy/resource.go
```diff
@@ -240,49 +240,6 @@ func (s *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desire
 }
 
 func (s *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange interface{}) error {
-	// TODO The legacy resource needs to support scaling workers via NewUpdatePatch.
```
rossf7: Removing the old updateFunc which was already commented out.

PR #504, opened 2017-11-13
Update to master

fgimenez:

Pulls latest changes from master


rossf7:
The installation name also needs to be passed to the legacy resource so it can tag the VPC. Just pushed a fix.

service/service.go
```diff
@@ -254,7 +276,7 @@ func New(config Config) (*Service, error) {
 		frameworkConfig.BackOff = frameworkBackOff
 		frameworkConfig.InitCtxFunc = initCtxFunc
 		frameworkConfig.Logger = config.Logger
-		frameworkConfig.Resources = resources
+		frameworkConfig.ResourceRouter = framework.NewDefaultResourceRouter(resources)
```
rossf7: Here we use the new ResourceRouter. It's not fully implemented yet so all resources are returned.

vendor/github.com/giantswarm/k8scloudconfig/v_0_1_0/master_template.go
```diff
@@ -854,11 +855,12 @@ write_files:
           serviceAccountName: nginx-ingress-controller
           containers:
           - name: nginx-ingress-controller
-            image: {{.Cluster.Kubernetes.IngressController.Docker.Image}}
+            image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.9.0-beta.17
```
rossf7: Can you confirm we want these ingress changes to be vendored?

fgimenez: Very good point, checked with @puja108 and it's ok to have this in place for new clusters.

PR #505, opened 2017-11-14
fix-var-lib-mount3

calvix:
towards https://github.com/giantswarm/giantswarm/issues/2049

Looks like previous PR was not enough, this is still blocking formatting the hdd with xfs for second start


calvix:
Yeah looks like it, with workers on AWS we don't reboot as scaling group always start a new one in case one is stopped.

PR #507, opened 2017-11-16
Extend master policy to allow ELB management

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2111

Tested on gauss


fgimenez:
@rossf7 exactly! :D also this page from aws docs http://docs.aws.amazon.com/en_en/elasticloadbalancing/latest/userguide/load-balancer-authentication-access-control.html

PR #516, opened 2017-11-23
introduce adaptor and implemented first fields

fgimenez:

This PR mainly proposes an implementation of the adaptor pattern for the stack's model layer, I've tried to make it easy to add new components (both template and related adaptor logic) and test separately the adaptor logic concerned with each of them, Ross please let me know WDYT. 

I've also modified the AWS client of the cloudformation resource, now includes clients for more kind of resources, AIUI given that the operator executes the reconciliation of resources sequentially (first legacy, then, unrelated, cloudformation), during the cloudformation stage we would need to query the AWS API to get references of resources created in the legacy stage, pls correct me if I'm wrong.

There are only implemented a few fields as an example, once we agree with the approach I'll continue with them.


service/resource/cloudformation/adaptor.go
```diff
@@ -0,0 +1,85 @@
...
+
+	awsutil "github.com/giantswarm/aws-operator/client/aws"
+)
+
+type hydrater func(awstpr.CustomObject, awsutil.Clients) error
+
+type adaptor struct {
+	ASGType string
+
+	// worker launch configuration
```
rossf7: I like the adaptor structure but I wonder how it will scale as we add more resources. Some attributes like SecurityGroupID will be needed multiple times and will be different for master and workers.

So I'm worried about collisions. Will we have multiple adaptors or can we nest structs within this adaptor struct?

service/resource/cloudformation/adaptor.go
```diff
@@ -0,0 +1,85 @@
...
+
+	var g errgroup.Group
+
+	hydraters := []hydrater{
+		a.getAutoScalingGroup,
+		a.getLaunchConfiguration,
+	}
+
+	// each hydrater works on different data and may need to query AWS API, so that
+	// they are run in separate goroutines
```
rossf7: For a start we'll need to query the AWS API to get data like the subnet IDs. But as we move more resources to CF we should be able to use references.

Having separate goroutines is good but if it's for the API calls then I think that's a temporary state. Lastly very nitpicky but can we make the comments full sentenences.

service/resource/cloudformation/create.go
```diff
@@ -5,8 +5,9 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	awscloudformation "github.com/aws/aws-sdk-go/service/cloudformation"
-	"github.com/giantswarm/aws-operator/service/key"
 	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/aws-operator/service/key"
```
rossf7: <3

service/resource/cloudformation/template.go
```diff
@@ -1,18 +1,22 @@
...
 const (
-	MainTemplate = `AWSTemplateFormatVersion: 2010-09-09
-Description: {{ .ASGType }} autoscaling group
+	MainTemplate = `{{define "main"}}AWSTemplateFormatVersion: 2010-09-09
+Description: Main CloudFormation stack.
 Resources:
-  {{ .ASGType }}LaunchConfiguration:
+  {{template "launch_configuration" .}}
+  {{template "autoscaling_group" .}}
+{{end}}`
```
rossf7: These nested templates in the main stack are really nice.

service/resource/cloudformation/template.go
```diff
@@ -1,18 +1,22 @@
...
+
+	LaunchConfigurationTemplate = `{{define "launch_configuration"}}{{ .ASGType }}LaunchConfiguration:
```
rossf7: Using go files for the templates is good for now. But I think it would be good to switch to separate template files as it makes them easier to edit. WDYT?

The downside with using template files is we need to add them to the docker image.

service/resource/cloudformation/template.go
```diff
@@ -1,18 +1,22 @@
...
     Type: "AWS::AutoScaling::LaunchConfiguration"
     Description: {{ .ASGType }} launch configuration
     Properties:
-      ImageId: !Ref ImageID
+      ImageId: {{ .ImageID }}
```
rossf7: This templating in the resources is nice and much less cumbersome than all the parameters <3

service/resource/cloudformation/adaptor.go
fgimenez: Totally, I started having separate types for each component, and then a main adaptor type that embedded them anonymously, so that all the fields were available when doing the template execution.

IMO this approach is better than having multiple adaptors, that would lead us to execute separately each template and then merge the resulting buffers, i find it more natural to have this last stage in a single step (one single template execution with one single data struct), let me know WDYT.

service/resource/cloudformation/adaptor.go
fgimenez: Agree, let's remove the `errGroup` for now until it is actually needed.

service/resource/cloudformation/template.go
fgimenez: Very nice, I was thinking of having separate go files for each resource but it was kind of weird, I'll prepare the template files and the changes to the Dockerfile to add them to the image.

service/resource/cloudformation/adaptor.go
rossf7: Yes I think having a single adaptor with per component types is good. I also prefer a single template execution.

service/resource/cloudformation/template.go
rossf7: Great!  I think this is much better than having huge template files like in k8scloudconfig.

service/resource/cloudformation/adaptor.go
fgimenez: Done, now we have separate types for each template, they are embedded as anonymous members in the main adapter so that the fields are accessible from the included templates and, more importantly, we let the compiler help us with ambiguous fields. 

About those common fields, I think we can promote them to the top level adapter type, so that we create them once and can be used in all the embedded templates, WDYT?

service/resource/cloudformation/template.go
fgimenez: Done, in fact the Dockerfile already had it in place :) I've added a relative reference in `main_stack.go`, it works fine in the unit tests, will do a quick check on the container but should work the same.

fgimenez:
Additionally I've added these changes:
* adaptor -> adapter, it's the proper name of the pattern ;)
* the main adapter is now created through a `new` function (we could extract them to their own package if things grow too much), the hydraters are kept as methods.

resources/templates/cloudformation/auto_scaling_group.yaml
```diff
@@ -1,87 +1,21 @@
...
         # minimum amount of instances that must always be running during a rolling update
-        MinInstancesInService: 2
+        MinInstancesInService: {{ .MinInstancesInService }}
         # only do a rolling update of this amount of instances max
-        MaxBatchSize: 2
+        MaxBatchSize: {{ .MaxBatchSize }}
         # after creating a new instance, pause operations on the ASG for this amount of time
-        PauseTime: PT10S
+        PauseTime: {{ .RollingUpdatePauseTime }}
+{{end}}
```
rossf7: I like the new template files but could you change the path to `service/templates/cloudformation`? We want to remove the resources dir and have everything under `service/` to match the other operators.

Also later we'll need to support multiple versions of the templates. Once we've implemented versioning the Cloud Formation resource in operatorkit we could use the same version in the templates dir. WDYT?

service/resource/cloudformation/adapter.go
```diff
@@ -0,0 +1,81 @@
...
+
+type lauchConfigAdapter struct {
+	ImageID                  string
+	SecurityGroupID          string
+	InstanceType             string
+	IAMInstanceProfileName   string
+	BlockDeviceMappings      []BlockDeviceMapping
+	AssociatePublicIPAddress bool
+	SmallCloudConfig         string
+}
```
rossf7: Nitpick but I think having these ordered alphabetically would be good. We found that useful in the old style AWS resources.

service/resource/cloudformation/adapter.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	a.ImageID = customObject.Spec.AWS.Workers[0].ImageID
+	a.InstanceType = customObject.Spec.AWS.Workers[0].InstanceType
+
+	return nil
+}
+
+func (a *adapter) getAutoScalingGroup(customObject awstpr.CustomObject, clients awsutil.Clients) error {
+	a.AZ = customObject.Spec.AWS.Region
```
rossf7: This should be AZ rather than Region. AFAIK Region is just used for creating the AWS client.

At the moment we only support a single AZ but supporting multiple AZs is something we want to support later.

resources/templates/cloudformation/auto_scaling_group.yaml
fgimenez: Sure, makes perfect sense. I've also modified the Dockerfile and tested it, worked but putting the resources out of `/opt` so that the relative path works both in the container and in unit test runs, pls take look.

service/resource/cloudformation/adapter.go
fgimenez: Done

service/resource/cloudformation/adapter.go
fgimenez: yep, fixed :)

PR #517, opened 2017-11-24
Adds key rotation enabled by default

pipo02mix:
I would like to open this PR in order to enable KMS rotation as a requirement of some customers (https://github.com/giantswarm/vodafone/issues/267). It is my first Go contribution so I am glad to listen to opinions


resources/aws/kms.go
```diff
@@ -55,6 +55,12 @@ func (kk *KMSKey) CreateOrFail() error {
 		return microerror.Mask(err)
 	}
 
+	if _, err := kk.Clients.KMS.EnableKeyRotation(&kms.EnableKeyRotationInput{
+		KeyId: aws.String(*key.KeyMetadata.KeyId),
```
rossf7: You can use `key.KeyMetadata.KeyId` here without the `aws.String` helper. As `KeyId` is already a string pointer.

pipo02mix: ok thx for the comment

PR #519, opened 2017-11-24
workers' cloudformation asg template: launch config

fgimenez:

Added the missing fields of launch configuration, IMO the meat of this PR is the definition of the interfaces for the AWS clients, each of them as small as possible. This decouples us from the foreign API and allows us to do proper unit testing.

I've tried to reuse as possible code from the legacy libraries, but the dependency on external types makes it difficult. Regarding the actual retrieval of the required fields I've tried to prevent premature optimizations, maybe the code that gets the IAM user is only used once and given its simplicity we don't need to extract a function for it.

About the `SmallCloudConfig`, the function that generates it was a method of a `Service` object, but the receiver was not being used, I've changed it to be a simple function.

I think that we could write more `key` functions for accessing the custom object fields, we can iterate on that.


service/resource/cloudformation/adapter_test.go
```diff
@@ -64,14 +76,27 @@ func TestAdapterLaunchConfigurationRegularFields(t *testing.T) {
...
 		},
 	}
-
-	clients := awsutil.Clients{}
-	a := adapter{}
 	for _, tc := range testCases {
+		clients := Clients{
+			EC2: &eC2ClientMock{sgExists: true},
+			IAM: &iAMClientMock{},
+		}
```
rossf7: I like the test coverage and the mocking approach.

service/resource/cloudformation/mocks.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/iam"
+)
+
+type eC2ClientMock struct {
+	sgExists bool
+	sgID     string
+}
+
+func (e *eC2ClientMock) DescribeSecurityGroups(input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
```
rossf7: Having all the mocks in a single file is nice. I also like that the interface is small and DescribeSecurityGroups can be removed later and we can use a ref in the CF template.

service/resource/cloudformation/adapter.go
```diff
@@ -63,18 +71,77 @@ func newAdapter(customObject awstpr.CustomObject, clients awsutil.Clients) (adap
...
+	}
+	l.SecurityGroupID = *output.SecurityGroups[0].GroupId
+
+	// cloud config
+	resp, err := clients.IAM.GetUser(&iam.GetUserInput{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	userArn := *resp.User.Arn
+	accountID := strings.Split(userArn, ":")[4]
```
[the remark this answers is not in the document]
fgimenez: Sure, very good point, I've moved the code from client to the resource package and updated the references from legacy, let me know if that makes sense to you.

service/resource/cloudformation/adapter.go
```diff
@@ -63,18 +71,77 @@ func newAdapter(customObject awstpr.CustomObject, clients awsutil.Clients) (adap
...
+	clusterID := key.ClusterID(customObject)
+	s3URI := fmt.Sprintf("%s-g8s-%s", accountID, clusterID)
+
+	cloudConfigConfig := legacy.SmallCloudconfigConfig{
+		MachineType: prefixWorker,
+		Region:      customObject.Spec.AWS.Region,
+		S3URI:       s3URI,
+	}
+	smallCloudConfig, err := legacy.SmallCloudconfig(cloudConfigConfig)
```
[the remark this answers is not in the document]
fgimenez: Indeed, moved to the cloudformation resource package too.

fgimenez:
Thanks a lot for your review Ross, the small cloudconfig template makes perfect sense. I've moved code  from the legacy resource to cloudformation, let me know if we should put it elsewhere.

service/resource/cloudformation/adapter.go
```diff
@@ -63,19 +75,117 @@ func newAdapter(customObject awstpr.CustomObject, clients awsutil.Clients) (adap
 	return a, nil
 }
 
-func (a *adapter) getLaunchConfiguration(customObject awstpr.CustomObject, clients awsutil.Clients) error {
+func (l *lauchConfigAdapter) getLaunchConfiguration(customObject awstpr.CustomObject, clients Clients) error {
```
rossf7: Typo on launchConfig

service/templates/cloudconfig/small_cloudconfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+done
+
+/usr/bin/rkt run \
+    --net=host \
+    --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf  \
+    --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos \
+    --trust-keys-from-https \
+    quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600 -- aws s3 --region {{.Region}} cp s3://{{.S3URI}}/cloudconfig/$USERDATA_FILE /var/run/coreos/temp.txt
+base64 -d /var/run/coreos/temp.txt | gunzip > /var/run/coreos/$USERDATA_FILE
+exec /usr/bin/coreos-cloudinit --from-file /var/run/coreos/$USERDATA_FILE
```
rossf7: Really nice that this is a template now <3

service/resource/cloudformation/adapter.go
fgimenez: Good catch, fixed!

fgimenez:
> On moving code to the legacy resource I think it makes sense in this case. So we only have a single template. But the legacy resource is end of life now. So in some cases it may not make sense to update it.

Indeed, in this case all the code has moved from legacy to cloudformation, I totally agree with you, this should be the direction we should follow, adapt legacy to use code from cf and only continue adding code to legacy when strictly necessary.

PR #522, opened 2017-11-27
Add Namespace resource and implement Current and Desired state

rossf7:

Adds the Namespace resource to create a k8s namespace per guest cluster. This will be used for the k8s Service resource needed for guest cluster scraping. Just the Current and Desired states are implemented for now to keep the PR smaller.

The resource is basically the same as in `kvm-operator` but it doesn't cancel the reconciliation if the namespace is already terminating. See [my comment](https://github.com/giantswarm/aws-operator/pull/522#discussion_r153208507) on that.


service/resource/namespace/current.go
```diff
@@ -0,0 +1,38 @@
...
+		if apierrors.IsNotFound(err) {
+			r.logger.LogCtx(ctx, "debug", "did not find the namespace in the Kubernetes API")
+			// fall through
+		} else if err != nil {
+			return nil, microerror.Mask(err)
+		} else {
+			r.logger.LogCtx(ctx, "debug", "found the namespace in the Kubernetes API")
+			namespace = manifest
+		}
+	}
```
rossf7: This is the same as in kvm-operator expect it doesn't cancel the reconciliation if the namespace is already terminating.

I think this is a difference between KVM and AWS. I want to be sure that the Cloud Formation resource processes the reconciliation and the CF stack gets deleted.

https://github.com/giantswarm/kvm-operator/blob/b49f98529108b39676a214971f420b54c203a0b0/service/resource/namespace/current.go#L41

service/key/key.go
```diff
@@ -23,10 +23,18 @@ func AvailabilityZone(customObject awstpr.CustomObject) string {
 	return customObject.Spec.AWS.AZ
 }
 
+func ClusterCustomer(customObject awstpr.CustomObject) string {
+	return customObject.Spec.Cluster.Customer.ID
+}
```
rossf7: ClusterCustomer is a bit confusing because it's called Customer in `clustertpr` but its actually the Organisation in Happa / gsctl. This is used for labeling the k8s namespace and I think the label is useful.

PR #525, opened 2017-11-27
added autoscaling group adapter fields

fgimenez:

Moved some more code from legacy for getting the load balancer name, subnet methods are more rooted. Also extracted some code from the adapter file.


service/key/key.go
```diff
@@ -55,6 +55,10 @@ func HasClusterVersion(customObject awstpr.CustomObject) bool {
 	}
 }
 
+func IngressDomain(customObject awstpr.CustomObject) string {
+	return customObject.Spec.Cluster.Kubernetes.IngressController.Domain
+}
```
rossf7: I like that you've created a new key function for this.

IMO we should create a key function if we think a value is needed multiple times. If we only need it once I'm OK with getting the value from the custom object.

service/resource/cloudformation/adapter.go
```diff
@@ -151,41 +149,46 @@ func (l *launchConfigAdapter) getLaunchConfiguration(customObject awstpr.CustomO
...
+	a.ASGMaxSize = workers
+	a.ASGMinSize = workers
+	a.MaxBatchSize = strconv.FormatFloat(asgMaxBatchSizeRatio, 'f', -1, 32)
+	a.MinInstancesInService = strconv.FormatFloat(asgMinInstancesRatio, 'f', -1, 32)
+	a.HealthCheckGracePeriod = gracePeriodSeconds
+	a.RollingUpdatePauseTime = rollingUpdatePauseTime
+
+	// load balancer name
+	// TODO: remove this code once the ingress load balancer is created by cloudformation
+	// and add a reference in the template
```
rossf7: Like these comments so it's clear this is part of the transition to CF.

service/resource/cloudformation/iam.go
```diff
@@ -0,0 +1,22 @@
...
+	case accountID == "":
+		return microerror.Mask(emptyAmazonAccountIDError)
+	case len(accountID) != accountIDLength:
+		return microerror.Mask(wrongAmazonAccountIDLengthError)
+	case !r.MatchString(accountID):
+		return microerror.Mask(malformedAmazonAccountIDError)
+	}
+
+	return nil
+}
```
rossf7: <3

service/resource/cloudformation/load_balancer.go
```diff
@@ -0,0 +1,44 @@
...
+
+	componentName, err := componentName(domainName)
+	if err != nil {
+		return "", microerror.Maskf(malformedCloudConfigKeyError, "spec.cluster.cluster.id")
+	}
+
+	lbName := fmt.Sprintf("%s-%s", key.ClusterID(cluster), componentName)
+
+	return lbName, nil
+}
```
rossf7: I think this should be in the key package.

service/resource/cloudformation/load_balancer_test.go
```diff
@@ -1,4 +1,4 @@
-package legacy
+package cloudformation
```
rossf7: I don't think we should move logic from the legacy package to the cloud formation resource. I think it is better to treat legacy as frozen and add the new code to the operatorkit resources.

IMO some minor changes for enabling / disabling logic are OK. In this case we could move LoadBalancerName to the key package.

service/resource/legacy/load_balancer.go
```diff
@@ -35,7 +35,7 @@ type LoadBalancerInput struct {
 }
 
 func (s *Resource) createLoadBalancer(input LoadBalancerInput) (*awsresources.ELB, error) {
-	lbName, err := loadBalancerName(input.Name, input.Cluster)
+	lbName, err := cloudformation.LoadBalancerName(input.Name, input.Cluster)
```
rossf7: If we move LoadBalancerName to the key package I'd be fine with updating it here. Although IMO leaving the logic as is would also be OK.

service/key/key.go
fgimenez: Makes sense, fixed

service/resource/cloudformation/load_balancer.go
fgimenez: Sure, moved.

service/resource/cloudformation/load_balancer_test.go
fgimenez: Yes, but IMO when possible it would be better to have only one code path, if we duplicate code during the transition it can be hard to track problems. In this case I've moved the logic that generates the load balancer name to the key package as you suggest later and that code is used for both legacy and cloudformation, let me know WDYT.

service/resource/legacy/load_balancer.go
fgimenez: Yep, moved to the key package.

PR #527, opened 2017-11-28
update local lab readme

fgimenez:
Addresses latest changes in cert-operator lab.


examples/README.md
```diff
@@ -21,19 +20,21 @@ setup. See [this guide][cert-operator-local-setup] for details. The operator and
 to be used during aws-operator setup can be installed with:
 
 ```bash
+helm registry install quay.io/giantswarm/vaultlab-chart:stable -- \
```
rossf7: Is the `--` here needed?

examples/README.md
```diff
@@ -21,19 +20,21 @@ setup. See [this guide][cert-operator-local-setup] for details. The operator and
...
+helm install -n cert-operator-lab ./cert-operator/examples/cert-operator-lab-chart/ \
   --set commonDomain=my-common-domain \
   --set imageTag=latest \
   --wait
 
 # here the certificate TPR is created, wait until `kubectl get certificate` returns
 # `No resources found.` before running the next command
 
-helm install -n cert-resource-lab ./cert-operator/examples/cert-resource-lab-chart/ \
+helm registry install quay.io/giantswarm/cert-resource-lab-chart:stable -- -n cert-resource-lab \
```
rossf7: Also here

examples/README.md
fgimenez: In this case it's not strictly required but as soon as we add an option to registry it would be. It's a way to keep separate the options related to the registry plugin from those related to the helm command itself. From `helm registry install --help`:
```
Fetch a Chart from the app-registry and execute `helm install`.
Helm's options can be passed on the command:
$ appr helm install [APPR_OPTS] -- [HELM_OPTS]
$ appr helm install quay.io/ant31/cookieapp -- --set imageTag=v0.4.5 --namespace demo
```
Although we could remove them in this case I'd rather keep it, WDYT?

fgimenez:
> Could you add the helm del commands?

Sure, done

examples/README.md
rossf7: Got it thanks. Yes I'm fine with keeping them.

PR #534, opened 2017-11-30
fix cloudformation template paths

fgimenez:
These changes fix the template location for both unit testing and actual operator execution. I've put the logic that calculates the root directory from the value returned by `os.Getwd` in the `key` package, pls let me know if we should move it elsewhere.


service/resource/cloudformation/adapter/small_cloudconfig.go
```diff
@@ -4,18 +4,23 @@ import (
...
 )
 
 func SmallCloudconfig(config SmallCloudconfigConfig) (string, error) {
-	_, filename, _, _ := runtime.Caller(1)
-	templateFile, err := filepath.Abs(filepath.Join(filepath.Dir(filename), "../../../../", smallCloudConfigTemplate))
+	baseDir, err := os.Getwd()
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
+	rootDir, err := key.RootDir(baseDir, "aws-operator")
```
rossf7: I think we should use a constant here.

service/resource/cloudformation/main_stack.go
```diff
@@ -43,18 +43,24 @@ func (r *Resource) getMainTemplateBody(customObject awstpr.CustomObject) (string
...
 	// parse templates
-	_, filename, _, _ := runtime.Caller(1)
-	baseDir, err := filepath.Abs(filepath.Join(filepath.Dir(filename), "../../../", cloudFormationTemplatesDirectory))
+	baseDir, err := os.Getwd()
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
-	files, err := ioutil.ReadDir(baseDir)
+
+	rootDir, err := key.RootDir(baseDir, "aws-operator")
```
rossf7: Same here.

fgimenez:
Done, thanks!

PR #538, opened 2017-11-30
Versioned resources

rossf7:
Adds v1 suffix to resource names as part of the CRD migration. Also some boyscouting to simplify the test numbering.


service/resource_router_test.go
```diff
@@ -124,7 +124,7 @@ func Test_Service_NewResourceRouter(t *testing.T) {
 		errorMatcher      func(err error) bool
 		resourceRouter    func(ctx context.Context, obj interface{}) ([]framework.Resource, error)
 	}{
-		// Case 1. No version in version bundle so return legacy resources.
+		// Case 0. No version in version bundle so return legacy resources.
```
rossf7: Here is the boyscouting. Everything else is renaming.

PR #539, opened 2017-12-01
add template outputs

fgimenez:

This PR adds outputs to the stack, those will be used in the reconciliation loop to determine if the stack needs to be changed.

Instead of getting values from the created resources we interpolate the values directly, this is due to limitations on the kind of values that we can get from those resources http://docs.aws.amazon.com/en_en/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html#intrinsic-function-reference-getatt-attrib for instance we can't get `ImageId` from a `LaunchConfiguration` resource.

### DIFF
--- a/service/resource/cloudformationv1/adapter/outputs.go
+++ b/service/resource/cloudformationv1/adapter/outputs.go
@@ -1,0 +1,18 @@
+package adapter
+
+import (
+	"github.com/giantswarm/aws-operator/service/key"
+	"github.com/giantswarm/awstpr"
+)
+
+// template related to this adapter: service/templates/cloudformation/outputs.yaml
+
+type outputsAdapter struct {
+	ClusterVersion string
+}
+
+func (o *outputsAdapter) getOutputs(customObject awstpr.CustomObject, clients Clients) error {
+	o.ClusterVersion = key.VersionBundleVersion(customObject)
+
+	return nil
+}
--- a/service/resource/cloudformationv1/adapter/outputs.go
+++ b/service/resource/cloudformationv1/adapter/outputs.go
@@ -12,7 +12,7 @@ type outputsAdapter struct {
 }
 
 func (o *outputsAdapter) getOutputs(customObject awstpr.CustomObject, clients Clients) error {
-	o.ClusterVersion = key.VersionBundleVersion(customObject)
+	o.ClusterVersion = key.ClusterVersion(customObject)
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


service/resource/cloudformationv1/adapter/outputs.go
```diff
@@ -0,0 +1,18 @@
+package adapter
+
+import (
+	"github.com/giantswarm/aws-operator/service/key"
+	"github.com/giantswarm/awstpr"
+)
+
+// template related to this adapter: service/templates/cloudformation/outputs.yaml
```
rossf7: <3

service/resource/cloudformationv1/adapter/outputs.go
```diff
@@ -0,0 +1,18 @@
...
+
+type outputsAdapter struct {
+	ClusterVersion string
+}
+
+func (o *outputsAdapter) getOutputs(customObject awstpr.CustomObject, clients Clients) error {
+	o.ClusterVersion = key.VersionBundleVersion(customObject)
```
rossf7: I think this should be key.ClusterVersion. That gets the k8scloudconfig version from the custom object. If that changes we need to trigger an update to replace the nodes.

The version bundle version is used by the router to decide which version of the resources to execute. I don't think we need an output for that. As its part of the operator and not the cloud formation resource.

WDYT?

service/resource/cloudformationv1/main_stack_test.go
```diff
@@ -97,4 +97,22 @@ func TestMainTemplateExistingFields(t *testing.T) {
 		fmt.Println(body)
 		t.Error("asg element not found")
 	}
+
+	if !strings.Contains(body, "Outputs:") {
+		fmt.Println(body)
+		t.Error("outputs header not found")
+	}
+
+	if !strings.Contains(body, "  WorkersOutput:") {
```
rossf7: The whitespace here worries me a bit. But I guess the indentation is part of the cloud formation yaml?

service/resource/cloudformationv1/adapter/outputs.go
fgimenez: Totally agree, I missed that, fixed.

service/resource/cloudformationv1/main_stack_test.go
fgimenez: Indeed, I've removed it and added constants for the output identifiers, pls take another look.

PR #540, opened 2017-12-01
Adds sec group rule for kubelet scraping

JosephSalisbury:

The newer guest cluster scraping design uses the kubernetes service discovery to discover nodes, and then connects to them via their private IPs. So, we need 10250 (the kubelet port) open between the host cluster and the guest clusters, so prometheus can make the connection.


JosephSalisbury:
@rossf7 @fgimenez i couldn't find the security group in the cloudformation templates - is that not in yet?

PR #542, opened 2017-12-01
enable worker's ASG from template

fgimenez:


service/templates/cloudformation/auto_scaling_group.yaml
```diff
@@ -12,6 +12,10 @@
       HealthCheckGracePeriod: {{ .HealthCheckGracePeriod }}
       MetricsCollection:
         - Granularity: "1Minute"
+      Tags:
+        - Key: Name
+          Value: {{ .ClusterID }}-{{ .ASGType }}
+          PropagateAtLaunch: true
```
rossf7: For the tags we need to have KubernetesCluster. This is used by the cloud provider for persistent volumes and load balancers. We should use the same tags as the current ASG.

https://github.com/giantswarm/aws-operator/blob/afdb61b9c5f3ccbbab67423c2ede52a1b17b814a/resources/aws/auto_scaling_group.go#L83

integration/operator_test.go
```diff
@@ -135,7 +136,12 @@ func TestGuestClusterIsCreated(t *testing.T) {
 		t.Errorf("unexpected error getting operator pod name: %v", err)
 	}
 
-	if err := waitForPodLog(cs, "giantswarm", "cluster '${CLUSTER_NAME}' processed", operatorPodName); err != nil {
+	logEntry := "cluster '${CLUSTER_NAME}' processed"
+	if os.Getenv("VERSION_BUNDLE_VERSION") == "0.2.0" {
+		logEntry = "creating AWS cloudformation stack: created"
+	}
```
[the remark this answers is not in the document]
fgimenez: Yep, i'm using this for local runs, do you think we should set the env var in CI for running on new clusters?

service/resource/cloudformationv1/update.go
```diff
@@ -82,20 +81,6 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
-			},
-			{
-				ParameterKey:   aws.String(imageIDParameterKey),
-				ParameterValue: aws.String(desiredStackState.ImageID),
-			},
-			{
-				ParameterKey:   aws.String(clusterVersionParameterKey),
-				ParameterValue: aws.String(desiredStackState.ClusterVersion),
-			},
-		}
```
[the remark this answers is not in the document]
fgimenez: Yes, the input values are interpolated in the templates, we use outputs for checking/managing the stack state.

service/templates/cloudformation/auto_scaling_group.yaml
fgimenez: Thanks! Sone, in a local test the stack was created, the api was responding but no nodes showed up, could it be related to this?

integration/operator_test.go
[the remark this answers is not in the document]
rossf7: Yes I think that's a good idea.

Even once we're migrated to CF we'll need to create new versions of the resources for breaking changes. Having it as an env var lets us test for that.

service/templates/cloudformation/auto_scaling_group.yaml
rossf7: It could be but it might also be a problem registering with the Ingress ELB. Is it set correctly in the ASG?

rossf7: Sorry it's not that I can see the instances are registered OK. Plus that would just affect ingress and not them appearing as nodes.

I don't think it's the tag because the cloud provider is optional. Can you SSH to the workers using the VPC peering?

fgimenez: Nope, from master it asks for password

rossf7: Sounds like a problem with the cloud config and there are no SSH users.

The only way around that I've found is re-enabling the EC2 keypair in the legacy resource and providing it to the ASG.

PR #543, opened 2017-12-04
Add s3bucket resource and implement get current and desired states

rossf7:

Adds the s3bucket resource and implements getting the current and desired states. I'll implement create and delete as later PRs.

Also extracts getting the IAM user to an `aws` service for AWS logic that needs to be shared between resources. If this looks OK then I'll update the `cloudformation` resource to use it.


service/resource/s3bucketv1/mocks.go
```diff
@@ -0,0 +1,35 @@
...
+	// (splitting by colon)
+	accountIDIndex  = 4
+	accountIDLength = 12
+)
+
+type IAMClientMock struct {
+	accountID string
+}
+
+func (i *IAMClientMock) GetUser(input *iam.GetUserInput) (*iam.GetUserOutput, error) {
```
rossf7: Currently GetUser is duplicated. I think we could move this to an `aws` service. WDYT?

service/resource/s3bucketv1/resource.go
```diff
@@ -0,0 +1,100 @@
...
+		return "", microerror.Mask(err)
+	}
+	userArn := *resp.User.Arn
+	accountID := strings.Split(userArn, ":")[accountIDIndex]
+	if err := adapter.ValidateAccountID(accountID); err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	return accountID, nil
+}
```
rossf7: Again currently getAccountID is duplicated.

service/aws/mocks.go
```diff
@@ -0,0 +1,35 @@
+package aws
```
rossf7: Here is the mock for the AWS service.

service/aws/service.go
```diff
@@ -0,0 +1,81 @@
+package aws
```
rossf7: The aws service has methods that need to be shared between resources.

service/resource/s3bucketv1/resource.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Config represents the configuration used to create a new s3bucket resource.
+type Config struct {
+	// Dependencies.
+	AwsService *awsservice.Service
+	Clients    Clients
+	Logger     micrologger.Logger
+
+	// Settings.
+	AwsConfig awsutil.Config
```
fgimenez: Do we need the AwsConfig here?

service/resource/s3bucketv1/current.go
```diff
@@ -0,0 +1,45 @@
...
+	bucketName := key.BucketName(customObject, accountID)
+	headInput := &s3.HeadBucketInput{
+		Bucket: aws.String(bucketName),
+	}
+	_, err = r.clients.S3.HeadBucket(headInput)
+	if IsBucketNotFound(err) {
+		r.logger.LogCtx(ctx, "debug", "did not find the S3 bucket")
+		return BucketState{}, nil
+
+	} else if err != nil {
```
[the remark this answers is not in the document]
rossf7: Good point, removed.

service/resource/s3bucketv1/resource.go
rossf7: No I'd missed removing it. Thanks its gone now.

PR #544, opened 2017-12-04
s3 worker cloudconfig resource: create

fgimenez:
`Create` methods and tests, plus basic resource's structure.


service/resource/s3workercloudconfigv1/create.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	s3PutInput, err := toPutObjectInput(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	_, err = r.awsClients.S3.PutObject(&s3PutInput)
```
rossf7: I think we should do the standard logging with creating the resource and whether it was created or has already been created.

https://github.com/giantswarm/aws-operator/blob/87c7d592b80c256dafb1ddc204c472989d2dcce3/service/resource/namespacev1/create.go#L18

service/resource/s3workercloudconfigv1/delete.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/framework"
+)
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
+	return nil
+}
+
+func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*framework.Patch, error) {
```
rossf7: See my other comment. I don't think we should implement deleting of objects.

service/resource/s3workercloudconfigv1/create.go
fgimenez: Good point, done

service/resource/s3workercloudconfigv1/delete.go
fgimenez: Indeed, They were there just for implementing the interface, but your suggestion makes total sense, I've moved the stub methods to `resource.go`.

fgimenez:
> IMO the resource name should be more generic. Something like s3object as we'll also need to generate the master CC. I think the desired state can implement generating both cloud configs.

Done, I started only with that generic name but later changed to the specific one, in order to have concerns well delimited, but yes, makes more sense to have a single component to deal with all the s3 objects.
 
> Also we're going to need to version the cloudconfigs in the S3 bucket so that we can do rolling updates of the workers.

Ok, I'll implement this in the next PR with current and desired.

PR #546, opened 2017-12-04
implement s3objectv1's GetCurrentState method

fgimenez:

Highlights from my side in this PR:
* I've added a new worker's member to the `BucketObjectState` struct so that we can use the same resource for other S3 objects.
* The versioning is implemented at the object key level (keys look like `cloudconfig/v_1_0_0/worker`), we could also use s3 versioning, let me know WDYT.
* Used AwsService (thanks for creating this! :)) and added mocking, maybe we could move the mock to the service for reusing.

### DIFF
--- a/service/key/key.go
+++ b/service/key/key.go
@@ -35,6 +35,10 @@ func BucketName(customObject awstpr.CustomObject, accountID string) string {
 	return fmt.Sprintf("%s-g8s-%s", ClusterID(customObject), accountID)
 }
 
+func BucketObjectName(customObject awstpr.CustomObject, prefix string) string {
+	return fmt.Sprintf("cloudconfig/%s/%s", ClusterVersion(customObject), prefix)
+}
+
 func ClusterCustomer(customObject awstpr.CustomObject) string {
 	return customObject.Spec.Cluster.Customer.ID
 }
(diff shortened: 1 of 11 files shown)


service/key/key.go
```diff
@@ -35,6 +35,10 @@ func BucketName(customObject awstpr.CustomObject, accountID string) string {
 	return fmt.Sprintf("%s-g8s-%s", ClusterID(customObject), accountID)
 }
 
+func BucketObjectName(customObject awstpr.CustomObject, prefix string) string {
+	return fmt.Sprintf("cloudconfig/%s/%s", ClusterVersion(customObject), prefix)
+}
```
rossf7: Nice, I think this is the correct key structure. We can look at S3 versioning later if necessary.

service/resource/s3objectv1/spec.go
```diff
@@ -2,7 +2,15 @@ package s3objectv1
...
 import "github.com/aws/aws-sdk-go/service/s3"
 
-type BucketObject struct {
+const (
+	prefixWorker = "worker"
+)
+
+type BucketObjectState struct {
+	WorkerCloudConfig BucketObjectInstance
+}
```
rossf7: We could also have an array of bucket objects? Fine with this for now and we can revisit when we add the master cloud config.

fgimenez: Maybe a `map[string]BucketObjectInstance`? IMO with an array we would depend on order, but introducing a map would lead us to implement checks for existent keys, WDYT? We can discuss later in any case.

rossf7: Yes, a map is better and prevents ordering problems.

PR #547, opened 2017-12-04
Implement deleting the S3 bucket

rossf7:

Final PR that implements deleting the S3 bucket.


rossf7:
Going with this for now. Good point about extending the mock to increase the test coverage. I'll try and tackle that as a follow up change.

PR #550, opened 2017-12-05
implement s3objectv1's GetDesiredState

fgimenez:

Implementation of the `GetDesiredState` method, it duplicates part of the TLS assets handling from the legacy resource. The resource needs more dependencies now, CertWatcher and CloudConfig, and we need also an additional KMS client for encryption.

The KMS creation is still done in legacy, here we just query the key, this is done in `service/aws`. Also `AwsServiceMock` is moved to that package so that we can reuse.

### DIFF
--- a/service/aws/mocks.go
+++ b/service/aws/mocks.go
@@ -1,6 +1,8 @@
 package aws
 
 import (
+	"fmt"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/iam"
 )
@@ -32,4 +34,26 @@ func (i *IAMClientMock) GetUser(input *iam.GetUserInput) (*iam.GetUserOutput, er
 	}
 
 	return output, nil
+}
+
+type AwsServiceMock struct {
+	AccountID string
+	KeyArn    string
+	IsError   bool
+}
+
+func (a AwsServiceMock) GetAccountID() (string, error) {
+	if a.IsError {
+		return "", fmt.Errorf("error!!")
+	}
+
+	return a.AccountID, nil
+}
+
+func (a AwsServiceMock) GetKeyArn(cllustterID string) (string, error) {
+	if a.IsError {
+		return "", fmt.Errorf("error!!")
+	}
+
+	return a.KeyArn, nil
 }
--- a/service/aws/mocks.go
+++ b/service/aws/mocks.go
@@ -50,7 +50,7 @@ func (a AwsServiceMock) GetAccountID() (string, error) {
 	return a.AccountID, nil
 }
 
-func (a AwsServiceMock) GetKeyArn(cllustterID string) (string, error) {
+func (a AwsServiceMock) GetKeyArn(clusterID string) (string, error) {
 	if a.IsError {
 		return "", fmt.Errorf("error!!")
 	}
--- a/service/aws/service.go
+++ b/service/aws/service.go
@@ -1,10 +1,13 @@
 package aws
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/giantswarm/microerror"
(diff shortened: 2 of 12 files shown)


service/aws/mocks.go
```diff
@@ -33,3 +35,25 @@ func (i *IAMClientMock) GetUser(input *iam.GetUserInput) (*iam.GetUserOutput, er
...
+
+func (a AwsServiceMock) GetAccountID() (string, error) {
+	if a.IsError {
+		return "", fmt.Errorf("error!!")
+	}
+
+	return a.AccountID, nil
+}
+
+func (a AwsServiceMock) GetKeyArn(cllustterID string) (string, error) {
```
rossf7: Typo on clusterID

service/aws/service.go
```diff
@@ -79,3 +82,20 @@ func ValidateAccountID(accountID string) error {
 
 	return nil
 }
+
+// GetKeyArn returns the key ARN associated with the given cluster ID.
```
rossf7: Good that this is in the AWS service. I think we'll need another resource for the KMS key since it needs to be created before cloud formation.

service/resource/s3objectv1/create_test.go
```diff
@@ -87,11 +89,13 @@ func Test_Resource_S3Object_newCreate(t *testing.T) {
 	resourceConfig.Clients = Clients{
 		S3: &S3ClientMock{},
 	}
-	resourceConfig.AwsService = AwsServiceMock{}
+	resourceConfig.AwsService = awsservice.AwsServiceMock{}
+	resourceConfig.CertWatcher = &CertWatcherMock{}
```
rossf7: Nice that this is mocked.

It has a different purpose but there is also the certificatetprtest package which is useful for meeting the dependencies in tests. Mentioning it in case you haven't seen it.

https://github.com/giantswarm/certificatetpr/blob/master/certificatetprtest/service.go

service/resource/s3objectv1/desired.go
```diff
@@ -1,7 +1,51 @@
...
+	body, err := r.cloudConfig.NewWorkerTemplate(customObject, *tlsAssets)
+	if err != nil {
+		return output, microerror.Mask(err)
+	}
+
+	output.WorkerCloudConfig = BucketObjectInstance{
+		Bucket: key.BucketName(customObject, accountID),
+		Body:   body,
+		Key:    key.BucketObjectName(customObject, prefixWorker),
+	}
```
rossf7: I like that the desired state is pretty short given the amount of steps that need to be taken.

service/resource/s3objectv1/resource.go
```diff
@@ -17,28 +17,34 @@ const (
...
 type Config struct {
 	// Dependencies.
-	AwsService AwsService
-	Clients    Clients
-	Logger     micrologger.Logger
+	AwsService  AwsService
+	CertWatcher CertWatcher
+	Clients     Clients
+	CloudConfig CloudConfigService
+	Logger      micrologger.Logger
```
rossf7: Doesn't this also need the KeyWatcher from `randomkeytpr`?

service/aws/mocks.go
fgimenez: Sure, fixed

service/aws/service.go
fgimenez: Indeed, as soon as we need to create it we'll need a separate resource.

service/resource/s3objectv1/create_test.go
fgimenez: Nice! Didn't know, i've removed the local mock, thx!

service/resource/s3objectv1/resource.go
fgimenez: Not yet, the encoded `clusterKeys` generated by using the keys obtained by `KeyWatcher` in legacy are used to generate the master's cloudconfig https://github.com/giantswarm/aws-operator/blob/master/service/resource/legacyv1/launch_configuration.go#L45 but not for worker's https://github.com/giantswarm/aws-operator/blob/master/service/resource/legacyv1/launch_configuration.go#L50, once this is ported we'll need a `KeyWatcher` dependency too.

PR #551, opened 2017-12-05
Extract creating operator framework to framework.go

rossf7:
This is preparation for the CRD migration.

The creation of the operator framework is extracted to framework.go. A second framework that watches the CROs will be added as well as v2 packages that use the CRD. 

Also added some tests for service creation.

### DIFF
--- a/service/service_test.go
+++ b/service/service_test.go
@@ -1,0 +1,63 @@
+package service
+
+import (
+	"testing"
+
+	"github.com/giantswarm/aws-operator/flag"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/spf13/viper"
+)
+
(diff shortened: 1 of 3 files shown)


service/service_test.go
```diff
@@ -0,0 +1,63 @@
...
+func Test_Service_New(t *testing.T) {
+	tests := []struct {
+		config               func() Config
+		description          string
+		expectedErrorHandler func(error) bool
+	}{
+		{
+			description:          "default config is invalid",
+			config:               DefaultConfig,
+			expectedErrorHandler: IsInvalidConfig,
```
fgimenez: Nice!

rossf7:
Yes, 300 lines is still way too long. Refactoring this after we've removed the legacy resource would also be good IMO.

PR #553, opened 2017-12-05
specify region when creating the S3 client

fgimenez:
It will prevent this kind of error at the first stages of the `s3bucketv1` resource handling:
```
2017/12/05 11:39:36 {"caller":"github.com/giantswarm/aws-operator/service/resource/s3bucketv1/current.go:18","debug":"looking for the S3 bucket","function":"GetCurrentState","object":"/apis/cluster.giantswarm.io/v1/namespaces/default/awses/fede-test","resource":"s3bucket","time":"2017-12-05 11:35:06.541"}
2017/12/05 11:39:36 {"caller":"github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/operatorkit/framework/framework.go:305","error":"[{/home/fgimenez/workspace/gocode/src/github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/operatorkit/framework/framework.go:547: } {/home/fgimenez/workspace/gocode/src/github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/operatorkit/framework/resource/metricsresource/resource.go:86: } {/home/fgimenez/workspace/gocode/src/github.com/giantswarm/aws-operator/service/resource/s3bucketv1/current.go:35: } {MissingRegion: could not find region configuration}]","event":"update","object":"/apis/cluster.giantswarm.io/v1/namespaces/default/awses/fede-test","time":"2017-12-05 11:35:06.768"}
```

### DIFF
--- a/client/aws/aws.go
+++ b/client/aws/aws.go
@@ -41,6 +41,7 @@ type Clients struct {
 const (
 	accountIDPosition = 4
 	accountIDLength   = 12
+	defaultRegion     = "eu-central-1"
 )
 
 func NewClients(config Config) Clients {
@@ -57,7 +58,7 @@ func NewClients(config Config) Clients {
 		IAM:            iam.New(s),
 		KMS:            kms.New(s),
 		Route53:        route53.New(s),
-		S3:             s3.New(s),
+		S3:             s3.New(s, aws.NewConfig().WithRegion(defaultRegion)),
 	}
 
 	return clients
--- a/client/aws/aws.go
+++ b/client/aws/aws.go
@@ -41,7 +41,6 @@ type Clients struct {
 const (
 	accountIDPosition = 4
 	accountIDLength   = 12
-	defaultRegion     = "eu-central-1"
 )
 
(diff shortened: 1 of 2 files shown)


rossf7:
I'm sorry you hit that bug. I think its due to the restructuring I did to use framework,go as I didn't see it when I was testing the bucket resource.

The region handling is a bit complex. In the legacy resource we get the region from the custom object.

https://github.com/giantswarm/aws-operator/blob/b4bfc7c9174e755ca2fa905a100968ba0d785777/service/resource/legacyv1/resource.go#L263

For the alerter service I added a flag for region which comes from the installations repo. 

https://github.com/giantswarm/aws-operator/blob/b4bfc7c9174e755ca2fa905a100968ba0d785777/service/service.go#L124

I think we should use the flag for this. As at least the region will match the installation. In the resources we should either make the region configurable or remove it from the custom object.

fgimenez:
Thanks, and yes, makes sense to use the flag, already added and tested pls take another look.

PR #555, opened 2017-12-05
Delete the main stack in the legacy resource

rossf7:
When using the Cloud Formation transition version we need to delete the main stack before deleting other resources.

This is so the worker nodes are terminated. Otherwise the internet gateway and subnets can't be deleted as well as the VPC itself.


rossf7:
Good point, I've removed the wait for the stack deletion. Tested locally and the VPC and all other resources were deleted.

PR #556, opened 2017-12-05
Add a new release for 1.8.4, deprecate previous releases

oponder:
I think this does what I'd want. It makes it clear that it is impossible to select 1.8.1 clusters now (by making them deprecated)
And makes a 1.0.0 release so that we communicate confidence.

This makes me think that we should have validation somewhere that stops people from selecting invalid or deprecated versions. Do we already have that @xh3b4sd?

### DIFF
--- a/service/framework.go
+++ b/service/framework.go
@@ -268,9 +268,9 @@ func newCustomObjectFramework(config Config) (*framework.Framework, error) {
 	// We provide a map of resource lists keyed by the version bundle version
 	// to the resource router.
 	versionedResources := map[string][]framework.Resource{
-		key.LegacyVersion:         legacyResources,
-		key.CloudFormationVersion: resources,
-		"1.0.0":                   legacyResources,
+		keyv1.LegacyVersion:         legacyResources,
+		keyv1.CloudFormationVersion: resources,
+		"1.0.0":                     legacyResources,
 	}
 
 	initCtxFunc := func(ctx context.Context, obj interface{}) (context.Context, error) {
--- a/service/framework.go
+++ b/service/framework.go
@@ -270,7 +270,7 @@ func newCustomObjectFramework(config Config) (*framework.Framework, error) {
 	versionedResources := map[string][]framework.Resource{
 		key.LegacyVersion:         legacyResources,
 		key.CloudFormationVersion: resources,
-		"1.0.0":                   resources,
+		"1.0.0":                   legacyResources,
 	}
 
 	initCtxFunc := func(ctx context.Context, obj interface{}) (context.Context, error) {
--- a/service/framework.go
+++ b/service/framework.go
@@ -268,8 +268,9 @@ func newCustomObjectFramework(config Config) (*framework.Framework, error) {
 	// We provide a map of resource lists keyed by the version bundle version
 	// to the resource router.
 	versionedResources := map[string][]framework.Resource{
-		keyv1.LegacyVersion:         legacyResources,
-		keyv1.CloudFormationVersion: resources,
+		key.LegacyVersion:         legacyResources,
+		key.CloudFormationVersion: resources,
+		"1.0.0":                   resources,
 	}
 
 	initCtxFunc := func(ctx context.Context, obj interface{}) (context.Context, error) {
--- a/service/resource_router_test.go
+++ b/service/resource_router_test.go
@@ -117,7 +117,7 @@ func Test_Service_NewResourceRouter(t *testing.T) {
 	versionedResources := make(map[string][]framework.Resource)
 	versionedResources["0.1.0"] = legacyResources
 	versionedResources["0.2.0"] = allResources
-	versionedResources["1.0.0"] = allResources
+	versionedResources["1.0.0"] = legacyResources
 
 	testCases := []struct {
 		customObject      awstpr.CustomObject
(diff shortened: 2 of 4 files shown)


oponder:
fyi @pipo02mix

oponder:
screen shot 2017-12-05 at 13 45 34


We see again `Vault version updated` for this new release. So the way the aggregator works is not entirely going to lead to changelog text that makes sense. (Vault comes from cert-operator)

oponder:
btw, It says 7 hours ago because I set it for 1pm today, but I'm of course not in the right time zone.

The exact correct time of the release isn't super important IMHO.

xh3b4sd:
> This makes me think that we should have validation somewhere that stops people from selecting invalid or deprecated versions. Do we already have that @xh3b4sd?

We do not have this but it should be quite easy to add. We can talk about this and the approaches separately. Lets just talk. 

> The exact correct time of the release isn't super important IMHO.

Agreed, as long as the day is the correct one and there is no conflict with version bundles created the same day.

oponder:
> We also need to add them to the resource router, otherwise the operator will not accept custom objects having the older version bundle versions.

I don't know what this means or how to do this yet. If it's easy can you do a commit on this PR?

xh3b4sd:
Done. See https://github.com/giantswarm/aws-operator/pull/556/commits/d563d950dc84e41977752d1e6c6d12f63501bc1a.

xh3b4sd:
I also made the version bundles being deprecated as not being WIP anymore. See https://github.com/giantswarm/aws-operator/pull/556/commits/5d919c6c18d09ad473f08012cf899017299bb67e.

oponder:
What's the idea behind WIP btw. Should those not show up as user selectable?

xh3b4sd:
We can think about what it should mean. My intention back then was to be able to have version bundles in a state before and after development cycles. So `wip` means the version bundle is fresh and under development and `deprecated` means there is no maintenance anymore. So yeah, I guess both should not be selectable by users. We can make these available to us though to be able to test and develop. I think the api components can simply do the filtering based on access rights.

service/framework.go
```diff
@@ -251,6 +251,7 @@ func newCustomObjectFramework(config Config) (*framework.Framework, error) {
 	versionedResources := map[string][]framework.Resource{
 		key.LegacyVersion:         legacyResources,
 		key.CloudFormationVersion: resources,
+		"1.0.0":                   resources,
```
rossf7: This should be `legacyResources`. The cloud formation changes aren't ready to expose to customers yet.

service/version_bundles.go
```diff
@@ -109,11 +109,52 @@ func NewVersionBundles() []versionbundle.Bundle {
 				},
 			},
 			Dependencies: []versionbundle.Dependency{},
-			Deprecated:   false,
+			Deprecated:   true,
 			Name:         "aws-operator",
 			Time:         time.Date(2017, time.November, 29, 16, 16, 0, 0, time.UTC),
 			Version:      "0.2.0",
-			WIP:          true,
+			WIP:          false,
```
rossf7: I think this should be true. The cloud formation changes are still WIP.

xh3b4sd: Ack. When it is WIP it must not be deprecated.

service/framework.go
xh3b4sd: Ack.

xh3b4sd:
@oponder would be nice if you could take over again.

xh3b4sd:
Uncertain, I am not too deep into the project here. @rossf7 may have a better opinion here.

rossf7:
Yes `0.1.0` should be deprecated and not WIP.

We used this so that existing clusters are only processed by the legacy resource. But `1.0.0` makes more sense since its what is running in production

pipo02mix:
Nice work

pipo02mix:
I have updated the PR with @rossf7 suggestions. So, it does mean version 0.2.0 (cloud formation) will be on WIP and cluster service (I guess) should take care to not show it in the releases.

cc/ @marians

service/resource_router_test.go
```diff
@@ -117,6 +117,7 @@ func Test_Service_NewResourceRouter(t *testing.T) {
 	versionedResources := make(map[string][]framework.Resource)
 	versionedResources["0.1.0"] = legacyResources
 	versionedResources["0.2.0"] = allResources
+	versionedResources["1.0.0"] = allResources
```
oponder: is this still right?

oponder: ping @rossf7 .. sorry Im not looking super critically. just seeing `allResources` next to `1.0.0` when we said it should use `legacyResources`. so just double checking

pipo02mix: yeah I guess test should be aligned

pipo02mix:
@xh3b4sd could you review too? as you requested some changes too :)

service/resource_router_test.go
rossf7: Yes test should be aligned. Sorry I missed that and thanks for fixing Fer.

oponder:
related to: https://github.com/giantswarm/giantswarm/issues/2237#issuecomment-349926940

oponder:
Ok, going to rebase and fix conflicts and merge.

thanks for all the help guys

service/resource_router_test.go
xh3b4sd: I noticed this as well. IMO this is a sign for a bad test design. We test what we compute in the test setup. That makes not that much sense. We should rather test what we use in production kind of, which means that the resources we compute should be configured with stubs and tested here. Lets try to improve this in the future.

PR #581, opened 2017-12-13
migrated worker policy

fgimenez:


service/resource/cloudformationv2/adapter/auto_scaling_group.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/giantswarm/aws-operator/service/keyv2"
 )
 
-// template related to this adapter: service/templates/cloudformation/auto_scaling_group.go
+// template related to this adapter: service/templates/cloudformation/auto_scaling_group.yaml
```
rossf7: <3

service/resource/cloudformationv2/adapter/launch_configuration.go
```diff
@@ -81,15 +79,10 @@ func (l *launchConfigAdapter) getLaunchConfiguration(customObject v1alpha1.AWSCo
 	l.SecurityGroupID = *output.SecurityGroups[0].GroupId
 
 	// cloud config
-	resp, err := clients.IAM.GetUser(&iam.GetUserInput{})
+	accountID, err := AccountID(clients)
```
rossf7: Nitpick but the cloud config comment is a bit confusing.

service/resource/cloudformationv2/create.go
```diff
@@ -52,6 +52,10 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
 		createState.StackName = aws.String(desiredStackState.Name)
 		createState.TemplateBody = aws.String(mainTemplate)
 		createState.TimeoutInMinutes = aws.Int64(defaultCreationTimeout)
+		// CAPABILITY_NAMED_IAM is required for creating IAM roles (worker policy)
+		createState.Capabilities = []*string{
+			aws.String("CAPABILITY_NAMED_IAM"),
```
rossf7: A constant for this would be good.

service/templates/cloudformation/main.yaml
```diff
@@ -3,5 +3,6 @@ Description: Main CloudFormation stack.
 Resources:
   {{template "launch_configuration" .}}
   {{template "autoscaling_group" .}}
+  {{template "worker_policy" .}}
```
rossf7: I'm assuming Cloud Formation doesn't care about the ordering here?

But having the worker policy before the launch config makes more sense to me. Since it's needed by the launch config.

service/resource/cloudformationv2/adapter/launch_configuration.go
fgimenez: Good point, I've modified it to make clear that it is related to the logic needed to determine the small cloud config field.

service/resource/cloudformationv2/create.go
fgimenez: Sure, done

service/templates/cloudformation/main.yaml
fgimenez: Makes sense, done

service/resource/legacyv2/resource.go
```diff
@@ -207,7 +213,27 @@ func (s *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desire
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		update, err := s.newUpdateChange(ctx, obj, currentState, desiredState)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		patch := framework.NewPatch()
```
rossf7: Redefining patch here causes problems when launching clusters with cloud formation.  No stack is created because the create and update changes aren't passed to the framework.

I found this when testing my master node changes.

fgimenez: Very good catch! Feel free to push your fix pls :)

rossf7: Fix is pushed.

fgimenez:
Thx Ross! There were also some missing tweaks for the recordsets, already in place and tested.

PR #583, opened 2017-12-14
unified transition resource

fgimenez:

Requires #581 (for nothing special, has been forked from it because it had more components migrated)

These changes are motivated by the issues we are finding with dependent resources in operatorkit https://github.com/giantswarm/operatorkit/issues/139 having a single resource for both the legacy and the new cloudformation code solves the dependency problems, we can decide when to create or delete each of the components.

I've tested locally and works fine for both 0.1.0 and 0.2.0, let me know WDYT. The downside is that the single resource is more complex, I think we can iterate to make things more manageable. The old cloudformationv{1,2} are removed to prevent test failures when the template would evolve. I haven't put in place all the existing unit tests for resource-related methods, if we agree in this approach I'll move them here.


service/resource/legacyv2/resource.go
```diff
@@ -207,12 +215,31 @@ func (s *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desire
 
 	s.logger.Log("info", fmt.Sprintf("cluster '%s' processed", keyv2.ClusterID(cluster)))
 
-	return &framework.Patch{}, nil
+	patch := framework.NewPatch()
+
+	// cloudformation logic
```
rossf7: I think more comments here would be good. To mention that the resources are combined during the transition due to the deletion order problem and that they will be separated later.

service/resource/legacyv2/resource.go
```diff
@@ -180,22 +190,20 @@ type Resource struct {
...
 
-type Event struct {
-	Type   string
-	Object *v1alpha1.AWSConfig
-}
-
 // NewUpdatePatch is called upon observed custom object change. It creates the
 // AWS resources for the cluster.
 func (s *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*framework.Patch, error) {
+	s.logger.Log("info", "in NewUpdatePatch")
```
rossf7: Are these logs for debugging or will they be kept?

service/resource/s3objectv2/desired.go
```diff
@@ -27,6 +27,10 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	}
 
 	kmsArn, err := r.awsService.GetKeyArn(clusterID)
+	if IsKeyNotFound(err) {
+		// we can get here during deletion, if he key is already deleted we can safely exit
```
rossf7: Good this is handled and commented. Just a typo on "the key" and can you make it a sentence.

service/resource/legacyv2/resource.go
fgimenez: Only for making clear the code flow during debug, removed.

service/resource/legacyv2/resource.go
fgimenez: Indeed, I've extended the existing ones, please take a look.

service/resource/s3objectv2/desired.go
fgimenez: Sure, done

PR #584, opened 2017-12-15
Migrate route53  recordsets

fgimenez:

### DIFF
--- a/service/resource/legacyv2/adapter/error.go
+++ b/service/resource/legacyv2/adapter/error.go
@@ -20,7 +20,7 @@ func IsMalformedAmazonAccountID(err error) bool {
 
 var wrongAmazonAccountIDLengthError = microerror.New("wrong amazon account ID length")
 
-// IsWrongAmazonIDLength asserts wrongAmazonAccountIDLengthError.
+// IsWrongAmazonAccountIDLength asserts wrongAmazonAccountIDLengthError.
 func IsWrongAmazonAccountIDLength(err error) bool {
 	return microerror.Cause(err) == wrongAmazonAccountIDLengthError
 }
--- a/service/resource/legacyv2/adapter/recordsets.go
+++ b/service/resource/legacyv2/adapter/recordsets.go
@@ -1,0 +1,77 @@
+package adapter
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elb"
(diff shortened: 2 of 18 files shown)


service/templates/cloudformation/worker_policy.yaml
```diff
@@ -9,12 +10,12 @@
           Principal:
             Service: "ec2.amazonaws.com"
           Action: "sts:AssumeRole"
-  workerRolePolicy:
+  WorkerRolePolicy:
```
rossf7: I like that this makes the resource names consistent and starting with a capital. Should we do the same for the R53 recordsets?

service/resource/legacyv2/adapter/error.go
```diff
@@ -20,7 +20,7 @@ func IsMalformedAmazonAccountID(err error) bool {
 
 var wrongAmazonAccountIDLengthError = microerror.New("wrong amazon account ID length")
 
-// IsWrongAmazonIDLength asserts wrongAmazonAccountIDLengthError.
+// IsWrongAmazonAccountIDLength asserts wrongAmazonAccountIDLengthError.
```
rossf7: <3

service/resource/legacyv2/adapter/recordsets.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (r *recordSetsAdapter) getRecordSets(customObject v1alpha1.AWSConfig, clients Clients) error {
+	r.APIELBHostedZones = customObject.Spec.AWS.API.HostedZones
+	r.APIELBDomain = customObject.Spec.Cluster.Kubernetes.API.Domain
+	r.EtcdELBHostedZones = customObject.Spec.AWS.Etcd.HostedZones
+	r.EtcdELBDomain = customObject.Spec.Cluster.Etcd.Domain
+	r.IngressELBHostedZones = customObject.Spec.AWS.Ingress.HostedZones
+	r.IngressELBDomain = customObject.Spec.Cluster.Kubernetes.IngressController.Domain
+	r.IngressWildcardELBDomain = customObject.Spec.Cluster.Kubernetes.IngressController.WildcardDomain
```
rossf7: IMO no point having key helpers for these since they are only needed once.

service/templates/cloudformation/worker_policy.yaml
fgimenez: Makes perfect sense, fixed

PR #585, opened 2017-12-18
restructure introducing framework

fgimenez:
Code reorg before changes for waiting for the guest cluster API up and adding more actual tests. It aligns with upstream framework concept https://github.com/kubernetes/kubernetes/tree/master/test/e2e/framework. Currently belongs to the same `integration` package as the tests (and because of that all the methods are private) but on further iterations it should be probably moved to its specific package.


.circleci/config.yml
```diff
@@ -9,11 +10,6 @@ e2eTest: &e2eTest
 
     - run: ./e2e-harness setup --name=ci-awsop-${CIRCLE_SHA1:0:7}
 
-    - run:
-        name: additional setup required for compiling go on circleCI
-        command: |
-          mkdir -p /home/circleci/.go_workspace/src
-          cp -ar vendor/* /home/circleci/.go_workspace/src/
```
kopiczko: Why `working_directory` is needed and this is removed?

fgimenez: It is simpler to put the working directory inside `GOPATH` than copying the vendor directory there from another location, single directive and no copying involved.

integration/framework.go
```diff
@@ -0,0 +1,301 @@
...
+		pod := pods.Items[0]
+		phase := pod.Status.Phase
+		if phase != v1.PodRunning {
+			return microerror.Maskf(unexpectedStatusPhaseError, "current status: %s", string(phase))
+		}
+		return nil
+	}
+}
+
+func (f *framework) activeNamespaceFunc(name string) func() error {
```
kopiczko: Those names are strange. Even though they are passed as arguments to other functions I think names should reflect what they do, not how they are used.

fgimenez: Makes total sense, thanks for pointing it out, fixed.

PR #587, opened 2017-12-19
Add master node

rossf7:


service/resource/legacyv2/adapter/instance.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/aws-operator/service/keyv2"
+)
+
+// template related to this adapter: service/templates/cloudformation/instance.yaml
+
+type instanceAdapter struct {
+	MasterAZ                     string
```
rossf7: I've prefixed the AZ with Master but its needed in multiple adaptors. Should we have these shared fields in the main Adaptor struct?

service/resource/legacyv2/adapter/instance.go
```diff
@@ -0,0 +1,116 @@
...
+	MasterIAMInstanceProfileName string
```
rossf7: The other fields need the Master prefix to avoid collisions. If this looks OK I'll do a PR to prefix the Worker fields for the launch config & ASG.

service/resource/legacyv2/adapter/instance.go
fgimenez: Yes, I think the shared fields should be promoted to the main Adapter, maybe the Master prefix still makes sense there.

service/resource/legacyv2/adapter/instance.go
fgimenez: Totally, better than qualifying the fields IMO.

service/resource/legacyv2/adapter/instance.go
```diff
@@ -0,0 +1,116 @@
...
+		return microerror.Mask(invalidConfigError)
+	}
+
+	i.MasterAZ = keyv2.AvailabilityZone(customObject)
+	i.MasterImageID = keyv2.MasterImageID(customObject)
+	i.MasterInstanceType = keyv2.MasterInstanceType(customObject)
+	i.MasterIAMInstanceProfileName = keyv2.InstanceProfileName(customObject, prefixMaster)
+
+	// security group
+	// TODO: remove this code once the security group is created by cloudformation
```
fgimenez: Nice! I need to update other adapters to make this clear.

service/resource/legacyv2/adapter/instance.go
```diff
@@ -0,0 +1,116 @@
...
+	subnetOutput, err := clients.EC2.DescribeSubnets(describeSubnetInput)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+	if len(subnetOutput.Subnets) > 1 {
+		return microerror.Mask(tooManyResultsError)
+	}
+	i.MasterSubnetID = *subnetOutput.Subnets[0].SubnetId
+
+	resp, err := clients.IAM.GetUser(&iam.GetUserInput{})
```
fgimenez: We have `AccountID` for this now https://github.com/giantswarm/aws-operator/blob/master/service/resource/legacyv2/adapter/iam.go#L11

rossf7: Done, thanks that's much nicer.

service/resource/legacyv2/adapter/instance.go
rossf7: Good point I'm going to keep the Master prefix for the AZ. I'll do a PR to prefix the worker fields.

PR #588, opened 2017-12-19
Add Worker prefix to fields in launch config and ASG adapters

rossf7:

Adds prefix to fields that are also needed by the master node. This avoids the need to fully qualify the adapter fields. Also minor change to rename recordsets.yml to .yaml.


service/resource/legacyv2/adapter/auto_scaling_group.go
```diff
@@ -17,18 +17,18 @@ import (
 type autoScalingGroupAdapter struct {
 	ASGMaxSize             int
 	ASGMinSize             int
-	AZ                     string
+	ClusterID              string
```
rossf7: I just made this alphabetical.

service/resource/legacyv2/adapter/auto_scaling_group.go
```diff
@@ -17,18 +17,18 @@ import (
...
 	HealthCheckGracePeriod int
 	LoadBalancerName       string
 	MaxBatchSize           string
 	MinInstancesInService  string
 	RollingUpdatePauseTime string
-	SubnetID               string
-	ClusterID              string
+	WorkerAZ               string
+	WorkerSubnetID         string
```
rossf7: I only qualified the AZ and Subnet ID as the rest are worker specific.

fgimenez: :+1: makes sense

PR #589, opened 2017-12-19
e2e tests: wait for guest cluster API up

fgimenez:
In the e2e tests, after the operator reports the guest cluster as created, initialize a guest cluster k8s client and wait until the API is accessible. There's a stub for an actual test (DNS resolution from a pod, as part of the general guest cluster validation) still not implemented, will come next-


integration/framework.go
```diff
@@ -293,9 +295,60 @@ func (f *framework) deleteGuestCluster() error {
...
+	// when the migration is done we will need to check here the cloudformation stack deletion
+	// message
 	logEntry := "cluster '${CLUSTER_NAME}' deleted"
-	if os.Getenv("VERSION_BUNDLE_VERSION") == "0.2.0" {
-		logEntry = "deleting AWS cloudformation stack: deleted"
-	}
 	return f.waitForPodLog("giantswarm", logEntry, operatorPodName)
 }
+
+func (f *framework) initGuestClientset() error {
```
kopiczko: I'd wrap this in https://golang.org/pkg/sync/#Once and call every time in `apiUp`. Otherwise it requires to much knowledge outside the objects how to use it.

I also tend to export the methods which are used outside the object even if it's private. I.e. always think about the object as it could be exported one day by just changing the first letter. I think it makes it easier to maintain that way.

fgimenez: > I'd wrap this in https://golang.org/pkg/sync/#Once and call every time in apiUp. Otherwise it requires to much knowledge outside the objects how to use it.

I agree with the encapsulation of inner workings, but we don't need to use `sync.Once` here, just checking for the zero value of the clientset is enough IMO, let me know WDYT.

> I also tend to export the methods which are used outside the object even if it's private. I.e. always think about the object as it could be exported one day by just changing the first letter. I think it makes it easier to maintain that way.

Sure, maybe it's still too early but fine with me.

kopiczko: > but we don't need to use sync.Once here, just checking for the zero value of the clientset is enough IMO

100% I missed that detail.

integration/framework.go
```diff
@@ -332,7 +332,12 @@ func (f *framework) initGuestClientset() error {
 	return nil
 }
 
-func (f *framework) waitForAPIUp() error {
+func (f *framework) WaitForAPIUp() error {
+	if f.guestCS == nil {
+		if err := f.initGuestClientset(); err != nil {
+			return microerror.Maskf(err, "unexpected error initializing guest clientset")
+		}
+	}
```
kopiczko: Shouldn't that `if` be inside `initGuestClientset`? That way you'll avoid repetition if you need to call `initGuestClientset` in another method.

fgimenez: Indeed, makes sense.

PR #592, opened 2017-12-20
migrate NAT gateway to cloudformation

fgimenez:

### DIFF
--- a/service/resource/legacyv2/adapter/adapter.go
+++ b/service/resource/legacyv2/adapter/adapter.go
@@ -27,13 +27,15 @@ package adapter
 
 import (
 	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/aws-operator/service/keyv2"
 	"github.com/giantswarm/microerror"
 )
 
 type hydrater func(v1alpha1.AWSConfig, Clients) error
 
 type Adapter struct {
-	ASGType string
+	ASGType   string
+	ClusterID string
 
 	launchConfigAdapter
 	autoScalingGroupAdapter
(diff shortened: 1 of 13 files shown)


service/resource/legacyv2/adapter/adapter.go
```diff
@@ -37,9 +37,10 @@ type Adapter struct {
 
 	launchConfigAdapter
 	autoScalingGroupAdapter
-	outputsAdapter
+	natGatewayAdapter
 	workerPolicyAdapter
 	recordSetsAdapter
+	outputsAdapter
```
rossf7: This ordering with outputs last is much nicer. <3

service/templates/cloudformation/nat_gateway.yaml
```diff
@@ -0,0 +1,21 @@
+{{define "nat_gateway"}}
+  NATGateway:
+    Type: AWS::EC2::NatGateway
+    Properties:
+      AllocationId:
+        Fn::GetAtt:
+        - NATEIP
+        - AllocationId
+      SubnetId: {{.PublicSubnetID}}
```
rossf7: Can we add a Name tag here with the cluster ID? 

We do that in the old AWS resources. We get Cloud Formation tags added automatically with the stack name that has the cluster ID. But the name tag is useful in the AWS console.

fgimenez: Sure, done. I've taken the opportunity to promote `ClusterID` to the top level adapter (it was used in ASG too). Also now we don't try to delete NAT gw if on CloudFormation.

rossf7: Thanks, and yes I like that shared fields are moved to the top level adapter.

PR #599, opened 2017-12-27
Add ClusterID tag key/value for all resources

pipo02mix:
Towards [#261](https://github.com/giantswarm/vodafone/issues/261)

Add standard tags to all resources in order to help our customer to manage them

### DIFF
--- a/service/resource/legacyv2/resource.go
+++ b/service/resource/legacyv2/resource.go
@@ -461,8 +461,9 @@ func (s *Resource) processCluster(cluster v1alpha1.AWSConfig) error {
 	{
 		var err error
 		bucket = &awsresources.Bucket{
-			Name:      bucketName,
-			AWSEntity: awsresources.AWSEntity{Clients: clients},
+			AWSEntity:  awsresources.AWSEntity{Clients: clients},
+			Name:       bucketName,
+			ClusterID:  keyv2.ClusterID(cluster),
 		}
 		bucketCreated, err = bucket.CreateIfNotExists()
 		if err != nil {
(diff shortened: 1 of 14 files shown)


service/resource/legacyv2/resource.go
```diff
@@ -461,8 +461,9 @@ func (s *Resource) processCluster(cluster v1alpha1.AWSConfig) error {
 	{
 		var err error
 		bucket = &awsresources.Bucket{
-			Name:      bucketName,
-			AWSEntity: awsresources.AWSEntity{Clients: clients},
+			AWSEntity:  awsresources.AWSEntity{Clients: clients},
+			Name:       bucketName,
+			ClusterID:  keyv2.ClusterID(cluster),
```
rossf7: I'd like us to have a function in the `keyv2` package that takes in a custom object and returns a map of keys and values. That will give us one place where we can add standard tags to all AWS resources.

pipo02mix:
ok then I dont need to go through all resources because the stack will propagate the tags. Where do you define the tags for the entire stack?

rossf7:
The tags are defined when you create the stack. That is in the `legacy` resource but Fede is about to split it into a separate `cloudformation` resource. It will be better to add the tags after.

For now you could add tags to the `kmskey` and `s3object` resources. As these are not managed with cloud formation.

https://github.com/giantswarm/aws-operator/blob/a9a03cc12dd88772744e0a82c74f57e2e6640946/service/resource/kmskeyv2/create.go#L11
https://github.com/giantswarm/aws-operator/blob/a9a03cc12dd88772744e0a82c74f57e2e6640946/service/resource/s3bucketv2/create.go#L11

pipo02mix:
Closing this since now is not needed. PR for S3 buckets and key KMS on the way

PR #600, opened 2018-01-02
Add Internet Gateway to Cloud Formation

rossf7:


service/resource/legacyv2/adapter/internet_gateway.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/aws-operator/service/keyv2"
+)
+
+// template related to this adapter: service/templates/cloudformation/internet_gateway.yaml
+
+type internetGatewayAdapter struct {
+	VPCID string
+}
```
rossf7: The adapter is used to render data in the template. For the inet gateway we just need the VPC ID.

service/resource/legacyv2/resource.go
```diff
@@ -523,23 +523,25 @@ func (s *Resource) processCluster(cluster v1alpha1.AWSConfig) error {
...
-	}
-	internetGatewayCreated, err := internetGateway.CreateIfNotExists()
-	if err != nil {
-		return microerror.Maskf(executionFailedError, fmt.Sprintf("could not create internet gateway: '%#v'", err))
-	}
-	if internetGatewayCreated {
-		s.logger.Log("info", fmt.Sprintf("created internet gateway for cluster '%s'", keyv2.ClusterID(cluster)))
-	} else {
-		s.logger.Log("info", fmt.Sprintf("internet gateway for cluster '%s' already exists, reusing", keyv2.ClusterID(cluster)))
+	if !keyv2.UseCloudFormation(cluster) {
```
rossf7: Here we don't create the gateway if we're using Cloud Formation. This logic will be removed when we split the cloud formation and legacy resources.

service/templates/cloudformation/internet_gateway.yaml
```diff
@@ -0,0 +1,16 @@
...
+          Value: {{ .ClusterID }}
+
+  VPCGatewayAttachment:
+    Type: AWS::EC2::VPCGatewayAttachment 
+    Properties:
+      InternetGatewayId:
+        Ref: InternetGateway
+      VpcId: {{ .VPCID }}
+
+{{end}}
```
rossf7: Here is the Cloud Formation template.

service/resource/legacyv2/adapter/vpc.go
```diff
@@ -8,6 +8,28 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
+func VPCID(clients Clients, name string) (string, error) {
```
rossf7: Here we get the VPC ID. Once the VPC is created by Cloud Formation we can remove this and use a reference in the template.

PR #601, opened 2018-01-02
Complete KMS Key resource

rossf7:
  
Implements NewUpdatePatch to complete the resource and adds it to the framework. I've commented out for now until we separate the cloud formation resource again.


service/resource/kmskeyv2/error.go
```diff
@@ -15,3 +19,15 @@ var wrongTypeError = microerror.New("wrong type")
 func IsWrongType(err error) bool {
 	return microerror.Cause(err) == wrongTypeError
 }
+
+func IsKeyNotFound(err error) bool {
+	aerr, ok := err.(awserr.Error)
+	if !ok {
```
fgimenez: nitpick, maybe we can remove this check and do later: `if ok && aerr.Code() == kms.ErrCodeNotFoundException` so that we have a single `return true` and a single `return false`?

rossf7: Done, that's much nicer. Thx!

PR #605, opened 2018-01-04
Add route tables and subnets to Cloud Formation

rossf7:

Sorry this PR got a bit big. At first I tried to just migrate the private route table and subnet. But I couldn't get the public route table and subnet to delete reliably in the legacy resource.


service/resource/legacyv2/adapter/adapter.go
```diff
@@ -37,15 +37,16 @@ type Adapter struct {
 	ASGType          string
 	AvailabilityZone string
 	ClusterID        string
+	VPCID            string
```
rossf7: I moved VPCID to the main adapter as its needed by multiple adapters.

service/resource/legacyv2/adapter/internet_gateway.go
```diff
@@ -1,37 +0,0 @@
-package adapter
-
```
rossf7: This adapter no longer has any fields so I removed it.

service/resource/legacyv2/adapter/nat_gateway.go
```diff
@@ -1,39 +0,0 @@
-package adapter
```
rossf7: Same with the NAT gateway adapter.

service/resource/legacyv2/adapter/adapter.go
```diff
@@ -37,15 +37,16 @@ type Adapter struct {
...
 
 	autoScalingGroupAdapter
 	iamPoliciesAdapter
 	instanceAdapter
 	launchConfigAdapter
 	loadBalancersAdapter
-	internetGatewayAdapter
```
fgimenez: Very nice that we no longer need these adapters! :+1:

service/templates/cloudformation/auto_scaling_group.yaml
```diff
@@ -3,7 +3,7 @@
     Type: "AWS::AutoScaling::AutoScalingGroup"
     Properties:
       VPCZoneIdentifier:
-        - {{ .WorkerSubnetID }}
```
fgimenez: I really like these data actions changed to CF refs :heart:

PR #609, opened 2018-01-08
Add Security Groups to Cloud Formation

rossf7:


service/resource/legacyv2/adapter/load_balancers.go
```diff
@@ -17,9 +15,6 @@ const (
 	healthCheckInterval           = 5
 	healthCheckTimeout            = 3
 	healthCheckUnhealthyThreshold = 2
-
-	httpPort  = 80
-	httpsPort = 443
```
rossf7: I moved these constants to spec.go

service/resource/legacyv2/adapter/vpc.go
```diff
@@ -30,46 +30,24 @@ func VPCID(clients Clients, name string) (string, error) {
...
-	if len(output.Subnets) > 1 {
-		return "", microerror.Mask(tooManyResultsError)
-	}
-
-	return *output.Subnets[0].SubnetId, nil
-}
-
-func RouteTableID(clients Clients, name string) (string, error) {
-	input := &ec2.DescribeRouteTablesInput{
+func VpcCIDR(clients Clients, vpcID string) (string, error) {
```
rossf7: This function gets the CIDR of the host cluster for the sec group rules.

service/templates/cloudformation/security_groups.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - 
+        IpProtocol: {{ .Protocol }}
+        FromPort: {{ .Port }}
+        ToPort: {{ .Port }}
+        {{ if .SourceCIDR }}
+        CidrIp: {{ .SourceCIDR }}
+        {{ else }}
+        SourceSecurityGroupName: {{ .SourceSecurityGroupName }}
+        {{ end }}
+      {{ end }}
```
rossf7: The conditional logic here matches what we do now. So if there is a SourceCIDR it takes precedence. But its a bit untidy. I did think of having separate slices for CIDR and non-CIDR based rules. WDYT?

service/templates/cloudformation/security_groups.yaml
```diff
@@ -0,0 +1,49 @@
...
+      GroupDescription:  {{ .IngressSecurityGroupName }}
+      VpcId: {{ .VPCID }}
+      SecurityGroupIngress:
+      {{ range .IngressSecurityGroupRules }}
+      - 
+        IpProtocol: {{ .Protocol }}
+        FromPort: {{ .Port }}
+        ToPort: {{ .Port }}
+        CidrIp: {{ .SourceCIDR }}
+      {{ end }}
```
rossf7: Ingress is simpler as all rules are CIDR based.

service/templates/cloudformation/security_groups.yaml
fgimenez: IMO it's better this way, it's not too much logic and having separate slices would make things less clear both in the template and the adapter.

service/templates/cloudformation/security_groups.yaml
```diff
@@ -0,0 +1,49 @@
+{{define "security_groups" }}
+  MasterSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription:  {{ .MasterSecurityGroupName }}
+      VpcId: {{ .VPCID }}
+      Tags:
+        - Key: Name
+          Value:  {{ .MasterSecurityGroupName }}
```
rossf7: Forgot to mention that instead of setting the security group name I'm adding a Name tag. As there are restrictions on updates if you set the name.

> If you don't specify a GroupName, AWS CloudFormation generates a unique physical ID and uses that ID for the group name. For more information, see Name Type.


> If you specify a name, you cannot perform updates that require replacement of this resource. You can perform updates that require no or some interruption. If you must replace the resource, specify a new name.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html

service/resource/legacyv2/adapter/vpc.go
fgimenez: I think `clients` here are currently configured to work with the guest cluster, see how they are injected from `awsClients` https://github.com/giantswarm/aws-operator/blob/master/service/framework.go#L229, how `awsClients` is defined using `awsConfig` https://github.com/giantswarm/aws-operator/blob/master/service/framework.go#L173 and where `awsConfig` comes from https://github.com/giantswarm/aws-operator/blob/master/service/framework.go#L146. 

IMO we could add a `HostClients` member to the resource (only an EC2 client would do), WDYT? This will be needed for porting the VPC and VPCPeering resources too.

service/keyv2/key_test.go
```diff
@@ -186,6 +186,44 @@ func Test_HasClusterVersion(t *testing.T) {
...
 }
 
+func Test_IngressControllerInsecurePort(t *testing.T) {
+	expectedPort := 30010
+	customObject := v1alpha1.AWSConfig{
+		Spec: v1alpha1.AWSConfigSpec{
+			Cluster: v1alpha1.Cluster{
+				Kubernetes: v1alpha1.ClusterKubernetes{
+					IngressController: v1alpha1.ClusterKubernetesIngressController{
+						InsecurePort: 30010,
```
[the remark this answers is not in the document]
rossf7: Good catch, fixed.

service/keyv2/key_test.go
```diff
@@ -186,6 +186,44 @@ func Test_HasClusterVersion(t *testing.T) {
...
+}
+
+func Test_IngressControllerSecurePort(t *testing.T) {
+	expectedPort := 30011
+	customObject := v1alpha1.AWSConfig{
+		Spec: v1alpha1.AWSConfigSpec{
+			Cluster: v1alpha1.Cluster{
+				Kubernetes: v1alpha1.ClusterKubernetes{
+					IngressController: v1alpha1.ClusterKubernetesIngressController{
+						SecurePort: 30011,
```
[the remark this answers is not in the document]
rossf7: Also done.

service/templates/cloudformation/security_groups.yaml
rossf7: Thanks, that makes sense and I'll use the same approach for the master rules.

service/resource/legacyv2/adapter/vpc.go
rossf7: Good catch on this. Yes I think adding a HostClients member is good especially as we only need the EC2 client. I'll implement that. Thx!

rossf7:
@fgimenez For the security groups I'm doing some final testing but otherwise I think its ready. Can you take another look?

service/resource/legacyv2/adapter/security_groups.go
```diff
@@ -0,0 +1,144 @@
+package adapter
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/aws-operator/service/keyv2"
+)
+
+// template related to this adapter: service/templates/cloudformation/security_groups.yaml
```
fgimenez: nitpick, missing `guest` subdir here

rossf7:
Tested with `1.0.0` and `0.2.0` version bundles.

PR #610, opened 2018-01-08
conditional creation of S3 resources

fgimenez:

Before these changes, the S3 components were always created by the legacy resource, even though we had the S3 operatorkit resources in place. These changes enable tthe conditional creation of these components in legacy when we are not using cloudformation.

After enabling the proper usage of the operatorkit resources it looks like we hit again the same problem that we had with the fixed order of resource processing in operatorkit: we need s3bucketv2 before s3objectv2 on creation and s3objectv2 before s3bucketv2 on deletion. I've solved the problem by extending s3bucketv2 to be able to manage objects too, I don't quite like the approach (resources should be independent), but haven't found a better solution, other than merging the resources, let me know WDYT. Working for 0.1.0 and 0.2.0.

### DIFF
--- a/service/resource/s3bucketv2/delete.go
+++ b/service/resource/s3bucketv2/delete.go
@@ -18,7 +18,7 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 	if bucketInput.Name != "" {
 		r.logger.LogCtx(ctx, "debug", "deleting S3 bucket")
 
-		// make sure the bucket is empty.
+		// Make sure the bucket is empty.
 		input := &s3.ListObjectsV2Input{
 			Bucket: aws.String(bucketInput.Name),
 		}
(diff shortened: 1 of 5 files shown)


service/resource/s3bucketv2/delete.go
```diff
@@ -18,7 +18,26 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
 	if bucketInput.Name != "" {
 		r.logger.LogCtx(ctx, "debug", "deleting S3 bucket")
 
-		_, err := r.clients.S3.DeleteBucket(&s3.DeleteBucketInput{
+		// make sure the bucket is empty.
```
rossf7: Super nitpicky but I think the comments should start with a capital.

fgimenez: Sure, fixed :)

PR #611, opened 2018-01-08
make VPCGatewayAttachment depend on route tables

fgimenez:

This prevents errors like:
```
15:38:04 UTC+0100    CREATE_FAILED    AWS::EC2::Route    InternetGatewayRoute    route table rtb-41e68a2a and network gateway igw-83d05feb belong to different networks
```
Tested with 0.2.0


fgimenez:
Np, thank you for pointing to the right solution! :)

PR #613, opened 2018-01-09
WIP: migrate vpc and vpc peering to cloudformation

fgimenez:

Still WIP but hopefully not for too long, only missing the creation of the stack in the host account, it only has the cross-account access role but i think it's better to keep it as a stack for uniformity.

I've already added the `HostClients` configuration option, @rossf7 if you could take a look at the implementation and the name for the cross-account access role that would be great.


service/resource/legacyv2/adapter/adapter.go
```diff
@@ -61,14 +64,6 @@ func New(cfg Config) (Adapter, error) {
...
 	a.ClusterID = keyv2.ClusterID(cfg.CustomObject)
 
-	// TODO: remove this code once the VPC is created by cloudformation and add a
-	// reference in the template
-	vpcID, err := VPCID(cfg.Clients, keyv2.ClusterID(cfg.CustomObject))
-	if err != nil {
-		return Adapter{}, microerror.Mask(err)
-	}
-	a.VPCID = vpcID
-
```
rossf7: <3

service/resource/legacyv2/spec.go
```diff
@@ -9,7 +9,8 @@ const (
 	clusterVersionOutputKey = "ClusterVersionOutput"
 	workerRoleKey           = "WorkerRole"
 
-	cloudFormationTemplatesDirectory = "service/templates/cloudformation"
+	cloudFormationGuestTemplatesDirectory = "service/templates/cloudformation/guest"
+	cloudFormationHostTemplatesDirectory  = "service/templates/cloudformation/host"
```
rossf7: I like the separate directories for guest and host.

It just got me thinking that we'll need to version the cloud formation templates at some point. I think we can do that as part of the "update" changes rather than in the migration.

service/keyv2/key.go
```diff
@@ -119,6 +119,10 @@ func MasterInstanceType(customObject v1alpha1.AWSConfig) string {
 	return instanceType
 }
 
+func PeerAccessRoleName(customObject v1alpha1.AWSConfig) string {
+	return fmt.Sprintf("%s-peer-access", ClusterID(customObject))
```
rossf7: Maybe this could be `-vpc-peer-access`? A bit longer but makes it clear it's for the VPC peering.

service/templates/cloudformation/host/main.yaml
```diff
@@ -0,0 +1,24 @@
+{{define "main"}}AWSTemplateFormatVersion: 2010-09-09
+Description: Main CloudFormation stack.
+Resources:
+  PeerRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      RoleName: Peer-{{ .ClusterID }}
```
rossf7: Should this be using the key function?

service/templates/cloudformation/host/main.yaml
```diff
@@ -0,0 +1,24 @@
+{{define "main"}}AWSTemplateFormatVersion: 2010-09-09
+Description: Main CloudFormation stack.
```
rossf7: I think the description should say this is a host stack. Also it will need the route table entries for the guest cluster scraping.

service/templates/cloudformation/guest/vpc.yaml
```diff
@@ -0,0 +1,20 @@
...
+    Type: AWS::EC2::VPC
+    Properties:
+      CidrBlock: {{ .CidrBlock }}
+      EnableDnsSupport: 'true'
+      EnableDnsHostnames: 'true'
+      Tags:
+      - Key: Name
+        Value: {{ .ClusterID }}
+      - Key: Installation
+        Value: {{ .InstallationName }}
```
rossf7: Once we're migrated I'd like us to look at tagging. I think we can add some tags to the stack when we create it with the SDK. They will then propagate to all resources. I'd like to use it for the k8s cloud provider tags but maybe for the installation too.

service/resource/legacyv2/spec.go
fgimenez: Agree, makes more sense as part of update.

service/templates/cloudformation/guest/vpc.yaml
fgimenez: Great idea, will make things very easy. btw we should sync with @pipo02mix for this, he proposed https://github.com/giantswarm/aws-operator/pull/599, maybe that PR could be updated to accomplish what you suggest once the migration is done.

rossf7: Definitely, @pipo02mix I've seen your PR and it's great you're sorting out tagging. It will be easier to do this after the cloud formation migration.

service/templates/cloudformation/host/main.yaml
fgimenez: Indeed, description fixed (also the name of the stack object as seen from the console, if both accounts are the same we need to distinguish the stacks). Totally missed the route table entries you mention, will work on them next.

service/keyv2/key.go
fgimenez: Sure, done

service/templates/cloudformation/host/main.yaml
fgimenez: Yep, I'll add it to the rest of the host stack changes.

service/templates/cloudformation/host/main.yaml
rossf7: Great, thanks! We'll also need the `service` and `endpoint` k8s resources for the guest cluster scraping. It has quite a lot of moving parts! We can do these as later PRs.

fgimenez: @rossf7 Host route tables are in place, could you please take a look? This is the template https://github.com/giantswarm/aws-operator/pull/613/files#diff-ed833ccba861d6b81fbcab204ebf07fd and this the adapter https://github.com/giantswarm/aws-operator/pull/613/files#diff-b2ba10ced5c9cb6d1388bf3062a520bf

service/templates/cloudformation/host/route_tables.yaml
```diff
@@ -0,0 +1,17 @@
+{{define "route_tables"}}
+  {{ range .RouteTables }}
+  PrivateRouteTable-{{.Name}}:
+    Type: AWS::EC2::RouteTable
+    Properties:
+      VpcId: {{.PeerVPCID}}
+      Tags:
+      - Key: Name
+        Value: {{ .Name }}
```
rossf7: The route tables aren't needed. They are created in the host cluster setup. We just need to create the routes.

service/templates/cloudformation/host/main.yaml
```diff
@@ -0,0 +1,6 @@
+{{define "main"}}AWSTemplateFormatVersion: 2010-09-09
+Description: Main Host CloudFormation stack.
+Resources:
+  {{template "iam_roles" .}}
+  {{template "route_tables" .}}
```
rossf7: I really like the template structure and good to see it in the host stack too! <3

fgimenez:
@rossf7 I've pushed the changes we talked about to create the pre-guest host stack (with the cross-account access role) and the post-guest host stack (with the peering routes), tested successfully on 0.1.0 and 0.2.0 (without using different host and guest accounts though), pls take another look.

service/keyv2/key.go
```diff
@@ -105,10 +105,22 @@ func LoadBalancerName(domainName string, cluster v1alpha1.AWSConfig) (string, er
...
 	clusterID := ClusterID(customObject)
 
-	return fmt.Sprintf("%s-main", clusterID)
+	return fmt.Sprintf("%s-guest-main", clusterID)
+}
+
+func MainHostPreStackName(customObject v1alpha1.AWSConfig) string {
+	clusterID := ClusterID(customObject)
+
+	return fmt.Sprintf("%s-host-pre-main", clusterID)
```
rossf7: This had to be split but I'm not sure the pre and post names are intuitive. What about if we called the pre-stack `%s-host-setup` and the post stack `%s-host-main`?

Maybe in the stack descriptions we can say the stacks need to run pre and post the guest stack?

service/resource/legacyv2/adapter/host_route_tables.go
```diff
@@ -0,0 +1,127 @@
...
+
+	for _, routeTableName := range cfg.CustomObject.Spec.AWS.VPC.RouteTableNames {
+		routeTableID, err := routeTableID(routeTableName, cfg)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		rt := RouteTable{
+			Name:         routeTableName,
+			RouteTableID: routeTableID,
+			// Requester CIDR block, we create the peering connection from the guest's private subnet.
```
rossf7: The comments are great because the peering is confusing! <3

service/templates/cloudformation/guest/route_tables.yaml
```diff
@@ -2,15 +2,15 @@
   PublicRouteTable:
     Type: AWS::EC2::RouteTable
     Properties:
-      VpcId: {{ .VPCID }}
+      VpcId: !Ref VPC
```
rossf7: <3 all these Ref's :)

service/resource/legacyv2/resource.go
```diff
@@ -1652,10 +1673,23 @@ func (s *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 
 		_, err = s.awsClients.CloudFormation.CreateStack(&stackInput)
 		if err != nil {
-			return err
+			return microerror.Mask(err)
+		}
+		err = s.awsHostClients.CloudFormation.WaitUntilStackCreateComplete(&cloudformation.DescribeStacksInput{
+			StackName: stackInput.StackName,
+		})
```
rossf7: I see why we need to wait but I have a concern we'll block the operator for a long time. Due to the issue with the framework not processing resources in parallel.

Could we launch the post stack and have it wait for the main stack resources?

service/keyv2/key.go
fgimenez: Sure, much better, done

service/resource/legacyv2/resource.go
fgimenez: Indeed this can be a problem, also we currently don't get any feedback about what's going on during any stack creation. I think the host pre and post stack are not a real problem, the creation is fast, but yes, we should do something to address the sequential launch.

Agree to try to resolve it once the resources are split.

PR #615, opened 2018-01-10
Use KMSKeyState when creating and deleting keys

rossf7:
Fixes problems with creating and deleting keys that I missed in my previous PR.

I've also changed the resource methods to use the KMSKeyState struct rather than the AWS SDK structs. I think this is simpler and matches what we do for the S3 bucket.

### DIFF
--- a/service/resource/kmskeyv2/create.go
+++ b/service/resource/kmskeyv2/create.go
@@ -6,30 +6,22 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/giantswarm/microerror"
-
-	"github.com/giantswarm/aws-operator/service/keyv2"
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	createKeyInput, err := toCreateKeyInput(createChange)
+	createInput, err := toKMSKeyState(createChange)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	if createKeyInput != nil {
-		customObject, err := keyv2.ToCustomObject(obj)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-		key, err := r.awsClients.KMS.CreateKey(createKeyInput)
+	if createInput.KeyAlias != "" {
+		key, err := r.awsClients.KMS.CreateKey(&kms.CreateKeyInput{})
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
-		clusterID := keyv2.ClusterID(customObject)
-		keyAlias := toAlias(clusterID)
 		if _, err := r.awsClients.KMS.CreateAlias(&kms.CreateAliasInput{
-			AliasName:   aws.String(keyAlias),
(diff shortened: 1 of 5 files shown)


service/resource/kmskeyv2/delete.go
```diff
@@ -10,24 +10,27 @@ import (
...
 	}
 
-	keyAlias := deleteAliasInput.AliasName
-	if keyAlias != nil && *keyAlias != "" {
-		if _, err := r.awsClients.KMS.DeleteAlias(&deleteAliasInput); err != nil {
-			return microerror.Mask(err)
-		}
-
+	if deleteInput.KeyAlias != "" {
+		// Get the KMS key ID using the key alias.
```
rossf7: Now we get the key Arn using the alias before deleting it. So we can be sure we have the right value.

service/resource/kmskeyv2/create.go
```diff
@@ -50,19 +42,19 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
 
-	desiredKMSState, err := toKMSKeyState(desiredState)
+	desiredKeyState, err := toKMSKeyState(desiredState)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	if desiredKMSState.KeyAlias != currentKMSState.KeyAlias {
-		return &kms.CreateKeyInput{}, nil
+	if currentKeyState.KeyAlias == "" || desiredKeyState.KeyAlias != currentKeyState.KeyAlias {
```
rossf7: Conditional now matches the s3ucket resource and I think is more robust.

service/resource/kmskeyv2/delete.go
```diff
@@ -57,22 +60,22 @@ func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desire
...
+		return nil, microerror.Mask(err)
 	}
 
 	r.logger.LogCtx(ctx, "debug", "finding out if the KMS key should be deleted")
 
-	if currentKeyState.KeyAlias != "" && desiredKeyState.KeyAlias != currentKeyState.KeyAlias {
-		deleteChange.AliasName = aws.String(currentKeyState.KeyAlias)
+	var kmsKeyToDelete KMSKeyState
+	if currentKeyState.KeyAlias != "" {
+		kmsKeyToDelete = desiredKeyState
```
rossf7: Delete check also updated to be consistent with the s3bucket resource.

service/resource/kmskeyv2/create.go
```diff
@@ -6,30 +6,22 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/giantswarm/microerror"
-
-	"github.com/giantswarm/aws-operator/service/keyv2"
 )
 
 func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	createKeyInput, err := toCreateKeyInput(createChange)
+	createInput, err := toKMSKeyState(createChange)
```
fgimenez: :+1:

PR #617, opened 2018-01-10
Set ELB healthy threshold to 2 seconds

r7vme:
Right now restart of kubernetes api causes 50 seconds
downtime. Because 10 tries x 5sec = 50 sec.

This change reduces number of tries to 2 before marking backend
as healthy.

I tested 2 tries threshold manually in ginger.


calvix:
make sense

PR #619, opened 2018-01-11
Add k8s service resource

rossf7:

Adds the service resource for guest cluster scraping. I'll enable once the endpoint resource is ready.


service/resource/namespacev2/resource.go
```diff
@@ -13,14 +13,14 @@ const (
 	Name = "namespacev2"
 )
 
-// Config represents the configuration used to create a new deployment resource.
+// Config represents the configuration used to create a new namespace resource.
```
fgimenez: <3

service/resource/servicev2/create_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/micrologger/microloggertest"
+	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
+)
+
+func Test_Resource_Service_newCreateChange(t *testing.T) {
+	testCases := []struct {
+		Obj             interface{}
```
fgimenez: Could we have these struct members as private? A description field here would be nice too, aligning with `desired_test`

service/resource/servicev2/create_test.go
```diff
@@ -0,0 +1,108 @@
...
+		resourceConfig.K8sClient = fake.NewSimpleClientset()
+		resourceConfig.Logger = microloggertest.New()
+		newResource, err = New(resourceConfig)
+		if err != nil {
+			t.Fatal("expected", nil, "got", err)
+		}
+	}
+
+	for i, tc := range testCases {
+		result, err := newResource.newCreateChange(context.TODO(), tc.Obj, tc.Cur, tc.Des)
```
fgimenez: Also like `desired_test`, it would be nice to have subtests here.

service/resource/servicev2/create_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			t.Fatal("case", i+1, "expected", nil, "got", err)
+		}
+		if tc.ExpectedService == nil {
+			if tc.ExpectedService != result {
+				t.Fatal("case", i+1, "expected", tc.ExpectedService, "got", result)
+			}
+		} else {
+			name := result.(*apiv1.Service).Name
```
fgimenez: Could we check the type assertion here?

service/resource/servicev2/delete_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/micrologger/microloggertest"
+	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
+)
+
+func Test_Resource_Service_newDeleteChange(t *testing.T) {
+	testCases := []struct {
+		Obj             interface{}
```
fgimenez: Same comments as in `create_test`

service/resource/servicev2/delete.go
```diff
@@ -0,0 +1,76 @@
...
+	desiredService, err := toService(desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "debug", "finding out if the service has to be deleted")
+
+	var serviceToDelete *apiv1.Service
+	if currentService != nil {
+		serviceToDelete = desiredService
```
fgimenez: Shouldn't we delete the current service?

service/resource/servicev2/delete_test.go
```diff
@@ -0,0 +1,108 @@
...
+					Name: "al9qy",
+				},
+			},
+			Des: &apiv1.Service{
+				TypeMeta: apismetav1.TypeMeta{
+					Kind:       "Service",
+					APIVersion: "v1",
+				},
+				ObjectMeta: apismetav1.ObjectMeta{
+					Name: "al9qy",
```
fgimenez: Maybe another test case with different and not nil current and desired state would be good.

service/resource/servicev2/create_test.go
rossf7: Good point, fixed. These were copy paste from another resource.

service/resource/servicev2/create_test.go
rossf7: Done and yes I think we should use subtests in more projects.

service/resource/servicev2/delete_test.go
rossf7: Good point, done.

service/resource/servicev2/delete.go
rossf7: I think it should be the desired state but I changed the checks in newCreateChange and newDeleteChange to also check the service name.

rossf7:
@fgimenez Changes made. PTAL

PR #620, opened 2018-01-11
Split legacy and cloudformation resources

fgimenez:

Not tested yet, will od tomorrow @rossf7 if you could have a quick look that would be great, do we need `cloudformationv1`?


service/resource/cloudformationv2/create.go
```diff
@@ -0,0 +1,90 @@
...
+
+	desiredStackState, err := toStackState(desiredState)
+	if err != nil {
+		return cloudformation.CreateStackInput{}, microerror.Mask(err)
+	}
+
+	r.logger.LogCtx(ctx, "debug", "finding out if the main stack should be created")
+
+	createState := cloudformation.CreateStackInput{
+		StackName: aws.String(""),
```
rossf7: Why is the stack name empty here?

service/resource/cloudformationv2/create.go
```diff
@@ -0,0 +1,90 @@
...
+	return nil
+}
+
+func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error) {
+	customObject, err := keyv2.ToCustomObject(obj)
+	if err != nil {
+		return cloudformation.CreateStackInput{}, microerror.Mask(err)
+	}
+
+	desiredStackState, err := toStackState(desiredState)
```
rossf7: Not a blocker but I wonder if we need to update the StackState struct? So if covers all 3 stacks.

service/resource/cloudformationv2/desired.go
```diff
@@ -0,0 +1,23 @@
...
+func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
+	customObject, err := keyv2.ToCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	mainStack, err := newMainStack(customObject)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
rossf7: Also here I'd expect the desired state to include all 3 stacks.

service/resource/cloudformationv2/update_test.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			updateChange, ok := result.(awscloudformation.UpdateStackInput)
+			if !ok {
+				t.Errorf("expected '%T', got '%T'", updateChange, result)
+			}
+			/*
+				if updateChange.StackName != nil && *updateChange.StackName != *tc.expectedChange.StackName {
+					t.Errorf("expected %v, got %v", *tc.expectedChange.StackName, *updateChange.StackName)
+				}
+			*/
```
rossf7: Should this be commented out? If so can you add a comment when it can be enabled

service/resource/cloudformationv2/create.go
fgimenez: Not sure, do we need to update the host stacks? Currently we have there the cross-account access  role (only depends on the guest account id) and the peering routes (depend on the host VPC and the VPCPeeringID). As you suggest maybe better discuss it in a separate issue, will create it.

service/resource/cloudformationv2/create.go
fgimenez: Good catch! not needed, removed here and in update too

service/resource/cloudformationv2/desired.go
fgimenez: Yep, let's discuss in the related issue, will link here when created

service/resource/cloudformationv2/update_test.go
fgimenez: Sorry, i removed this while doing a huge rebase, the test was failing but looks to me like the test case was wrong, if current is empty and desired is not, the update change should be empty (the create change should not), could you please have another look?

service/resource/cloudformationv2/desired.go
fgimenez: Here it is https://github.com/giantswarm/aws-operator/issues/621

service/resource/cloudformationv2/create.go
rossf7: Good point if there is nothing to change then it doesn't need to be in the stack state. I think later we might have to add more to the main host stack but we could do that later.

Yes let's discuss more in the issue.

service/resource/cloudformationv2/update_test.go
rossf7: No worries and yes I agree the test is wrong. This case should be a create change.

PR #623, opened 2018-01-12
Add endpoints resource

rossf7:

TODO: Implement update and delete framework methods.


service/resource/endpointsv2/desired.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	instanceName := keyv2.MasterInstanceName(customObject)
+	masterInstance, err := r.findMasterInstance(instanceName)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
```
rossf7: Here is the desired state with the private IP of the master.

service/resource/endpointsv2/instance.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/giantswarm/microerror"
+)
+
+const (
+	// http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#InstanceState
+	ec2RunningState = 16
+	tagKeyName      = "Name"
+)
+
```
rossf7: Uses the EC2 API to find the master instance using its name label.

service/resource/servicev2/create_test.go
```diff
@@ -34,7 +34,7 @@ func Test_Resource_Service_newCreateChange(t *testing.T) {
 					APIVersion: "v1",
 				},
 				ObjectMeta: apismetav1.ObjectMeta{
-					Name: "al9qy",
+					Name: "master",
```
rossf7: Some fixes for the service tests to use the correct name.

PR #627, opened 2018-01-14
Add missing route for VPC peering

rossf7:

Adds missing route to the guest private route table for VPC peering. Route is from the host cluster CIDR to the VPC peering connection.


rossf7:
@JosephSalisbury @puja108 Added you as a reviewer as Fede is AFK. This route got missed and means SSH access wasn't working.

PR #628, opened 2018-01-15
Setting memory request equal to limit

marians:
We want aws-operator scheduling to be guaranteed, right? So the memory request should be equal to limits.

For discussion: Should we increase the request and limit? We had a case of OOM killing here: https://github.com/giantswarm/giantswarm/issues/2364


JosephSalisbury:
I feel like if the aws-operator is using more than 250Mi, something is wrong in the application. @rossf7 thoughts?

marians:
Let's discuss the OOM part in https://github.com/giantswarm/giantswarm/issues/2364

Issue #508, opened 2017-11-17
etcd data should be stored on separate block device


r7vme:
We should use separate EBS volume for etcd data, right now it's stored on root device.

```
core@ip-172-19-66-205 ~ $ lsblk 
NAME    MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
xvda    202:0    0    8G  0 disk  
|-xvda6 202:6    0  128M  0 part  /usr/share/oem
|-xvda4 202:4    0    1G  0 part  
|-xvda2 202:2    0    2M  0 part  
|-xvda9 202:9    0  5.7G  0 part  /
|-xvda7 202:7    0   64M  0 part  
|-xvda3 202:3    0    1G  0 part  
| `-usr 254:0    0 1016M  1 crypt /usr
`-xvda1 202:1    0  128M  0 part  /boot
xvdb    202:16   0 37.5G  0 disk  /var/lib/docker

```

rossf7:
This will be done as part of the Cloud Formation migration in giantswarm/giantswarm#2023

PR #631, opened 2018-01-17
create and setup etcd EBS volume

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2023 https://github.com/giantswarm/aws-operator/issues/508

First tests worked but still not verified that the volume was properly mounted and used.

### DIFF
--- a/integration/operator_test.go
+++ b/integration/operator_test.go
@@ -3,6 +3,7 @@
 package integration
 
 import (
+	"fmt"
 	"html/template"
 	"log"
 	"os"
@@ -81,6 +82,7 @@ func newAWSClient() aWSClient {
 
 var (
 	f *framework
+	c aWSClient
 )
 
 // TestMain allows us to have common setup and teardown steps that are run
@@ -93,6 +95,8 @@ func TestMain(m *testing.M) {
 		log.Printf("unexpected error: %v\n", err)
 		os.Exit(1)
 	}
+
+	c = newAWSClient()
 
 	if err := f.SetUp(); err != nil {
 		log.Printf("unexpected error: %v\n", err)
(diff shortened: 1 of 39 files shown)


service/templates/cloudformation/guest/instance.yaml
```diff
@@ -16,4 +16,16 @@
...
+    Properties:
+      Size: 100
+      VolumeType: gp2
+      AvailabilityZone: !GetAtt MasterInstance.AvailabilityZone
+  MountPoint:
+    Type: AWS::EC2::VolumeAttachment
+    Properties:
+      InstanceId: !Ref MasterInstance
+      VolumeId: !Ref EtcdVolume
+      Device: /dev/sdh
```
rossf7: Could you add a name tag to the volume?

fgimenez: Sure done

fgimenez:
@rossf7 all good now with the mount unit, pls take another look. 

I've taken the opportunity to add an e2e test that reboots the master instance and checks that the guest is ready (API up and nodes running) after a while, a successful log now looks like this https://circleci.com/gh/giantswarm/aws-operator/2681. This is actually the first e2e test (what we had so far was just test preparation) and IMO can be useful not only for stating that the etcd volume is properly mounted after reboot, let me know WDYT.

fgimenez:
Etcd volume is still not mounting properly, we need a unit to format the volume before mounting. Now extending the test to uncover the failure and then adding the code to fix it.

integration/operator_test.go
```diff
@@ -115,7 +119,54 @@ func TestMain(m *testing.M) {
...
+			masterID,
+		},
+	}
+	_, err = c.EC2.RebootInstances(rebootInput)
+	if err != nil {
+		t.Errorf("unexpected error rebooting  master %v", err)
+	}
+
+	if err := f.WaitForAPIDown(); err != nil {
+		t.Errorf("unexpected error wainting for master shutting down %v", err)
```
rossf7: typo on waiting

integration/operator_test.go
```diff
@@ -115,7 +119,54 @@ func TestMain(m *testing.M) {
...
+	}
+	res, err := c.EC2.DescribeInstances(describeInput)
+	if err != nil {
+		t.Errorf("unexpected error getting master id %v", err)
+	}
+	if len(res.Reservations) != 1 {
+		t.Errorf("unexpected number of reservations %d", len(res.Reservations))
+	}
+	if len(res.Reservations[0].Instances) != 1 {
+		t.Errorf("unexpected number of reservations %d", len(res.Reservations[0].Instances))
```
rossf7: nitpick but should error text say instances?

fgimenez:
All working fine with latest changes, the state is kept between reboots.

PR #637, opened 2018-01-19
Add 2.0.0 WIP version bundle

rossf7:

Adds a `2.0.0` version bundle to replace the current `0.2.0`. Since the new version is a rewrite of the operator.

- Bundle is still marked WIP.
- We'll need to update the changelog once the final changes for k8scloudconfig v3 are in place.


service/version_bundles.go
```diff
@@ -156,5 +156,61 @@ func newVersionBundles() []versionbundle.Bundle {
 			Version:      "1.0.0",
 			WIP:          false,
 		},
+		{
+			Changelogs: []versionbundle.Changelog{
+				{
```
rossf7: We'll need to update the changelog since we're adding more features to k8scloudconfig v3.

service/version_bundles.go
```diff
@@ -156,5 +156,61 @@ func newVersionBundles() []versionbundle.Bundle {
...
+					Name:    "nginx-ingress-controller",
+					Version: "0.9.0",
+				},
+			},
+			Dependencies: []versionbundle.Dependency{},
+			Deprecated:   false,
+			Name:         "aws-operator",
+			Time:         time.Date(2018, time.January, 19, 15, 00, 0, 0, time.UTC),
+			Version:      "2.0.0",
```
rossf7: Version is 2.0.0 because we've rewritten the operator! 😄

service/resource_router_test.go
```diff
@@ -269,7 +269,20 @@ func Test_Service_NewTPRResourceRouter(t *testing.T) {
 			errorMatcher:      nil,
 			resourceRouter:    NewTPRResourceRouter(versionedResources),
 		},
-		// Case 4. Invalid version returns an error.
+		// Case 4. Version 2.0.0 returns all resources except legacy.
```
fgimenez: Feel free to ignore, can we add a description field here with the comment contents? And subtests using that description field.

rossf7: I updated the TPR resource router by mistake which doesn't use subtests. It should be frozen until we delete the TPR code.

service/version_bundles.go
```diff
@@ -156,5 +156,61 @@ func newVersionBundles() []versionbundle.Bundle {
...
+					Kind:        versionbundle.KindChanged,
+				},
+			},
+			Components: []versionbundle.Component{
+				{
+					Name:    "calico",
+					Version: "3.0.1",
+				},
+				{
+					Name:    "docker",
```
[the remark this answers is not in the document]
rossf7: I updated docker to `17.09.0` it reports as `17.09.0-ce` but alpha chars aren't allowed in version bundles currently.

I also opened #642 to move the AMI to the operator from the custom object.

service/version_bundles.go
```diff
@@ -156,5 +156,61 @@ func newVersionBundles() []versionbundle.Bundle {
...
+					Version: "1.12.6",
+				},
+				{
+					Name:    "etcd",
+					Version: "3.2.7",
+				},
+				{
+					Name:    "kubedns",
```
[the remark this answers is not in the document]
rossf7: Good catch, fixed.

rossf7:
@puja108 Changes made. Could you take another look at the version bundle?

PR #644, opened 2018-01-23
add master outputs and rename current ones

fgimenez:

Introduced new fields for master's imageID and CloudConfig version, also removed ClusterVersion and added WorkerCloudConfig version and renamed worker-related fields. I've taken the opportunity to promote master and worker image ID from `launch_config` and `instance` to the main adapter.

Still not doing phased updates, the complete stack is updated as a whole.


service/templates/cloudformation/guest/outputs.yaml
```diff
@@ -1,9 +1,13 @@
 {{define "outputs"}}
 Outputs:
-  WorkersOutput:
+  MasterImageID:
+    Value: {{ .MasterImageID }}
+  MasterCloudConfigVersion:
+    Value: {{ .MasterCloudConfigVersion }}
+  Workers:
```
rossf7: I think WorkerCount would be a clearer name?

service/templates/cloudformation/guest/outputs.yaml
```diff
@@ -1,9 +1,13 @@
 {{define "outputs"}}
 Outputs:
-  WorkersOutput:
+  MasterImageID:
+    Value: {{ .MasterImageID }}
```
rossf7: I think we should add outputs for master and worker instance types.

We don't support changing instance types yet but it should be straight forward to implement later. WDYT?

service/templates/cloudformation/guest/outputs.yaml
fgimenez: Much better, done.

service/templates/cloudformation/guest/outputs.yaml
fgimenez: Totally, already added, PTAL.

PR #645, opened 2018-01-23
Adds kube-state-metrics port

JosephSalisbury:


JosephSalisbury:
That was my thinking! :D

JosephSalisbury:
Tested on `gauss`, guest cluster master and worker security groups both have port 10301 open from the host vpc.

PR #648, opened 2018-01-23
Updates vendor with k8s 1.9.2

corest:


corest:
Skip requests for review, pending before k8scc merge

PR #652, opened 2018-01-24
Update alerter service to use apiextensions

rossf7:
Towards #640 

Updates the `alerter` service to use CRDs. It checks for orphan VPCs that don't have a CRO associated with them. We're not fully using it but could be useful until we have finalizer support.


rossf7:
Yes, this one needs more context. The alerter service is only used by us internally to detect orphan resources. So it isn't versioned. It wasn't updated to use CRDs which is why this PR was needed.

I don't think we need to alert for TPRs. We should delete all resources associated with legacy clusters once customers are migrated off them.

PR #660, opened 2018-01-24
Cloud provider tags

rossf7:

This PR adds the new cloud provider tags to all 3 stacks and they propagate to the child resources. Tested scaling and creating LBs and PVCs and looks good.

I'd still like to tag the KMS key and S3 bucket as they aren't managed by Cloud Formation. But I'll do that as a later change.

screen shot 2018-01-24 at 4 45 07 pm

screen shot 2018-01-24 at 4 45 29 pm

### DIFF
--- a/service/keyv2/key.go
+++ b/service/keyv2/key.go
@@ -15,6 +15,13 @@ const (
 	// TODO Remove once the migration is complete.
 	CloudFormationVersion = "0.2.0"
 
+	// CloudProviderTagName is used to add Cloud Provider tags to AWS resources.
+	CloudProviderTagName = "kubernetes.io/cluster/%s"
+
+	// CloudProviderTagOwnedValue is used to indicate an AWS resource is owned
+	// and managed by a cluster.
+	CloudProviderTagOwnedValue = "owned"
+
 	// LegacyVersion is the version in the version bundle for existing clusters.
 	LegacyVersion = "0.1.0"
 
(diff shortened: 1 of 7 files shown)


service/keyv2/key.go
```diff
@@ -54,6 +61,15 @@ func ClusterNamespace(customObject v1alpha1.AWSConfig) string {
...
 
+func ClusterTags(customObject v1alpha1.AWSConfig) map[string]string {
+	cloudProviderTag := fmt.Sprintf(CloudProviderTagName, ClusterID(customObject))
+	tags := map[string]string{
+		cloudProviderTag: CloudProviderTagOwnedValue,
+	}
+
+	return tags
+}
+
```
rossf7: This key helper returns a map of tags so we can extend easily later.

service/resource/cloudformationv2/create.go
```diff
@@ -94,6 +94,8 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
 		createState.Capabilities = []*string{
 			aws.String(namedIAMCapability),
 		}
+
+		createState.SetTags(getCloudFormationTags(customObject))
```
rossf7: We tag all 3 stacks and resources propagate to the child resources.

service/templates/cloudformation/guest/auto_scaling_group.yaml
```diff
@@ -17,9 +17,6 @@
         - Key: Name
           Value: {{ .ClusterID }}-{{ .ASGType }}
           PropagateAtLaunch: true
-        - Key: KubernetesCluster
-          Value: {{ .ClusterID }}
-          PropagateAtLaunch: true
```
rossf7: Old tags are removed.

fgimenez: Nice! I've just realized that we need an additional cluster tag in the public subnet, otherwise the services of type LoadBalancer won't work. This change makes that the subnets will be tagged too, is that correct?

rossf7:
@fgimenez Yes subnets will get tagged too with this which is really nice.

rossf7:
Created #662 to tag the remaining resources.

PR #661, opened 2018-01-24
Update to latest deps (AKA big ball of mud)

rossf7:
Towards #640 

This is the big change to get us on the latest deps. I used `giantswarm/certs/legacy` for everything since this is blocking folk. 

I still need to update cloudconfigv3 and s3objectv2 to use the new interface. But I'd like to do that in a later PR.


glide.yaml
```diff
@@ -2,36 +2,54 @@ package: github.com/giantswarm/aws-operator
 import:
 - package: github.com/aws/aws-sdk-go
   version: v1.12.3
+- package: google.golang.org/grpc
+  version: v1.2.1
+- package: github.com/stretchr/testify
+  version: v1.1.4
```
rossf7: Some boyscouting to put external deps first

glide.yaml
```diff
@@ -2,36 +2,54 @@ package: github.com/giantswarm/aws-operator
...
 - package: github.com/giantswarm/randomkeytpr
   version: master
 - package: github.com/giantswarm/e2e-harness
   version: master
-- package: github.com/giantswarm/apiextensions
-  version: clientset
+- package: github.com/giantswarm/versionbundle
+  version: master
+
+# Pinned to version using glide and compatible with pinned version of backoff.
```
rossf7: This is from the kubernetesd PR earlier.

service/framework.go
```diff
@@ -54,9 +51,9 @@ func newCRDFramework(config Config) (*framework.Framework, error) {
 
 	var err error
 
-	var k8sExtClient apiextensionsclient.Interface
+	var restConfig *rest.Config
```
rossf7: This should be the state of the art for creating the framework. But would appreciate a 2nd opinion.

rossf7:
@kopiczko Will do! For now I'm not going to migrate the `legacyv2` resource to the new `certs` structure.

I hope we can move all installs off the legacy clusters quickly. But if that drags we can also migrate `legacyv2`.

PR #664, opened 2018-01-24
Update changelog for CoreDNS

rossf7:
Makes the changelog look nicer and updates CoreDNS to 1.0.4.

@corest I know your change isn't merged yet but wanted to do this before I forgot!


service/version_bundles.go
```diff
@@ -185,7 +185,7 @@ func newVersionBundles() []versionbundle.Bundle {
 				},
 				{
 					Component:   "CoreDNS",
-					Description: "CoreDNS 1.0.1 replaces kube-dns.",
+					Description: "Version 1.0.4 replaces kube-dns.",
```
rossf7: Makes the changelog look nicer and updates for 1.0.4.

rossf7:
@marians FYI changelog formatting is fixed for CoreDNS. Thanks for raising!

marians:
Nice! Always cool to see a suggestion be picked up that fast. Thanks.

Branch can be deleted, right?

rossf7:
Branch is gone :)

PR #665, opened 2018-01-24
Update vendor for nginx-ingress-controller 0.10.2 and coredns 1.0.5

corest:


corest:
CoreDNS is updated and there is a changelog for IC

rossf7:
@corest Sorry can you check again.  I still see `1.0.4` here.

https://github.com/giantswarm/aws-operator/blob/ceecf0f2c1b9deab10e444049f19dfaee83bbe96/service/version_bundles.go#L207

Also nothing in the 2.0.0 changelog for IC.

https://github.com/giantswarm/aws-operator/blob/ceecf0f2c1b9deab10e444049f19dfaee83bbe96/service/version_bundles.go#L160

PR #667, opened 2018-01-25
Use alpine base image

rossf7:
I've no idea why we're using `busybox:ubuntu-14.04` as the base image. Updating to use the latest alpine image.


rossf7:
Tested locally and with e2e tests.

PR #668, opened 2018-01-25
Specify Container Linux AMI in the operator

### LINKED ISSUE #642: CoreOS AMI should be set in the operator to allow upgrades

The CoreOS AMI for the master and worker nodes should be set by the operator rather than come from the custom object.

This is so it can be versioned via the version bundle. This also allows us to include the docker version in the changelog.

rossf7:
Fixes #642 

This change specifies the Container Linux AMI in the operator rather than it coming from `installations`. This is because we want to.

- Control upgrades via version bundles.
- Include the Docker Engine version in our releases which comes from the Container Linux version.
- Include the Container Linux version in our release changelogs.

### DIFF
--- a/service/resource/cloudformationv2/adapter/adapter.go
+++ b/service/resource/cloudformationv2/adapter/adapter.go
@@ -139,8 +139,15 @@ func NewHostPost(cfg Config) (Adapter, error) {
 func getImageID(customObject v1alpha1.AWSConfig) (string, error) {
 	region := keyv2.Region(customObject)
 
-	// Container Linux AMIs for each active region.
-	// NOTE: AMIs should always be for HVM virtualisation and not PV.
+	/*
+		Container Linux AMIs for each active AWS region.
+		From: https://coreos.com/os/docs/latest/booting-on-ec2.html
+
+		NOTE 1: AMIs should always be for HVM virtualisation and not PV.
+		NOTE 2: You also need to update adapter_test.go.
+
+		Current Release: CoreOS Container Linux stable 1576.5.0 (HVM)
+	*/
 	imageIDs := map[string]string{
 		"eu-central-1": "ami-90c152ff",
 		"eu-west-1":    "ami-32d1474b",
--- a/service/resource/cloudformationv2/adapter/adapter.go
+++ b/service/resource/cloudformationv2/adapter/adapter.go
@@ -70,8 +70,15 @@ func NewGuest(cfg Config) (Adapter, error) {
 
 	a.ASGType = prefixWorker
 	a.ClusterID = keyv2.ClusterID(cfg.CustomObject)
-	a.MasterImageID = keyv2.MasterImageID(cfg.CustomObject)
-	a.WorkerImageID = keyv2.WorkerImageID(cfg.CustomObject)
+
+	// Get the EC2 AMI for the configured region.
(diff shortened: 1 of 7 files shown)


service/resource/cloudformationv2/adapter/adapter.go
```diff
@@ -127,3 +134,23 @@ func NewHostPost(cfg Config) (Adapter, error) {
 
 	return a, nil
 }
+
```
rossf7: Here we get the AMI for the region in the custom object.

service/resource/cloudformationv2/adapter/adapter.go
```diff
@@ -127,3 +134,23 @@ func NewHostPost(cfg Config) (Adapter, error) {
...
+func getImageID(customObject v1alpha1.AWSConfig) (string, error) {
+	region := keyv2.Region(customObject)
+
+	// Container Linux AMIs for each active region.
+	// NOTE: AMIs should always be for HVM virtualisation and not PV.
+	imageIDs := map[string]string{
+		"eu-central-1": "ami-90c152ff",
+		"eu-west-1":    "ami-32d1474b",
+		"us-west-2":    "ami-dc4ce6a4",
+	}
```
rossf7: Here we need to specify the AMIs. I think it makes sense to have this in the main adapter as this is where we set these values.

I think we should only specify values for active regions. The error message is clear and we can add more as we expand to more regions. WDYT?

rossf7:
I think this is the best place to specify the AMIs in the operator. I think the update process won't be too painful but I'd like to get feedback on it.

service/resource/cloudformationv2/adapter/adapter.go
puja108: Should we add here in the comment the CoreOS version and link to https://coreos.com/os/docs/latest/booting-on-ec2.html ? This way it is easier to know in the code what those ami's stand for (without having to look into the version bundle) and its also easy to verify.

Alternatively we could also just link directly to the versioned link like following:
https://coreos.com/os/docs/1590.2.0/booting-on-ec2.html

Just some ideas as "documentation in code"

service/resource/cloudformationv2/adapter/adapter_test.go
```diff
@@ -24,28 +24,78 @@ var (
...
 					},
 				},
-				Workers: []v1alpha1.AWSConfigSpecAWSNode{
-					v1alpha1.AWSConfigSpecAWSNode{
-						ImageID: "ami-test-worker",
+			},
+			errorMatcher:      nil,
+			expectedASGType:   "worker",
+			expectedClusterID: "test-cluster",
+			expectedImageID:   "ami-90c152ff",
```
puja108: so this test needs to be updated with each CoreOS update, right? can we note that somewhere?

service/resource/cloudformationv2/adapter/adapter.go
corest: Any PR for changing `ami`s will look like we are changing some logic or so, because of file being updated :/
But still don't see any better place for that

rossf7:
@puja108 @corest Thanks for the great feedback. I've added more comments as suggested.

We can iterate on this and I think we will need to develop tooling to make this smoother at some point.

Issue #662, opened 2018-01-24
Tag KMS key and S3 bucket with cloud provider tags


rossf7:
All cloud formation stacks were tagged in #656 

This is to also tag the KMS key and S3 bucket. These aren't managed by cloud formation because you can't create S3 objects in templates.

rossf7:
Fernando, just switching assignments. Thanks again for picking this up.

PR #671, opened 2018-01-25
Add cluster tag to the KMS keys

pipo02mix:
Towards https://github.com/giantswarm/aws-operator/issues/662

Add cluster tags to the KMS keys too


service/resource/kmskeyv2/resource.go
```diff
@@ -3,9 +3,14 @@ package kmskeyv2
 import (
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
```
fgimenez: Could you pls put together all the gs imports in a separate block?

service/resource/kmskeyv2/spec.go
```diff
@@ -18,8 +18,9 @@ type Clients struct {
 type KMSClient interface {
 	CreateKey(*kms.CreateKeyInput) (*kms.CreateKeyOutput, error)
 	CreateAlias(*kms.CreateAliasInput) (*kms.CreateAliasOutput, error)
+	DescribeKey(*kms.DescribeKeyInput) (*kms.DescribeKeyOutput, error)
```
fgimenez: <3

service/resource/kmskeyv2/resource.go
pipo02mix: I am not sure what do you mean. I thought we set first built-in go libraries, then external from the project, and finally packages from the same project). I have seen it in others files too, so I am bit confuse

rossf7: Yes the convention is the 3rd block has packages from the same project. So this looks correct to me.

fgimenez: Indeed, didn't notice the external gs projects and assumed that all the gs imports came from this one, sorry about that :)

PR #676, opened 2018-01-26
add CF cluster validation

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2390

Adds simple structure for checking multiple validation conditions before cluster creation, aborting if any of them fails. Also adds an initial check for existing peering routes in the host accepter VPC route tables.

Still missing `Alerter` changes suggested in https://github.com/giantswarm/giantswarm/issues/2390.


service/resource/cloudformationv2/validate.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+	"github.com/giantswarm/microerror"
+)
+
+type validator func(v1alpha1.AWSConfig) error
+
+func (r *Resource) validateCluster(cluster v1alpha1.AWSConfig) error {
+	validators := []validator{
+		r.validateHostPeeringRoutes,
+	}
```
rossf7: Nice, I like that we can extend the validator easily.

service/resource/cloudformationv2/validate.go
```diff
@@ -0,0 +1,50 @@
...
+	return nil
+}
+
+func (r *Resource) validateHostPeeringRoutes(cluster v1alpha1.AWSConfig) error {
+	input := &ec2.DescribeRouteTablesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name: aws.String("route.destination-cidr-block"),
+				Values: []*string{
+					aws.String(cluster.Spec.AWS.VPC.PrivateSubnetCIDR),
```
rossf7: Nitpick but we could use a key helper as we use the PrivateSubnetCIDR a few times.

service/resource/cloudformationv2/validate.go
```diff
@@ -0,0 +1,50 @@
...
+				},
+			},
+			&ec2.Filter{
+				Name: aws.String("vpc-id"),
+				Values: []*string{
+					aws.String(cluster.Spec.AWS.VPC.PeerID),
```
rossf7: Same here.

service/resource/cloudformationv2/validate.go
fgimenez: Good point, done, I've updated all the references to the use the new key helpers.

PR #677, opened 2018-01-26
Remove last cloud formation switch logic from legacy resource

rossf7:


service/keyv2/key.go
```diff
@@ -10,11 +10,6 @@ import (
 )
 
 const (
-	// CloudFormationVersion is the version in the version bundle for
-	// transitioning to Cloud Formation.
-	// TODO Remove once the migration is complete.
-	CloudFormationVersion = "0.2.0"
-
```
rossf7: 🎉

PR #682, opened 2018-01-29
Always use keyv2.ImageID when getting the CoreOS AMI

rossf7:

Fixes a bug with updates because we were still getting the CoreOS AMI from the custom object when getting the desired state.

### DIFF
--- a/service/keyv2/key.go
+++ b/service/keyv2/key.go
@@ -272,3 +272,30 @@ func componentName(domainName string) (string, error) {
 
 	return splits[0], nil
 }
+
+// ImageID returns the EC2 AMI for the configured region.
+func ImageID(customObject v1alpha1.AWSConfig) (string, error) {
+	region := Region(customObject)
+
+	/*
+		Container Linux AMIs for each active AWS region.
+		From: https://coreos.com/os/docs/latest/booting-on-ec2.html
+
+		NOTE 1: AMIs should always be for HVM virtualisation and not PV.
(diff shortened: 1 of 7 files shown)


service/keyv2/key.go
```diff
@@ -272,3 +272,30 @@ func componentName(domainName string) (string, error) {
...
+		"us-west-2":    "ami-dc4ce6a4",
+	}
+
+	imageID, ok := imageIDs[region]
+	if !ok {
+		return "", microerror.Maskf(invalidConfigError, "no image id for region '%s'", region)
+	}
+
+	return imageID, nil
+}
```
rossf7: Helper moves from `adapter` package to `keyv2`.

service/resource/cloudformationv2/main_stack.go
```diff
@@ -18,31 +18,36 @@ import (
 
 func newMainStack(customObject v1alpha1.AWSConfig) (StackState, error) {
 	stackName := keyv2.MainGuestStackName(customObject)
-	workers := len(customObject.Spec.AWS.Workers)
-	var workerImageID, workerInstanceType string
-	// FIXME: the imageID should not depend on the number of workers.
+	workerCount := keyv2.WorkerCount(customObject)
```
rossf7: As well as using the ImageID I also did some boyscouting to use keyv2 helpers in more places.

rossf7:
CI now passes after moving the tests to `2.0.0`. Thanks, increasing test coverage on this would be good.

Going to merge now and rebase the PR that deploys `2.0.0`.

PR #685, opened 2018-01-29
Use 2.0.0 version bundle in integration tests

rossf7:


rossf7:
I want to deprecate `0.2.0` so it no longer appears in `gsctl list releases`. I should have left a comment about that.

fgimenez:
np at all :) just wanted to make sure that the change didn't get in the PR without noticing.

rossf7:
CI is green! 😄

PR #688, opened 2018-01-31
add missing imageid unit tests

fgimenez:
This uncovers the bug we hit during CF migration validation where the operator tried to update when anything had changed in the state.


fgimenez:
Sure, good point, done.

PR #692, opened 2018-01-31
awsconfig/v4: add OIDC integration

corest:
Towards: https://github.com/giantswarm/giantswarm/issues/2413


corest:
No apiserver args implemented yet

Sample data: https://github.com/giantswarm/installations/blob/oidc/gauss/draughtsman-configmap-values.yaml#L42

corest:
@rossf7 but what next? Will we add new and new implementations for every specific auth provider? And then copy-paste that into all other operators?
Or this is gonna be just temporary before cluster-operator will be ready and we will move this stuff there?

rossf7:
OIDC is already pretty generic. So I don't think we'll need to support a lot of providers.

You're right we need to do this in kvm-operator too and azure-operator at some point. But I don't know enough about cluster-operator and it isn't ready. So I can't see a better solution at the moment.

flag/service/installation/guest/kubernetes/api/auth/provider/oidc/oidc.go
```diff
@@ -0,0 +1,8 @@
+package oidc
+
+type OIDC struct {
+	ClientID      string
+	IssuerUrl     string
```
rossf7: Nitpick: URL should be in caps.

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

service/framework.go
```diff
@@ -252,6 +253,18 @@ func NewVersionedResources(config Config, k8sClient kubernetes.Interface, awsCon
 
 		ccServiceConfig.Logger = config.Logger
 
+		// setup OIDC
```
corest: This looks super ugly, but if it will work with empty flags - I'll update cloudconfig then

service/awsconfig/v4/cloudconfig/cloud_config.go
```diff
@@ -14,22 +14,25 @@ const (
...
+		K8sAPIExtraArgs: []string{},
 	}
 }
 
 // CloudConfig implements the cloud config service interface.
 type CloudConfig struct {
 	// Dependencies.
-	logger micrologger.Logger
+	logger          micrologger.Logger
+	k8sapiextraargs []string
```
rossf7: Casing is weird here. Why is it all lowercase?

service/framework.go
rossf7: Do we want to add empty flags? Why can't the slice be empty if OIDC is not enabled?

service/awsconfig/v4/cloudconfig/cloud_config.go
```diff
@@ -42,10 +44,19 @@ func New(config Config) (*CloudConfig, error) {
 		return nil, microerror.Maskf(invalidConfigError, "logger must not be empty")
 	}
 
+	var k8sAPIExtraArgs []string
+	{
+		for arg := range config.K8sAPIExtraArgs {
+			if len(strings.SplitN(config.K8sAPIExtraArgs[arg], "=", 2)) > 0 {
```
corest: In this case, next parameters will pass validation:
  - `--param1` - without value
  - `--param2=val2` - with simple value
  - `-param3=arg=value` - with nested parameters

And empty flags (e.g. `--param2=`) will be skipped

service/framework.go
```diff
@@ -462,6 +475,13 @@ func NewVersionedResources(config Config, k8sClient kubernetes.Interface, awsCon
 		}
 	}
 
+	// As 2.1.0 based on the same subset of resources as 1.0.1 - it just copied
+	// into new minor version.
+	var resourcesV2_1_0 []framework.Resource
```
corest: I have doubts about this, as we adding here new parameters, but in fact, do not change any resource in the bundle. Is it worth to make this cascade copying of all resources because of those flags?

rossf7: Yes, this is going to need changes. We need a new s3object resource that uses cloudconfigv4.

See the structure here. https://github.com/giantswarm/aws-operator/tree/master/service/awsconfig/v3

CC @kopiczko as you're also working on this.

service/version_bundles.go
```diff
@@ -313,8 +313,8 @@ func NewVersionBundles() []versionbundle.Bundle {
 		{
 			Changelogs: []versionbundle.Changelog{
 				{
-					Component:   "cloudformation",
-					Description: "Add missing region to host account credentials.",
+					Component:   "cloudconfig",
+					Description: "Add OIDC integration for k8s api-server.",
```
rossf7: Nitpick: `k8s` should be `Kubernetes`. 

The changelogs are exposed to customers via the UI.

helm/aws-operator-chart/templates/03-configmap.yaml
```diff
@@ -13,5 +13,23 @@ data:
         region: '{{ .Values.Installation.V1.Provider.AWS.Region }}'
       installation:
         name: '{{ .Values.Installation.V1.Name }}'
+        {{- if .Values.Installation.V1.Guest }}
+        guest:
+          {{- if .Values.Installation.V1.Guest.Kubernetes }}
+          kubernetes:
+            {{- if .Values.Installation.V1.Guest.Kubernetes.API }}
+            api:
```
kopiczko: To me it looks like we can skip `api:` and `provider:` parts of this already long path. Are there other parts of kubernetes that could require kubernetes?

helm/aws-operator-chart/templates/03-configmap.yaml
```diff
@@ -13,5 +13,23 @@ data:
         region: '{{ .Values.Installation.V1.Provider.AWS.Region }}'
       installation:
         name: '{{ .Values.Installation.V1.Name }}'
+        {{- if .Values.Installation.V1.Guest }}
+        guest:
+          {{- if .Values.Installation.V1.Guest.Kubernetes }}
```
kopiczko: Can we skip the ifs, and handle empty values in the code instead? We have default values in installations anyway which can be simply empty strings.

service/awsconfig/v4/cloudconfig/cloud_config.go
```diff
@@ -39,9 +44,21 @@ func New(config Config) (*CloudConfig, error) {
...
+	var k8sAPIExtraArgs []string
+	{
+		for _, arg := range config.K8sAPIExtraArgs {
+			if !strings.HasSuffix(arg, "=") {
+				k8sAPIExtraArgs = append(k8sAPIExtraArgs, arg)
+			} else {
+				config.Logger.Log("warning", "api-server arguments must not be empty", "argument", arg)
+			}
+		}
+	}
```
kopiczko: I wouldn't do that and add check below instead...

service/service.go
```diff
@@ -142,8 +143,14 @@ func New(config Config) (*Service, error) {
 				Region:          config.Viper.GetString(config.Flag.Service.AWS.Region),
 			},
 			InstallationName: config.Viper.GetString(config.Flag.Service.Installation.Name),
-			Name:             config.Name,
-			PubKeyFile:       config.Viper.GetString(config.Flag.Service.AWS.PubKeyFile),
+			K8sAPIExtraArgs: []string{
```
kopiczko: ... I mean here `c := awsconfig.FrameworkConfig{` <- above this line you can construct extra args only if the flag values aren't empty.

kopiczko:
Also there isn't changelog. I don't know if you don't have to import different CC version.

helm/aws-operator-chart/templates/03-configmap.yaml
corest: There was two possible ways:
  1. Add default values for all envs in installation - in this case, we will not need those ifs
  2. Do not add default values and check nested keys in configmap template. In this case, those checks should be done on every layer (best practices from here https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/values.md#flat-or-nested-values)

I took the first approach. But I'm also okay to go with second

service/awsconfig/v4/cloudconfig/cloud_config.go
corest: I did it in this way, because I wanted to bound change amap to cloudconfig resource, as this change is related only to parameters, which should eventually impact k8scloudconfig only. Another option is to move it to the `NewFramework` function, where we validate most config parameters.

helm/aws-operator-chart/templates/03-configmap.yaml
corest: Configmap just reflects structure in draugthsman configmap

kopiczko: So all 3 `if` above are not necessary.

helm/aws-operator-chart/templates/03-configmap.yaml
kopiczko: I'd rather add that to default values. Then it's easier to see what settings we have globally.

PR #693, opened 2018-01-31
Add version 2.0.3 with k8scloudconfig 3.1.0

corest:
Will be rebased/merged on top of cloudconfigv5 PR

Set's version 2.0.3 as WIP so that it can be used to test real changes in k8scloudconfig 3.1.0

### DIFF
--- a/vendor/github.com/giantswarm/k8scloudconfig/v_3_0_0/master_template.go
+++ b/vendor/github.com/giantswarm/k8scloudconfig/v_3_0_0/master_template.go
@@ -1768,6 +1768,207 @@ write_files:
         omitStages:
           - "RequestReceived"
 
+- path: /etc/kubernetes/manifests/k8s-api-server.yml
+  owner: root
+  permissions: 0644
+  content: |
+    apiVersion: v1
+    kind: Pod
+    metadata:
+      name: k8s-api-server
+      namespace: kube-system
+    spec:
+      hostNetwork: true
+      containers:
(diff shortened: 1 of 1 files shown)


service/version_bundles.go
```diff
@@ -309,5 +309,46 @@ func newVersionBundles() []versionbundle.Bundle {
...
+				{
+					Name:    "nginx-ingress-controller",
+					Version: "0.10.2",
+				},
+			},
+			Dependencies: []versionbundle.Dependency{},
+			Deprecated:   false,
+			Name:         "aws-operator",
+			Time:         time.Date(2018, time.January, 31, 23, 40, 0, 0, time.UTC),
+			Version:      "2.0.3",
```
rossf7: Version should be `2.1.0` IMO.

service/framework_test.go
```diff
@@ -85,6 +85,15 @@ func Test_Service_NewVersionedResources(t *testing.T) {
 					"servicev2",
 					"endpointsv2",
 				},
+				"2.0.3": []string{
+					"kmskeyv2",
+					"s3bucketv2",
+					"s3objectv3",
+					"cloudformationv2",
```
rossf7: We also need a `cloudformationv3` resource that uses `k8scloudconfigv5`. This is so the code is isolated and there are separate code paths for `2.0.2` and `2.1.0`.

corest:
https://github.com/giantswarm/aws-operator/pull/692

PR #694, opened 2018-01-31
copy cloudconfigv4 to cloudconfigv5

corest:
cloudconfigv5 is k8scloudconfig v_3_1_0

New version for operator here: https://github.com/giantswarm/aws-operator/pull/693


kopiczko:
I think we should close that and wait till refactoring is finished. I have a script ready for making most of the tedious work.

PR #699, opened 2018-02-01
Add check for duplicate master nodes to alerter service

rossf7:
Towards giantswarm/giantswarm#2325 and giantswarm/giantswarm#2392

The 2 master nodes are due to a bug in the legacy resource. The legacy code is EOL but the bug causes disruption and fixing is a pain. It extends the existing alerter service to add a metric for clusters with duplicate resources.


service/alerter/aws.go
```diff
@@ -1,10 +1,32 @@
 package alerter
 
 import (
+	"fmt"
+
 	awsresources "github.com/giantswarm/aws-operator/resources/aws"
 	"github.com/giantswarm/microerror"
 )
 
+// ListMasterInstances lists the master instances for this cluster.
```
rossf7: This uses the legacy `resources/aws` package. But the bug only affects legacy so we can delete this check when we retire the code.

service/alerter/metric.go
```diff
@@ -3,6 +3,15 @@ package alerter
...
 var (
+	duplicateResourcesTotal = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "awsoperator",
+			Subsystem: "resources",
+			Name:      "duplicate_resources_total",
+			Help:      "Number of clusters with duplicate resources to be cleaned up.",
+		},
+		[]string{"resource"},
+	)
```
rossf7: Joe, this is the new metric. Does it look OK? Also are the more steps I need to get it scraped by Prometheus?

service/alerter/service.go
```diff
@@ -105,29 +106,65 @@ func (s *Service) StartAlerts() {
 		for {
 			select {
 			case <-alertChan:
-				err := s.OrphanResourcesAlert()
+				err := s.RunAllChecks()
```
rossf7: This extends the code so we can run 2 checks.

service/alerter/metric.go
JosephSalisbury: You need to register the metric with prometheus, see https://github.com/giantswarm/aws-operator/blob/f683ebbc4fd7370640242f47a9e9d3c71b7467b7/service/alerter/metric.go#L27

JosephSalisbury: I don't see the register?

rossf7: Thanks, its there now.

service/alerter/service.go
```diff
@@ -167,8 +204,20 @@ func FindOrphanResources(clusterIDs []string, resourceNames []string) []string {
 	return orphanResources
 }
 
-// UpdateMetrics updates the metric and logs the results.
-func (s Service) UpdateMetrics(resourceType string, resourceNames []string) {
+// UpdateDuplicateResourceMetrics updates the metric and logs the results.
```
fgimenez: Very nice that this can be reused <3

service/alerter/aws.go
```diff
@@ -1,10 +1,32 @@
...
+func (s *Service) ListMasterInstances(clusterID string) ([]string, error) {
+	instanceNames := []string{}
+
+	instances, err := awsresources.FindInstances(awsresources.FindInstancesInput{
+		Clients: s.awsClients,
+		Logger:  s.logger,
+		Pattern: fmt.Sprintf("%s-master-0", clusterID),
```
fgimenez: Out of curiosity where does this `-0` come from? I've seen some instances with it, but AFAIK we are always using the value returned by the key method. If it is assigned by AWS when the master is duplicated, could we have `-1`, `-2`... too?

rossf7: The `-0` is because originally we launched the workers as EC2 instances not using as ASG. So the code supports numbering instances.

Masters will always be `-0` but I removed the suffix because the code does a wildcard search for instances tagged `[cluster]-master*`. I think this is a better check.

fgimenez: :100:

PR #702, opened 2018-02-02
dep: vendor operatorkit

kopiczko:


kopiczko:
`dep` is so nice. 6 files updated.

kopiczko:
OK, won't work that way. I need to bring it with other changes.

PR #705, opened 2018-02-02
Update architecture section of README

rossf7:
The current architecture section is so out of date its useless IMO. I've updated it with the new operatorkit / CloudFormation architecture. WDYT?


README.md
```diff
@@ -3,12 +3,7 @@
 
 # aws-operator
 
-The aws-operator handles Kubernetes clusters running on a Kubernetes cluster
-inside of AWS.
-
-
-## Prerequisites
-
+The aws-operator manages Kubernetes clusters running on AWS.
```
rossf7: This summary makes more sense to me.

README.md
```diff
@@ -29,62 +24,60 @@ Build the standard way.
...
-folders for each cluster.
+### CloudFormation
 
-Each cluster contains the cloudconfig for master and worker instances.
+The guest Kubernetes clusters are provisioned using [AWS CloudFormation](4). The
+resources are split between 3 CloudFormation stacks.
 
-For example, assuming that we have the following customers with clusters:
+* guest-main manages the guest cluster resources.
+* host-setup manages an IAM role used for VPC peering.
```
fgimenez: Maybe good to mention here or above that some of the resources are created in the host account, for `guest-main` it's clear that it is created in the guest, but maybe not that the others are created in the host.

rossf7: Good point, I've added a note on this.

PR #712, opened 2018-02-05
VERSION_BUNDLE_VERSION=1.0.0

kopiczko:


kopiczko:
e2e doesn't work with legacy.

PR #721, opened 2018-02-06
awsconfig/v3: update imports

kopiczko:


kopiczko:
I would opt to merge those kinds of PR w/o approval as well. I'll provide a script to make that thing. WDYT?

service/awsconfig/v3/resource/cloudformation/adapter/instance.go
```diff
@@ -5,7 +5,7 @@ import (
 
 	"github.com/giantswarm/microerror"
 
-	"github.com/giantswarm/aws-operator/service/awsconfig/v2/key"
+	"github.com/giantswarm/aws-operator/service/awsconfig/v3/key"
```
rossf7: Comment below can also be removed.

PR #716, opened 2018-02-06
awsconfig/v4: copy v3

kopiczko:
Pure copy PR.

PR #722, opened 2018-02-06
awsconfig/v4: revert copy v3

kopiczko:
This reverts commit f6d306a568029037a8d53f0471c7bf8d27cf0590.


kopiczko:
I clicked to fast on Merge in https://github.com/giantswarm/aws-operator/pull/716

Sorry, so many things to wrap your head around.

kopiczko:
> The complication is the cloudformation resource also needs the k8scc version because the large clodconfig S3 URLs are versioned.

I don't think that is a problem when all packages are copied. Is it?

rossf7:
@kopiczko Yes this is fine when the extra packages are copied.

PR #726, opened 2018-02-06
service: add 2.1.0 WIP version bundle handled by awsconfig/v4

kopiczko:


service/awsconfig/v4/error.go
```diff
@@ -1,4 +1,4 @@
-package v3
+package v4
```
kopiczko: I fixed the script. I'll share in https://github.com/giantswarm/giantswarm/issues/2335

PR #727, opened 2018-02-06
Update state check to prevent 2nd master being created

rossf7:
Fixes giantswarm/giantswarm#2392

The `findExisting` check only considers instances with status `pending` or `running`. If the node is rebooting when the operator reprocesses the cluster a 2nd master node is launched.

The check now passes with any status except `terminated`. Bug only affects legacy clusters and legacy cluster creation is deprecated.

### DIFF
--- a/resources/aws/instance.go
+++ b/resources/aws/instance.go
@@ -55,6 +55,16 @@ func statePendingOrRunning(instance *ec2.Instance) bool {
 	return false
 }
 
+func stateTerminated(instance *ec2.Instance) bool {
+	stateCode := *instance.State.Code
+	switch stateCode {
+	case int64(EC2TerminatedState):
+		return true
+	}
+
+	return false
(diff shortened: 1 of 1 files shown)


resources/aws/instance.go
```diff
@@ -55,6 +55,16 @@ func statePendingOrRunning(instance *ec2.Instance) bool {
 	return false
 }
 
+func stateTerminated(instance *ec2.Instance) bool {
+	stateCode := *instance.State.Code
+	switch stateCode {
```
fgimenez: Maybe we could introduce the switch when we have more codes to check, wouldn't work now `return *instance.State.Code == int64(EC2TerminatedState)` ?

rossf7: Yes, this was copy / paste from the existing check. This is much nicer, done.

PR #729, opened 2018-02-07
e2e: set version bundle version from defined bundles

fgimenez:
@kopiczko this is the version bundle version retrieval part, the rest in #714 is about parallel execution of WIP and current version bundle versions.


integration/framework.go
```diff
@@ -457,3 +458,21 @@ func (f *framework) ApplyAWSConfigPatch(patch []PatchSpec, clusterName string) e
...
+
+func (f *framework) GetVersionBundleVersion(bundle []versionbundle.Bundle, vType string) string {
+	var output string
+	log.Printf("Tested version %q", vType)
+
+	for _, v := range bundle {
+		if (vType == "current" || vType == "") && !v.Deprecated && !v.WIP {
+			output = v.Version
+			break
+		}
```
kopiczko: This is possible that we have multiple non-deprecated non-WIP versions. This isn't deterministic. I think you can use data to take latest.

integration/framework.go
```diff
@@ -457,3 +458,21 @@ func (f *framework) ApplyAWSConfigPatch(patch []PatchSpec, clusterName string) e
...
+		if vType == "wip" && v.WIP {
```
kopiczko: Similar situation to the above.

integration/framework.go
```diff
@@ -457,3 +458,21 @@ func (f *framework) ApplyAWSConfigPatch(patch []PatchSpec, clusterName string) e
 
 	return nil
 }
+
+func (f *framework) GetVersionBundleVersion(bundle []versionbundle.Bundle, vType string) string {
```
kopiczko: This doesn't have to be a method of framework, but this is a minor thing.

But, wouldn't it be nice to validate vType? Or if you don't want to return error make it a typed constant and handle it in TestMain?

fgimenez: Sure, made function.

Ok, I've added validation and now the function returns an error too.

integration/framework.go
fgimenez: Added code to return the latest non-deprecated non-wip for current.

integration/framework.go
fgimenez: Same here, latest wip is returned.

fgimenez:
@kopiczko comments addressed, PTAL

integration/framework.go
```diff
@@ -457,3 +459,36 @@ func (f *framework) ApplyAWSConfigPatch(patch []PatchSpec, clusterName string) e
...
+		return "", fmt.Errorf("%q is not a valid version bundle version type", vType)
+	}
+
+	var output string
+	log.Printf("Tested version %q", vType)
+
+	// sort bundle by time to get the newest vbv.
+	s := versionbundle.SortBundlesByTime(bundle)
+	sort.Sort(s)
+	for i := len(s) - 1; i >= 0; i-- {
```
kopiczko: I think using reverse is a bit cleaner https://golang.org/pkg/sort/#Reverse

fgimenez: Yep much better :) done

PR #732, opened 2018-02-08
awsconfigv4: switch k8scloudconfig to v3.1.0

corest:


corest:
Tested on gauss. Looks good

PR #740, opened 2018-02-13
add orphan cluster alert

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2390

Adds an alert for orphan clusters, custom objects that have been created long time ago and still don't have resources associated. This aims to generalize the validation failure detection, instead of checking for specific validation errors or enabling alert triggering from the reconciliation loop, we look for the effect of a validation error, a custom object without resources.


service/alerter/service.go
```diff
@@ -204,6 +233,33 @@ func FindOrphanResources(clusterIDs []string, resourceNames []string) []string {
...
+func FindOrphanClusters(awsConfigs []v1alpha1.AWSConfig, resourceNames []string) []string {
+	resources := map[string]bool{}
+	orphanClusters := []string{}
+
+	for _, resourceName := range resourceNames {
+		resources[resourceName] = true
+	}
+
+	n := time.Now()
+	l := n.Add(-10 * time.Minute)
```
rossf7: I like the delay here to prevent false positives / alert spam. What about extracting the time interval to a constant? If we add more alerts it would be good to use the same delay.

service/alerter/service.go
```diff
@@ -227,3 +283,16 @@ func (s Service) UpdateOrphanResourceMetrics(resourceType string, resourceNames
...
 	}
 }
+
+// UpdateOrphanClusterMetrics updates the metric and logs the results.
+func (s Service) UpdateOrphanClusterMetrics(resourceType string, clusterIDs []string) {
+	resourceCount := len(clusterIDs)
+
+	orphanClustersTotal.WithLabelValues(resourceType).Set(float64(resourceCount))
+
+	s.logger.Log("info", fmt.Sprintf("alerter service found %d clusters with missing resources", resourceCount))
```
rossf7: Nitpick but I think the log message should say missing AWS resources. I think thats clearer and matches the metrics help text.

service/alerter/service.go
fgimenez: Good point, done.

service/alerter/service.go
fgimenez: Sure, done. I've also taken the opportunity to add an additional test case.

service/alerter/service.go
```diff
@@ -204,6 +234,33 @@ func FindOrphanResources(clusterIDs []string, resourceNames []string) []string {
...
+
+	for _, resourceName := range resourceNames {
+		resources[resourceName] = true
+	}
+
+	n := time.Now()
+	l := n.Add(-1 * maxMinutesSinceClusterCreation * time.Minute)
+	for _, awsConfig := range awsConfigs {
+		// if the cluster has been recently created maybe the resources are not still there.
+		if awsConfig.CreationTimestamp.After(l) {
```
JosephSalisbury: As a random point, it feels like prometheus should take care of this part. If the alerting rule has a long period, for example.

fgimenez: I think a longer period wouldn't help here. Imagine that, by the time the rule kicks in, a cluster has been just created, if we don't check the creation timestamp we might think that the cluster is orphaned because its resources are not yet created, while this is totally normal (it takes some time to create the resources after the custom object is detected). 

IMO at some point we need to check the actual cluster creation time and only try to find resources of old enough clusters, this place seems to me like the more appropriate, or maybe I'm not getting right the longer period you suggest?

JosephSalisbury: So, my thinking was to expose the metrics directly to Prometheus, and then in the Prometheus alert define how long it should be till the alert is fired. Not certain it's a massive issue, but having the logic here feels slightly off?

JosephSalisbury: I reckon the alerts would be specific to the operators? Like, the aws alerts are all for the aws-operator anyway

If you wanna hang out and chat about aws-operator exposing more metrics, I would also love that too :D

fgimenez: Indeed, this alert is specific to this operator, maybe could be extended to other cloud operators, uncertain.

PR #742, opened 2018-02-14
Add method 'get' for service resource in RBAC

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2551

### DIFF
--- a/helm/aws-operator-chart/templates/01-rbac.yaml
+++ b/helm/aws-operator-chart/templates/01-rbac.yaml
@@ -39,9 +39,17 @@ rules:
       - ""
     resources:
       - endpoints
+    verbs:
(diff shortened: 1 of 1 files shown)


kopiczko:
@marians I believe `get` is the only one missing in the list.

pipo02mix:
If you follow the commits history, you will see how I started with 'get' and then I discovered now aws-operator needs all verbs :)

marians:
Yeah, I saw that. But that history is lost after a merge commit. And as we sometimes go back through the commit history to explain a change, a PR title that explains the actual change is helpful.

marians:
To illustrate what I mean, this is the history as it's looking now:

image

PR #745, opened 2018-02-15
awsconfig/v5: add loadbalancer resource structure

rossf7:

Adds 2.1.1 WIP version bundle and structure for new loadbalancer resource. Deletion logic will come in later PRs.

### DIFF
--- a/service/awsconfig/v5/resource/loadbalancer/delete.go
+++ b/service/awsconfig/v5/resource/loadbalancer/delete.go
@@ -1,0 +1,30 @@
+package loadbalancer
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/framework"
+)
+
+// TODO
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
+	return nil
+}
+
+func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*framework.Patch, error) {
+	delete, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	patch := framework.NewPatch()
+	patch.SetDeleteChange(delete)
+
+	return patch, nil
+}
+
+// TODO
+func (r *Resource) newDeleteChange(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error) {
+	return nil, nil
+}
--- a/service/awsconfig/v5/resource/loadbalancer/resource.go
+++ b/service/awsconfig/v5/resource/loadbalancer/resource.go
@@ -1,6 +1,8 @@
 package loadbalancer
 
 import (
+	"reflect"
+
 	"github.com/giantswarm/microerror"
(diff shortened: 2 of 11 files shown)


service/awsconfig/v5/resource/loadbalancer/delete.go
```diff
@@ -0,0 +1,30 @@
+package loadbalancer
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/framework"
+)
+
+// TODO
```
rossf7: Only delete needs to be implemented.

service/awsconfig/v5/resource_set.go
```diff
@@ -248,6 +265,7 @@ func NewResourceSet(config ResourceSetConfig) (*framework.ResourceSet, error) {
 		kmsKeyResource,
 		s3BucketResource,
 		s3BucketObjectResource,
+		loadBalancerResource,
```
rossf7: Resource runs before the cloudformation resource to do the GC. Otherwise the ELBs will be in the public subnet and the CloudFormation stacks will not delete cleanly.

service/awsconfig/v5/resource/loadbalancer/resource.go
```diff
@@ -0,0 +1,71 @@
...
+// DefaultConfig provides a default configuration to create a new loadbalancer
+// resource by best effort.
+func DefaultConfig() Config {
+	return Config{
+		// Dependencies.
+		AwsService: nil,
+		Clients:    Clients{},
+		Logger:     nil,
+	}
+}
```
kopiczko: We decided to drop DefaultConfig.

service/awsconfig/v5/resource/loadbalancer/resource.go
```diff
@@ -0,0 +1,71 @@
...
+
+// New creates a new configured loadbalancer resource.
+func New(config Config) (*Resource, error) {
+	// Dependencies.
+	if config.AwsService == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.AwsService must not be empty")
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty")
+	}
```
kopiczko: Aren't clients validated?

service/awsconfig/v5/resource/loadbalancer/resource.go
rossf7: Ack. I keep forgetting that. Will fix.

service/awsconfig/v5/resource/loadbalancer/resource.go
rossf7: No but they probably should be.

service/awsconfig/v5/resource/loadbalancer/spec.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/aws/aws-sdk-go/service/elb"
+)
+
+type Clients struct {
+	ELB ELBClient
+}
+
+// ELBClient describes the methods required to be implemented by an ELB AWS client.
+type ELBClient interface {
```
fgimenez: <3

PR #746, opened 2018-02-15
Add back the AWS tag KubernetesCluster with the clusterID

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2508


service/awsconfig/v5/key/key_test.go
```diff
@@ -123,6 +123,7 @@ func Test_ClusterTags(t *testing.T) {
 	expectedID := "test-cluster"
 	expectedTags := map[string]string{
 		"kubernetes.io/cluster/test-cluster": "owned",
+		"KubernetesCluster":                  "test-cluster",
```
rossf7: `KubernetesCluster` is deprecated but adding it back should be OK. But what about naming the tag `giantswarm.io/cluster`.

That still meets the reporting requirement but it's clear we define the cluster ID. WDYT?

service/awsconfig/v5/resource/cloudformation/resource_test.go
```diff
@@ -29,16 +29,33 @@ func Test_Resource_Cloudformation_GetCloudFormationTags(t *testing.T) {
...
+	}
+
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
 			tags := getCloudFormationTags(tc.obj)
 
-			if !reflect.DeepEqual(tc.expectedTags, tags) {
-				t.Fatalf("Expected cloud formation tags %v but was %v", tc.expectedTags, tags)
+			for _, tag := range tc.expectedTags {
+				if noContainsTag(tag, tags) {
```
rossf7: For me naming the function containsTag and using `!containsTag` would be clearer.

I'd also prefer the function to be defined separately rather than in a closure.

service/version_bundles.go
```diff
@@ -373,6 +373,11 @@ func NewVersionBundles() []versionbundle.Bundle {
 					Description: "Delete AWS Cloud Provider resources when deleting clusters.",
 					Kind:        versionbundle.KindAdded,
 				},
+				{
+					Component:   "aws-operator",
+					Description: "Get back Kubernetes Cluster ID tag helping cost analysis in AWS.",
```
rossf7: To me `Add KubernetesCluster AWS resource tag that was removed in 2.0.0.`  makes more sense.

I think the changelog should describe what is changing. The tag helps cost analysis but can be used for other things.

service/awsconfig/v5/resource/cloudformation/resource_test.go
fgimenez: +1 for `containsTag` as suggested by Ross. I still prefer the closure, the function is only needed in this test and we don't need to bloat the top-level namespace, if it is required later somewhere else we can always refactor IMO.

rossf7: That makes sense. I'm fine with keeping the closure :)

service/awsconfig/v5/key/key_test.go
pipo02mix: I dont really mind, I used the same one because then clients have the Cost Analysis history aligned (same key/value)

service/awsconfig/v5/resource/cloudformation/resource_test.go
pipo02mix: ok it was my first approach indeed hehe

service/version_bundles.go
pipo02mix: ok, but then I leave the old tag name?

rossf7: Lets use the new cluster name and the changelog can be something like

`Add AWS resource tag with GiantSwarm Cluster ID for reporting.`

rossf7:
<3

PR #759, opened 2018-02-20
Fix default pub key path for legacy clusters

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2581

This only affects very old clusters (legacy without cluster version https://github.com/giantswarm/aws-operator/blob/master/service/awsconfig/v1/resource/legacy/resource.go#L252) which need an SSH KeyPair to be created. The right path of the public key is `/.ssh/id_rsa.pub`, this is where the secret with the file content is mounted https://github.com/giantswarm/aws-operator/blob/master/helm/aws-operator-chart/templates/08-deployment.yaml#L53.


fgimenez:
If the operator needs to reconcile the KeyPair in (very) old clusters it is likely going to fail, if not all should be good (or as good as it was).

PR #766, opened 2018-02-21
Vendor kube-proxy fix

rossf7:
Fixes giantswarm/giantswarm#2598

Vendors k8scloudconfig v_2_0_2 with the fix to start kube-proxy if it is not already running.

### DIFF
--- a/service/awsconfig/v5/cloudconfig/master_template.go
+++ b/service/awsconfig/v5/cloudconfig/master_template.go
@@ -31,7 +31,7 @@ func (c *CloudConfig) NewMasterTemplate(customObject v1alpha1.AWSConfig, certs l
 			customObject: customObject,
 			keys:         keys,
 		}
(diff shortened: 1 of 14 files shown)


service/awsconfig/v5/cloudconfig/master_template.go
```diff
@@ -31,7 +31,7 @@ func (c *CloudConfig) NewMasterTemplate(customObject v1alpha1.AWSConfig, certs l
 			customObject: customObject,
 			keys:         keys,
 		}
-		params.Hyperkube.Apiserver.Docker.CommandExtraArgs = c.k8sAPIExtraArgs
+		params.Hyperkube.Apiserver.Pod.CommandExtraArgs = c.k8sAPIExtraArgs
```
rossf7: @corest After vendoring k8scloudconfig I needed to update to fix a compile error. Is the new path correct?

rossf7:
Tested with a locally launched cluster.

PR #767, opened 2018-02-21
Update k8scloudconfig with the changes needed

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2550


pipo02mix:
Tested in gauss. PTAL

rossf7:
Looks like the component versions and changelog for the version bundle still need to be updated.

service/version_bundles.go
```diff
@@ -383,6 +383,11 @@ func NewVersionBundles() []versionbundle.Bundle {
 					Description: "Change default etcd data dir to /var/lib/etcd.",
 					Kind:        versionbundle.KindChanged,
 				},
+				{
+					Component:   "aws-operator",
+					Description: "Allow persistent volumes be automatically extended when the claims are changed.",
+					Kind:        versionbundle.KindAdded,
+				},
```
pipo02mix: @rossf7 it is not here?

pipo02mix:
> Looks like the component versions and changelog for the version bundle still need to be updated.
Which component version?
it is k8scloudconfig update after adding a tag to the api server manifest

Gopkg.lock
```diff
@@ -136,12 +136,12 @@
   branch = "master"
   name = "github.com/giantswarm/k8scloudconfig"
   packages = [
-    "v_2_0_0",
```
fgimenez: I'm curious, where does this change come from?

rossf7:
@pipo02mix I checked the k8scloudconfig changelog in vendor. It says calico and etcd were updated but the component versions in the bundle are the old ones.

> Updated calico to 3.0.2.
> Updated etcd to 3.3.1.

Gopkg.lock
rossf7: Yesterday I changed legacy `awsconfig/v1` from `v_2_0_0` to `v_2_0_2` for the kube-proxy fix. Not sure why this didn't change then but I think this is OK.

See https://github.com/giantswarm/aws-operator/pull/766/files

PR #772, opened 2018-02-22
Update k8scc kube limit resources

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2523


pipo02mix:
I have tested with my latest changes but now my branch is behind. Updating and PTAL

rossf7:
@pipo02mix I've already vendored your change and it will go live with v2.1.1 in #779. Sorry I should have said.

pipo02mix:
ok then closing this PR

PR #802, opened 2018-03-02
Update k8scc with aggregation layer changes

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2656


pipo02mix:
Tested PTAL

service/awsconfig/v7/version_bundle.go
```diff
@@ -9,6 +9,11 @@ import (
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
+			{
+				Component:   "aws-operator",
```
rossf7: Nitpick but I think component should be `cloudconfig`.

PR #804, opened 2018-03-02
awsconfig/v7: switch to cloudconfig 3.2.1

corest:


corest:
@rossf7 I remember about changelog, but this PR does not bring any changes as cloudconfig 3.2.1 has any at this time. It is just like those PR to make further changes more clear

PR #815, opened 2018-03-05
Persist operator binary to e2e jobs

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2712 and https://github.com/giantswarm/giantswarm/issues/2678

With these changes the binary built on the first step will be available on the e2e test steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,9 @@ e2eTest: &e2eTest
     steps:
     - checkout
 
+    - attach_workspace:
+        at: .
+
     - run: |
         wget -q $(curl -sS https://api.github.com/repos/giantswarm/e2e-harness/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
         chmod +x ./e2e-harness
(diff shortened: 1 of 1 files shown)


fgimenez:
> This file is really confusing in general

We are using node anchors in `.circleci/config.yaml` so that we don't repeat blocks for each of the e2e test jobs. The anchor is defined here https://github.com/giantswarm/aws-operator/blob/master/.circleci/config.yml#L1-L28 and is reused with an alias in each of the jobs https://github.com/giantswarm/aws-operator/blob/master/.circleci/config.yml#L53, https://github.com/giantswarm/aws-operator/blob/master/.circleci/config.yml#L58, https://github.com/giantswarm/aws-operator/blob/master/.circleci/config.yml#L63 and https://github.com/giantswarm/aws-operator/blob/master/.circleci/config.yml#L68. This is the part of YAML spec that explains node anchors http://yaml.org/spec/1.2/spec.html#id2785586 and this, alias nodes http://yaml.org/spec/1.2/spec.html#id2786196, hope it helps :)

PR #817, opened 2018-03-06
Extract split ELB logic and add tests

rossf7:
Fixes giantswarm/giantswarm#2667

Cleanup from last weeks problem with cluster creation on `viking`. Extracts split logic and adds tests to close the PM.


rossf7:
Merged, e2e tests were green before rebasing.

PR #827, opened 2018-03-09
Change installation name in integration tests for reporting

rossf7:
Towards giantswarm/giantswarm#2775

Makes installation name clearer for cost reporting on the AWS resources we use for CI.

See https://github.com/giantswarm/giantswarm/issues/2775#issuecomment-371525532

### DIFF
--- a/integration/operator_test.go
+++ b/integration/operator_test.go
@@ -36,7 +36,7 @@ const (
                 GroupsClaim: ""
       Update:
         Enabled: ${GUEST_UPDATE_ENABLED}
-    Name: ci-awsop
+    Name: ci-aws-operator
     Provider:
       AWS:
         Region: ${AWS_REGION}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ e2eTest: &e2eTest
 
     - run:
         name: set CLUSTER_NAME env var
-        command: echo 'export CLUSTER_NAME=ci-awsop-${TESTED_VERSION}-${CIRCLE_SHA1:0:5}' >> $BASH_ENV
+        command: echo 'export CLUSTER_NAME=ci-aws-operator-${TESTED_VERSION}-${CIRCLE_SHA1:0:5}' >> $BASH_ENV
 
     - run: ./e2e-harness test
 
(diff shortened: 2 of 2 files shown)


integration/operator_test.go
```diff
@@ -37,7 +37,7 @@ const (
                 GroupsClaim: ""
       Update:
         Enabled: ${GUEST_UPDATE_ENABLED}
-    Name: ci-awsop
+    Name: ci-aws-operator
```
xh3b4sd: There is also https://github.com/giantswarm/aws-operator/blob/4d6eb492e6506841cdf9377dac17604ef24449be/.circleci/config.yml#L24. Why are there two different things and what is each used for?

xh3b4sd: BTW I like the explicit naming way more. I am just not sure if this has implications on some AWS magic, like CF stack name length restrictions or something stupid.

rossf7: I like the more explicit name too. CF stack name length is maximum 128 chars so we're OK.

I've changed and let's see if the e2e tests are OK.

xh3b4sd: Why do we have this in two places though? @fgimenez can you tell?

rossf7: I think its 2 different things. One is the installation name and the other if the cluster ID prefix.

I can extract to a constant because the value is the same? But IMO they aren't the same thing.

rossf7: The cluster ID is too long for the ELBs. So leaving it as is.

screen shot 2018-03-09 at 12 37 04 pm

xh3b4sd: Is there a way to just hash the ELB names or cap them somehow reliably?

xh3b4sd: It should be possible for us to use longer cluster IDs.

rossf7: > This name must be unique within your set of load balancers for the region, must have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, and cannot begin or end with a hyphen.

We use the format [cluster ID]-api and [cluster ID]-ingress. This makes the ELBs easy to find in the AWS console. Hashing would be more complex and make it harder to find them.

I think this is fine as it is.

xh3b4sd: Ok.

PR #834, opened 2018-03-12
only do full teardown when not in CI

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2710 and https://github.com/giantswarm/giantswarm/issues/2678

### DIFF
--- a/integration/operator_test.go
+++ b/integration/operator_test.go
@@ -168,8 +168,11 @@ func TestMain(m *testing.M) {
 
 	if os.Getenv("KEEP_RESOURCES") != "true" {
 		f.DeleteGuestCluster()
-		operatorTearDown()
-		f.TearDown()
+		// only do full teardown when not on CI
+		if os.Getenv("CIRCLECI") != "true" {
+			operatorTearDown()
+			f.TearDown()
+		}
 		deleteHostPeerVPC()
 	}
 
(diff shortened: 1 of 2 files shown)


integration/operator_test.go
```diff
@@ -168,8 +168,11 @@ func TestMain(m *testing.M) {
 
 	if os.Getenv("KEEP_RESOURCES") != "true" {
 		f.DeleteGuestCluster()
-		operatorTearDown()
-		f.TearDown()
+		// only do full teardown when not on CI
+		if os.Getenv("CIRCLECI") != "true" {
+			operatorTearDown()
+			f.TearDown()
+		}
```
kopiczko: What do those functions do?

fgimenez: They remove components installed for the build https://github.com/giantswarm/aws-operator/blob/master/integration/operator_test.go#L311-L316 and https://github.com/giantswarm/e2e-harness/blob/master/pkg/framework/framework.go#L211-L214 from minikube/localkube.

kopiczko: So this basically means we don't clean the minikube ran inside circleci?

fgimenez: Exactly, we don't care about the state of the circleCI machine once the build is done, it can be even useful to keep these components for debugging.

PR #849, opened 2018-03-15
Use latest Healthz structure and remove dead code

rossf7:
Uses the microendpoint healthz endpoint, removes the unused middleware and updates the healthcheck to use the latest approach.


service/healthz/service.go
```diff
@@ -62,26 +59,53 @@ func New(config Config) (*Service, error) {
...
+
+			// Get the current user.
+			_, err := clients.IAM.GetUser(&iam.GetUserInput{})
+			if err != nil {
+				ch <- err.Error()
+				return
+			}
+
+			ch <- ""
+		}()
```
rossf7: Here is the new check.

PR #858, opened 2018-03-20
Added more information to orphan metrics

fgimenez:
Adds more information to orphan resource metrics, so that we can show the ids of the affected resources in the alerts.


fgimenez:
Tested on gauss

PR #863, opened 2018-03-21
awsconfig/v9: switch to cloudconfig 3.2.3

corest:


service/awsconfig/v9/version_bundle.go
```diff
@@ -10,8 +10,13 @@ func VersionBundle() versionbundle.Bundle {
...
 		Changelogs: []versionbundle.Changelog{
 			{
-				Component:   "component",
-				Description: "Add your description here",
+				Component:   "cloudconfig",
+				Description: "Kubernetes updated to version 1.9.5.",
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "cloudconfig",
```
rossf7: Nitpick but `Nginx Ingress Controller` looks better in the release output IMO.

PR #884, opened 2018-04-03
Create a bucket for S3 access logs

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2726


service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -45,6 +45,24 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 			return microerror.Mask(err)
 		}
 
+		err = r.createAccessLogBucket(ctx, bucketInput)
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
rossf7: Creating the extra bucket here is OK. But I think it would be better to extend the resource so it can create multiple S3 buckets.

This is what we do for the s3object resource.

https://github.com/giantswarm/aws-operator/blob/087f506cfc669ca05ae25e84c295e01c8870e6a0/service/awsconfig/v10/resource/s3object/spec.go#L16-L20 

The desired spec for the bucket can include the logging settings.

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -68,3 +86,41 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
...
+	if IsBucketAlreadyExists(err) || IsBucketAlreadyOwnedByYou(err) {
+		// Fall through.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	_, err = r.clients.S3.PutBucketAcl(&s3.PutBucketAclInput{
+		Bucket:       aws.String(bucketInput.Name + "-logs"),
```
rossf7: We should use a key function for the bucket name and a constant for the S3 URIs.

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -68,3 +86,41 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
...
+		GrantReadACP: aws.String("uri=http://acs.amazonaws.com/groups/s3/LogDelivery"),
```
fgimenez: This string could be defined as a constant.

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -45,6 +45,24 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
+			Bucket: aws.String(bucketInput.Name),
+			BucketLoggingStatus: &s3.BucketLoggingStatus{
+				LoggingEnabled: &s3.LoggingEnabled{
+					TargetBucket: aws.String(bucketInput.Name + "-logs"),
+					TargetPrefix: aws.String(key.ClusterID(customObject) + "-access-logs/"),
+				},
+			},
+		})
+		if err != nil {
+			return err
```
fgimenez: Error masking is missing here (there are some more occurrences of this).

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -68,3 +86,41 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
...
+		return err
+	}
+
+	//Enable logs for the target bucket too
+	_, err = r.clients.S3.PutBucketLogging(&s3.PutBucketLoggingInput{
+		Bucket: aws.String(bucketInput.Name + "-logs"),
+		BucketLoggingStatus: &s3.BucketLoggingStatus{
+			LoggingEnabled: &s3.LoggingEnabled{
+				TargetBucket: aws.String(bucketInput.Name + "-logs"),
+				TargetPrefix: aws.String("self-access-logs/"),
```
fgimenez: Would be great to have a constant here too.

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -68,3 +86,41 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
 
 	return BucketState{}, nil
 }
+
+func (r *Resource) createAccessLogBucket(ctx context.Context, bucketInput BucketState) error {
```
fgimenez: Why do you need to pass the context here? It is not used in the method body.

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -45,6 +45,24 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		_, err = r.clients.S3.PutBucketLogging(&s3.PutBucketLoggingInput{
+			Bucket: aws.String(bucketInput.Name),
+			BucketLoggingStatus: &s3.BucketLoggingStatus{
+				LoggingEnabled: &s3.LoggingEnabled{
+					TargetBucket: aws.String(bucketInput.Name + "-logs"),
+					TargetPrefix: aws.String(key.ClusterID(customObject) + "-access-logs/"),
```
fgimenez: A key function would be helpful here too.

service/awsconfig/v10/resource/s3bucket/create.go
pipo02mix: Yeah, I made on purpose because I masked in the caller function. Is it a good practice to do it always?

fgimenez: Yep, current convention is masking everywhere.

service/awsconfig/v10/resource/s3bucket/create.go
pipo02mix: Ok so it does mean creating a new resource folder, add create/delete/update an instance and pass it to the framework?

rossf7: A new resource isn't needed. It means extending the current resource so it can create multiple S3 buckets. This is how the s3object resource already works.

So instead of having.

```
type BucketState struct {
	Name string
}
```

You have.

```
map[string]BucketState{}
```

The BucketState can be extended to have the extra logging config.

pipo02mix: ahh ok get it now! ;)

service/awsconfig/v10/key/key.go
```diff
@@ -290,6 +292,10 @@ func SubnetName(customObject v1alpha1.AWSConfig, suffix string) string {
 	return fmt.Sprintf("%s-%s", ClusterID(customObject), suffix)
 }
 
+func TargetLogBucketName(customObject v1alpha1.AWSConfig) string {
+	return fmt.Sprintf("%s-access-logs", ClusterID(customObject))
+}
```
pipo02mix: Align with host clusters format

service/awsconfig/v10/resource/s3bucket/current.go
```diff
@@ -15,31 +16,40 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+		}
+		if err != nil {
+			return []BucketState{}, microerror.Mask(err)
+		}
+		if err == nil {
+			bucketState := BucketState{
+				Name: bucketName,
+			}
+			bucketsState = append(bucketsState, bucketState)
+		}
```
pipo02mix: I am not sure if there is a prettier way to resolve it, suggestions?

service/awsconfig/v10/resource/s3bucket/current.go
```diff
@@ -15,31 +16,40 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
-		Name: bucketName,
+	bucketsState := []BucketState{}
+
+	for _, bucketName := range bucketNames {
+		headInput := &s3.HeadBucketInput{
+			Bucket: aws.String(bucketName),
+		}
+		_, err = r.clients.S3.HeadBucket(headInput)
+		if IsBucketNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the S3 bucket '%s'", bucketName))
```
fgimenez: Instead of `'%s'` you can use `%q`.

service/awsconfig/v10/resource/s3bucket/resource.go
```diff
@@ -81,17 +81,27 @@ func (r *Resource) Name() string {
...
 	if !ok {
-		return BucketState{}, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", BucketState{}, v)
+		return []BucketState{}, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", []BucketState{}, v)
 	}
 
-	return bucketState, nil
+	return bucketsState, nil
+}
+
+func containsBucketState(bucketState BucketState, bucketStateList []BucketState) bool {
```
fgimenez: Why not passing here just the `bucketState` name as a string? Also, this could be unit tested.

service/awsconfig/v10/resource/s3bucket/desired_test.go
```diff
@@ -64,13 +67,16 @@ func Test_Resource_S3Bucket_GetDesiredState(t *testing.T) {
...
-				t.Fatalf("case expected '%T', got '%T'", desiredBucket, result)
+				t.Fatalf("case expected '%T', got '%T'", desiredBuckets, result)
 			}
 
-			if tc.expectedName != desiredBucket.Name {
-				t.Fatalf("expected bucket name '%s' got '%s'", tc.expectedName, desiredBucket.Name)
+			// Order should be respected in the slice returned (always delivery log bucket first)
+			for key, desiredBucket := range desiredBuckets {
+				if tc.expectedNames[key] != desiredBucket.Name {
+					t.Fatalf("expected bucket name '%s' got '%s'", tc.expectedNames[key], desiredBucket.Name)
```
fgimenez: Same about `%q`.

service/awsconfig/v10/key/key.go
```diff
@@ -290,6 +292,10 @@ func SubnetName(customObject v1alpha1.AWSConfig, suffix string) string {
 	return fmt.Sprintf("%s-%s", ClusterID(customObject), suffix)
 }
 
+func TargetLogBucketName(customObject v1alpha1.AWSConfig) string {
```
fgimenez: We should have unit tests for this.

pipo02mix:
Comments addressed and tested PTAL

service/awsconfig/v10/resource/s3bucket/current.go
```diff
@@ -15,31 +17,45 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+		BucketState{
+			Name:           key.TargetLogBucketName(customObject),
+			IsDeliveryLog:  true,
+			LoggingEnabled: true,
+		},
+		BucketState{
+			Name:           key.BucketName(customObject, accountID),
+			IsDeliveryLog:  false,
+			LoggingEnabled: true,
+		},
```
rossf7: This looks strange to me because the desired state is repeated in the current state. It's a pain that you can't list S3 buckets with a tag filter via the API.

I think it would be better to just have a slice of bucket names here and compare them against the desired state. WDYT?

service/awsconfig/v10/resource/s3bucket/spec.go
```diff
@@ -6,7 +6,9 @@ import (
 
 // BucketState is the state representation on which the resource methods work.
 type BucketState struct {
-	Name string
+	Name           string
+	IsDeliveryLog  bool
```
rossf7: Naming ¯\_(ツ)_/¯ but `IsLoggingBucket` or `IsLogDestination` would make more sense to me.

xh3b4sd:
To be on the same page and keep the knowledge sharing up, please request reviews from @giantswarm/sig-operator whenever working on operator related tasks. 

Specifically in here please make sure that the update from the currently active version bundle to the WIP bundle works properly. For now we still need to verify this manually. I am working on automated integration tests for this.

service/awsconfig/v10/resource/s3bucket/current.go
kopiczko: I think this is wrong in general. CurrentState shouldn't assume some values are set, those settings should be retrieved. I mean `IsDeliveryLog` and `LoggingEnabled`

service/awsconfig/v10/resource/s3bucket/current.go
```diff
@@ -15,31 +17,45 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+		_, err = r.clients.S3.HeadBucket(headInput)
+		if IsBucketNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the S3 bucket %q", inputBucket.Name))
+		}
+		if err != nil {
+			return []BucketState{}, nil
+		}
+		if err == nil {
+			currentBucketState = append(currentBucketState, inputBucket)
+		}
```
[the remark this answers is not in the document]
pipo02mix: Well if the error is not found we leave it empty because it means the current state has no buckets, right?

In case a different error is produced then I agree

service/awsconfig/v10/resource/s3bucket/error.go
```diff
@@ -26,8 +28,12 @@ func IsBucketNotFound(err error) bool {
 	if !ok {
 		return false
 	}
-	// TODO Find constant in the Go SDK for the error code.
-	if aerr.Code() == "NotFound" {
+	if aerr.Code() == s3.ErrCodeNoSuchBucket {
```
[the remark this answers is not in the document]
pipo02mix: sorry I dont get you. What do you mean?

service/awsconfig/v10/resource/s3bucket/create.go
```diff
@@ -16,55 +17,88 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
...
 	}
-	desiredBucket, err := toBucketState(desiredState)
+	desiredBuckets, err := toBucketState(desiredState)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	if currentBucket.Name == "" || desiredBucket.Name != currentBucket.Name {
-		return desiredBucket, nil
+	createState := []BucketState{}
```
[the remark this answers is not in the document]
pipo02mix: I have just followed the same approach I saw for others resource, and for example [here](https://github.com/giantswarm/aws-operator/blob/master/service/awsconfig/v9/resource/s3bucket/create.go#L69) we return the empty object instead of `nil`. Does it change when we manage slices? why? and if so how it would look like?

service/awsconfig/v10/resource/s3bucket/error.go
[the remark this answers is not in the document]
xh3b4sd: `aerr` is the result of a type assertion above. Base for the assertion is `err`. Base should be `microerror.Cause(err)` because only the cause is the actual error we want to check against. This here can result to nasty side effects.

service/awsconfig/v10/resource/s3bucket/create.go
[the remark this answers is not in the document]
xh3b4sd: This is because of the difference between values and pointers. The point is the zero value. The zero value of a value is the initialized empty value. The zero value of a pointer is nil. A slice is a pointer.

service/awsconfig/v10/resource/s3bucket/spec.go
kopiczko: Also makes more sense to me but #naming as you said :)

service/awsconfig/v10/resource/s3bucket/error.go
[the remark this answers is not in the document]
pipo02mix: hmm but then it should be like
```
var notSuchBucket = microerror.New(s3.ErrCodeNoSuchBucket)

// IsBucketNotFound asserts bucket not found error from upstream's API code.
func IsBucketNotFound(err error) bool {
	return microerror.Cause(err) == notSuchBucket
}
```

service/awsconfig/v10/resource/s3bucket/create.go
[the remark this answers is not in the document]
pipo02mix: ok nice!

But is this will work too?
`var bs []BucketState`
If you don't set any value is `nil` by default, right? or is it better be more explicit?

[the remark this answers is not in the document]
kopiczko: `var bs []BucketState` yes, that works, and it sets the value to `nil`. In examples above I tried to be as clear (explicit) as possible.

service/awsconfig/v10/resource/s3bucket/error.go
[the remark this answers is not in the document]
xh3b4sd: IMO this will not work because the S3 error you want to check against is also wrapped. It is ok to have our native error and other stuff being checked in the matcher functions.

service/awsconfig/v10/resource/s3bucket/error.go
```diff
@@ -36,6 +36,9 @@ func IsBucketNotFound(err error) bool {
 	if strings.Contains(aerr.Code(), "Not Found") {
 		return true
 	}
+	if microerror.Cause(err) == notFoundError {
```
xh3b4sd: The cause should also be used to assert the AWS error in the very first line. Also, the not found error has nothing to do here. There is already an assertion for the not found error. The matcher here is for buckets not being found and thus there should be a specific error for this which we assert here. We always have an error and a matcher for it. Lets please not mix things up and abuse in different places.

service/awsconfig/v10/resource/s3bucket/current.go
```diff
@@ -15,31 +17,51 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+
+		headInput := &s3.HeadBucketInput{
+			Bucket: aws.String(inputBucketName),
+		}
+		_, err = r.clients.S3.HeadBucket(headInput)
+		if IsBucketNotFound(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find the S3 bucket %q", inputBucketName))
+		} else if err != nil {
+			return nil, err
```
xh3b4sd: We should always mask errors. Please.

pipo02mix: I missed this one. Thanks for catch

service/awsconfig/v10/resource/s3bucket/error.go
```diff
@@ -22,21 +20,15 @@ func IsNotFound(err error) bool {
 	return microerror.Cause(err) == notFoundError
 }
 
+var noSuchBucketError = microerror.New("Not Found")
```
xh3b4sd: You may want to read how we deal with errors. Please see https://github.com/giantswarm/fmt/blob/master/go/general_style_guide.md#errors. The matcher is `IsBucketNotFound`. The error should then be `bucketNotFoundError`. The message should then be `bucket not found`. Also, below you removed the string matcher checks. I did not request this. I am not into details here but this should may be reverted, or is there a specific reason for this?

service/awsconfig/v10/resource/s3bucket/error.go
pipo02mix: The problem here is the expression `microerror.Cause(aerr)` returns `"NotFound: Not Found\n\tstatus code: 404, request id: 00B7939985F4AE19, host id: Vh7D62xqbTSorifIzGY+2xbdnpPE2ahRH2CFIHoI+YnMzlubk7BGiolsfswZTj2USweivAUqDuQ="%`  I think the AWS API does not return the correct response here. It is the reason I used `Contains` method. Any idea how to solve it in a prettier way?

service/awsconfig/v10/resource/s3bucket/delete.go
```diff
@@ -69,19 +71,23 @@ func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desire
...
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	var bucketToDelete BucketState
-	if currentBucket.Name != "" {
-		bucketToDelete = desiredBucket
+	var bucketsToDelete []BucketState
+	for _, bucket := range currentBuckets {
+		// Destination Logs Bucket should no be delete because it has to keep logs
```
rossf7: "should not be deleted" and comment should end with a `.`.

service/awsconfig/v10/resource/s3bucket/desired.go
```diff
@@ -18,9 +19,20 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	bucketState := BucketState{
-		Name: key.BucketName(customObject, accountID),
+	// First bucket must to be the delivery log bucket because otherwise
```
rossf7: must be the delivery log

pipo02mix:
tested and it works fine. I have done some refactoring in the current state to improve readability (hopefully). PTAL

PR #885, opened 2018-04-03
Simplify namespace resource

rossf7:
As well as the `ebsvolume` and `loadbalancer` AWS resources the `namespace` and `service` K8s resources can also be simplified.

Continuing to get this done. It also reduces a bit the amount of code we need to copy for each resource set.


service/awsconfig/v10/resource/namespace/create.go
```diff
@@ -4,48 +4,33 @@ import (
...
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/giantswarm/aws-operator/service/awsconfig/v10/key"
 )
 
-func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
-	namespaceToCreate, err := toNamespace(createChange)
+// EnsureCreated creates a Kubernetes namespace in the host cluster for the
+// guest cluster.
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
```
rossf7: This is the create logic.

service/awsconfig/v10/resource/namespace/create.go
```diff
@@ -4,48 +4,33 @@ import (
...
-		r.logger.LogCtx(ctx, "level", "debug", "message", "creating Kubernetes namespace: created")
-	} else {
+	_, err = r.k8sClient.CoreV1().Namespaces().Create(namespaceToCreate)
+	if apierrors.IsAlreadyExists(err) {
 		r.logger.LogCtx(ctx, "level", "debug", "message", "creating Kubernetes namespace: already created")
-	}
-
-	return nil
-}
+		return nil
```
rossf7: We keep the logging if the namespace already exists.

service/awsconfig/v10/resource/namespace/delete.go
```diff
@@ -4,66 +4,34 @@ import (
...
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/aws-operator/service/awsconfig/v10/key"
 )
 
-func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
-	namespaceToDelete, err := toNamespace(deleteChange)
+// EnsureDeleted deletes the Kubernetes namespace in the host cluster for the
+// guest cluster.
```
rossf7: Delete logic.

service/awsconfig/v10/resource/namespace/k8s_test.go
```diff
@@ -0,0 +1,69 @@
...
+						Customer: v1alpha1.ClusterCustomer{
+							ID: "giantswarm",
+						},
+					},
+				},
+			},
+			expectedName: "al9qy",
+			expectedLabels: map[string]string{
+				"cluster":  "al9qy",
+				"customer": "giantswarm",
```
rossf7: Labels are out of date with `fmt`. I'll update as a separate change.

rossf7:
OK that makes sense. I got a bit carried with removing code. Since this is a k8s resource and we need to implement create and delete the benefit is less clear.

Closing. I'll update the k8s labels to get them in line with `fmt` as a separate change.

PR #887, opened 2018-04-05
Opens cadvisor port from host to guest

JosephSalisbury:
We currently go via the apiserver proxy to scrape cadvisor.
This will allow us to scrape cadvisor directly in future,
which is slightly more efficient.

### DIFF
--- a/service/awsconfig/v10/adapter/security_groups.go
+++ b/service/awsconfig/v10/adapter/security_groups.go
@@ -30,6 +30,7 @@ type securityGroupRule struct {
 
 const (
 	allPorts             = -1
+	cadvisorPort         = 4194
 	kubeletPort          = 10250
 	nodeExporterPort     = 10300
 	kubeStateMetricsPort = 10301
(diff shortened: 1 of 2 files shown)


rossf7:
I agree the comments are really useful and we could actually have them in the AWS console.

We just need to set the description field in the SecurityGroupIngress. I wanted to do this during the CF migration but I didn't get to it in the end.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html

JosephSalisbury:
@rossf7 || @xh3b4sd ptalaaaaaaaa

service/awsconfig/v10/version_bundle.go
```diff
@@ -34,6 +34,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Detach EBS volumes before deletion when deleting clusters.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "aws-operator",
+				Description: "Opened port 4194 for Cadvisor scraping from host cluster.",
```
rossf7: Engaging nitpick mode but this should be CAdvisor or cAdvisor which is what they use.

https://github.com/google/cadvisor

JosephSalisbury: omg i've been spelling this wrong forever

JosephSalisbury: Fixed

rossf7: Thanks! <3 I saw it and then I couldn't unsee it. You'll probably have the same problem now! 😆

JosephSalisbury: You have cursed me :(

PR #888, opened 2018-04-05
Bump k8scc to 3.2.5

r7vme:


service/awsconfig/v10/version_bundle.go
```diff
@@ -9,6 +9,26 @@ import (
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
+			{
```
r7vme: Only changelog changed

PR #910, opened 2018-04-25
Update containerlinux to 1688.5.3

rossf7:
Towards  giantswarm/giantswarm/issues/3021

### DIFF
--- a/service/controller/v10/version_bundle.go
+++ b/service/controller/v10/version_bundle.go
@@ -71,7 +71,7 @@ func VersionBundle() versionbundle.Bundle {
 			},
 			{
 				Name:    "docker",
-				Version: "17.09.0",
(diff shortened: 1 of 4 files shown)


service/controller/v10/version_bundle.go
```diff
@@ -62,7 +67,7 @@ func VersionBundle() versionbundle.Bundle {
 			},
 			{
 				Name:    "containerlinux",
-				Version: "1632.3.0",
+				Version: "1688.5.3",
 			},
 			{
 				Name:    "docker",
```
kopiczko: Docker version is 17.12.1 - that is the most important bit :)

rossf7: Ay it is indeed! Thanks good catch. Done.

PR #914, opened 2018-04-27
Add an option to mark installation as testing 

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2996


pipo02mix:
Tested. PTAL

main.go
```diff
@@ -121,6 +121,8 @@ func mainError() error {
 
 	daemonCommand.PersistentFlags().Bool(f.Service.Guest.Update.Enabled, false, "Whether updates of guest cluster nodes are allowed to be processed upon reconciliation.")
 
+	daemonCommand.PersistentFlags().Bool(f.Service.Installation.Testing, false, "Is the installation not used for production.")
```
fgimenez: I think it would be more clear `Is the installation used for testing`, WDYT?

service/controller/v10/resource/s3bucket/resource.go
```diff
@@ -127,3 +130,13 @@ func (r *Resource) getS3BucketTags(customObject v1alpha1.AWSConfig) []*s3.Tag {
 
 	return s3Tags
 }
+
+func (r *Resource) canBeDeleted(bucket BucketState) bool {
+	if bucket.IsLoggingBucket {
```
fgimenez: This function could be just `return bucket.IsLoggingBucket && !r.TestingEnvironment`.

vendor/github.com/giantswarm/e2etemplates/pkg/e2etemplates/aws_operator_chart_values.go
```diff
@@ -17,6 +17,8 @@ const AWSOperatorChartValues = `Installation:
       Update:
         Enabled: ${GUEST_UPDATE_ENABLED}
     Name: ci-aws-operator
+    Infra:
```
fgimenez: These changes should be proposed first to https://github.com/giantswarm/e2etemplates and then run here `dep ensure -update github.com/giantswarm/e2etemplates`, you shouldn't change `vendor` files directly.

pipo02mix:
Ok @xh3b4sd so If I have understood you correctly you are in favor of using an explicit label on the configuration. I think it is a good idea, I was doubting because it could make the configuration much verbose but it makes it clear and also give us the possibility to adapt every cluster differently. I will make the changes to use a new property (`DeleteLoggingBucket`)

service/controller/v10/resource/s3bucket/resource.go
rossf7: I think this is a nice simplification. Would be good to implement if its straight forward.

service/controller/v10/resource/s3bucket/resource_test.go
```diff
@@ -2,6 +2,9 @@ package s3bucket
 
 import (
 	"testing"
+
+	awsservice "github.com/giantswarm/aws-operator/service/aws"
+	"github.com/giantswarm/micrologger/microloggertest"
```
rossf7: Nitpick but awsservice should be in a separate block.

service/controller/v10/resource/s3bucket/resource.go
pipo02mix: indeed v10 should be untouched because is frozen. In the v11 is done as suggested. Let me remove changes from v10 :)

service/controller/v10/resource/s3bucket/resource_test.go
pipo02mix: but is an external project, why should be in different block?

rossf7: The aws service is part of aws-operator so it should be in a separate block.

service/controller/v10/resource/s3bucket/resource.go
rossf7: Good catch. Thanks for removing.

service/controller/v10/resource/s3bucket/resource_test.go
pipo02mix: Jesus Christ I was only reading the micrologger one 
Thanks for pointing this out

rossf7: Jeje np :)

PR #918, opened 2018-04-30
controller/v9_patch1: copy v9

r7vme:


kopiczko:
If you still have energy please let's make it `v9patch1`. It's better in therms on Go package name, and we can still split by `patch`.

If no enough energy LGTM.

PR #939, opened 2018-05-14
Whitelist host cluster NAT gateway EIPs

rossf7:

Extends the K8s API server whitelisting to include the elastic IPs of the host cluster NAT gateways. This is so cluster-operator can connect to AWS guest clusters when the whitelisting is enabled.

### DIFF
--- a/service/controller/v11/adapter/mocks.go
+++ b/service/controller/v11/adapter/mocks.go
@@ -23,6 +23,25 @@ type EC2ClientMock struct {
 	vpcCIDR             string
 	unexistingVPC       bool
 	peeringID           string
+	elasticIPs          []string
+}
+
+func (e *EC2ClientMock) DescribeAddresses(input *ec2.DescribeAddressesInput) (*ec2.DescribeAddressesOutput, error) {
+	addresses := make([]*ec2.Address, 0)
+
+	for _, eip := range e.elasticIPs {
+		address := &ec2.Address{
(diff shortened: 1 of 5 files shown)


service/controller/v11/adapter/mocks.go
```diff
@@ -23,6 +23,25 @@ type EC2ClientMock struct {
 	vpcCIDR             string
 	unexistingVPC       bool
 	peeringID           string
+	elasticIPs          []string
+}
+
+func (e *EC2ClientMock) DescribeAddresses(input *ec2.DescribeAddressesInput) (*ec2.DescribeAddressesOutput, error) {
```
fgimenez: <3

service/controller/v11/adapter/security_groups.go
```diff
@@ -173,41 +186,91 @@ func (s *securityGroupsAdapter) getIngressRules(customObject v1alpha1.AWSConfig)
...
+			{
+				Name: aws.String("tag:giantswarm.io/installation"),
+				Values: []*string{
+					aws.String(cfg.InstallationName),
+				},
+			},
+		},
+	}
+	output, err := cfg.HostClients.EC2.DescribeAddresses(describeAddressesInput)
+	if err != nil {
```
fgimenez: Nice how this error is checked all the way up :+1:

rossf7:
Tested on gauss. Going with this.

PR #948, opened 2018-05-21
Region dependent ARN

fgimenez:


service/controller/v11/key/key.go
```diff
@@ -295,6 +295,16 @@ func Region(customObject v1alpha1.AWSConfig) string {
 	return customObject.Spec.AWS.Region
 }
 
+func RegionARN(customObject v1alpha1.AWSConfig) string {
+	regionARN := "aws"
+
+	if customObject.Spec.AWS.Region == "cn-north-1" {
```
MarcelMue: Is this the only cn region? Will we have to do changes if Adidas wants to use another cn region?

service/controller/v11/templates/cloudformation/guest/iam_policies.go
```diff
@@ -37,11 +37,11 @@ const IAMPolicies = `{{define "iam_policies"}}
 
           - Effect: "Allow"
             Action: "s3:ListBucket"
-            Resource: "arn:aws:s3:::{{.S3Bucket}}"
+            Resource: "arn:{{.RegionARN}}:s3:::{{.S3Bucket}}"
```
MarcelMue: Do we need a migration path for this? I feel like an update from an old operator to this will fail, but I don't have much insight into aws-operator.

service/controller/v11/key/key.go
fgimenez: We are only using this one now, there's also `cn-northwest-1` but it is not being used. In general the approach for region-dependent data is to only include the regions being used, see for instance how we determine the AMIs to be used by region https://github.com/giantswarm/aws-operator/blob/master/service/controller/v11/key/key.go#L391-L397

service/controller/v11/templates/cloudformation/guest/iam_policies.go
fgimenez: Nope, as you can see in `key.RegionARN` in this same changeset `{{.RegionARN}}` will default to the previous value, `aws`.

service/controller/v11/key/key.go
rossf7: Would be good to use `key.Region` here. As we will restructure the cluster CRs at some point.

fgimenez: Good point! Done.

PR #951, opened 2018-05-22
add-etcd-backup-support

calvix:
allow access 2379 port from host cluster in order to do etcd backups

towards https://github.com/giantswarm/giantswarm/issues/3136


calvix:
> Is there any security implications?

etcd will be available from host cluster cidr, it still requires cert auth in order to do anything with it
This is similar to what we don't with kubelets.
Nothing major from security side.

PR #955, opened 2018-05-24
Use k8scloudconfig v3.3.1

rossf7:

Uses new k8scloudconfig with k8s 1.10.3 and removes kube-state-metrics so it can be managed by chart-operator.


rossf7:
CI is green. Going with this.

PR #957, opened 2018-05-24
Use Container Linux 1745.4.0

rossf7:
Updates AMIs to use latest stable `1745.4.0` this also updates Docker to `18.03.1`.

This was due to failed cluster upgrades due to the Etcd volume not attaching to the new master EC2 instance with m5.large instances. With this upgrade the updates seem much more stable.

Should also help with giantswarm/giantswarm#3159.

EDIT: Latest is now `1745.4.0`.

### DIFF
--- a/service/controller/v11/key/key.go
+++ b/service/controller/v11/key/key.go
@@ -392,11 +392,11 @@ func ImageID(customObject v1alpha1.AWSConfig) (string, error) {
 		NOTE 1: AMIs should always be for HVM virtualisation and not PV.
 		NOTE 2: You also need to update the tests.
 
-		service/awsconfig/v10/key/key_test.go
-		service/awsconfig/v10/resource/cloudformation/adapter/adapter_test.go
-		service/resource/cloudformationv2/main_stack_test.go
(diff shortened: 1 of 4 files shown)


calvix:
there has already been new stable ` 1745.4.0` so maybe you want to use that

service/controller/v11/key/key.go
```diff
@@ -399,11 +399,11 @@ func ImageID(customObject v1alpha1.AWSConfig) (string, error) {
 		Current Release: CoreOS Container Linux stable 1576.5.0 (HVM)
```
fgimenez: It would be nice to update the version in the comment.

rossf7: Good catch. I'll fix.

rossf7:
@calvix Ack. I'll switch to `1745.4.0`.  From the bug report looks like it only affects VMware / bare metal but good to be on the latest.

service/controller/v11/key/key.go
rossf7: Done.

PR #965, opened 2018-05-29
Add v12 to version bundles and drainer framework

rossf7:
This step got missed when I wired up v12.


rossf7:
@fgimenez Sure, will do.

rossf7:
I'd missed v12 from the drainer framework. Retrying e2e.

PR #967, opened 2018-05-29
Increase Docker EBS volume from 50 to 100 GB

rossf7:
Towards giantswarm/giantswarm#3118

Increases the volume size to resolve the PM above. As in some installations we were filling the 50 GB volume.


rossf7:
CI green. Merging.

PR #968, opened 2018-05-29
Remove v10

rossf7:
v10 introduced the k8s upgrade to 1.10 but we didn't set a major version update as per our new release process.

@tuommaki Is confirming no clusters exist that use v10. This PR removes v10 and makes v11 `4.0.0`.


rossf7:
v10 is in use so can't be removed. Closing.

PR #978, opened 2018-05-31
WIP: disable KMS for testing

rossf7:
As KMS is not available in China this build disables KMS. So we can do more testing and find other issues.


rossf7:
No longer needed. We can now create clusters on axolotl using Vault instead of KMS.

PR #982, opened 2018-06-01
Increase ASG pause time from 5 to 15 mins

rossf7:
Towards giantswarm/giantswarm#3118

Before rolling out 3.4.4 -> v9patch2 we want to increase the rolling update pause time to 15 mins. Previous value of 5 mins caused some disruption.

Will update v12 if this increase resolves the problem.


rossf7:
Going with this as change is minor and patch releases aren't covered by e2e yet.

PR #983, opened 2018-06-01
Increase ASG rolling update pause time to 15 mins

rossf7:
Same change as #982 but for v12 (K8s 1.10).

I think the pause time will apply in the e2e tests. So we might need to configure a lower value there. @fgimenez WDYT?


rossf7:
CI green. Rolling.

PR #984, opened 2018-06-04
examples, integartion, vendor: use apiextentions charts

kopiczko:
Towards: https://github.com/giantswarm/giantswarm/issues/3314

Examples are removed because they are becoming more and more outdated.


examples/README.md
```diff
@@ -1,233 +0,0 @@
-# Running aws-operator Locally
```
kopiczko: I removed whole examples directory. There are outdated charts, and no one cares about them.

Gopkg.toml
```diff
@@ -41,10 +37,6 @@ required = [
   name = "github.com/cenkalti/backoff"
   version = "1.1.0"
 
-[[constraint]]
-  branch = "master"
-  name = "github.com/docker/distribution"
```
kopiczko: `docker/distribution` and `Azure/go-autorest` weren't imported or required. I don't know why we have them here.
W/o them `dep ensure` could complete in reasonable time for me.

fgimenez: Weird, dep would complain if they are not used. Anyway :+1:

kopiczko:
@fgimenez I think we only need to remove `COMMON_DOMAIN_GUEST`. Otherwise all is good. Am I missing something?

fgimenez:
I see there's a `COMMON_DOMAIN` defined, if it has been added recently (in other words, if it is not a leftover from when we didn't have separated guest and host account) and it doesn't include `k8s` I think we are good.

kopiczko:
@fgimenez what is `COMMON_DOMAIN_GUEST` set to now?

fgimenez:
If it hasn't changed `k8s.gauss.eu-central-1.aws.gigantic.io`

kopiczko:
BTW we can't we keep those env vars in `.circleci/config.yml`?

kopiczko:
I checked the circle docs and the highest level you can set env variables is `job`. With that we'd have to maintain the same var in multiple places which is dump.

kopiczko:
Updated the COMMON_DOMAIN in circle, CI is green. Merging.

PR #985, opened 2018-06-04
Vendor update e2e-harness and versionbundle

fgimenez:
Aims to fix how versions are determined in e2e tests after recent index release changes.


fgimenez:
Thanks! :)

PR #987, opened 2018-06-05
Pass docker command extra args for worker kubelet

rossf7:
Towards giantswarm/giantswarm#2894

For China use the kubelet extra args to use the retagged pause container image (quay.io/giantswarm/pause-amd64:3.1). This is done for the master but is also needed in the worker cloudconfig.


rossf7:
Out of date. Will fix in new PR with other China fixes.

PR #988, opened 2018-06-05
Use Kubernetes 1.10.2 for latest release

rossf7:
Towards giantswarm/giantswarm/issues/3332

Updates K8s to `1.10.2` in `v12` which will then be made the active release. `v13` is coming! 😱 ☘️


rossf7:
@corest Yes I also prefer rolling forwards :( A shame K8s 1.10.4 isn't ready for us.

PR #990, opened 2018-06-06
v12: use constant LB idle timeouts

kopiczko:

@rossf7 I'm fine if it doesn't land in v12. I prefer that, but if not possible I'll migrate to v13.

### DIFF
--- a/service/controller/v12/version_bundle.go
+++ b/service/controller/v12/version_bundle.go
@@ -14,6 +14,11 @@ func VersionBundle() versionbundle.Bundle {
 			},
 			{
 				Component:   "aws-operator",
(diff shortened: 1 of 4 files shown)


service/controller/v12/version_bundle.go
```diff
@@ -12,6 +12,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Increased Docker EBS volume size from 50 to 100 GB.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "aws-operator",
+				Description: "Use global LB idle timeout settings, instead of CR configured ones.",
```
rossf7: I think using ELB here makes more sense.

PR #999, opened 2018-06-07
Fix depends on

r7vme:


service/controller/v12/templates/cloudformation/guest/instance.go
```diff
@@ -4,6 +4,9 @@ const Instance = `{{define "instance"}}
   {{ .Instance.Master.Instance.ResourceName }}:
     Type: "AWS::EC2::Instance"
     Description: Master instance
+    DependsOn:
```
r7vme: Problem was that VM did not depend on volumes to it was immediately created with etcd as first disk and in that time docker disk was resized (takes 3-5 minutes) and then docker disk attached as second disk and mounted to `/var/lib/etcd`.

This dependency makes sure that VM will wait to both volumes to be ready before get started.

service/controller/v12/templates/cloudformation/guest/instance.go
```diff
@@ -19,25 +22,21 @@ const Instance = `{{define "instance"}}
         Value: {{ .Instance.Cluster.ID }}-master
   DockerVolume:
     Type: AWS::EC2::Volume
-    DependsOn:
-    - {{ .Instance.Master.Instance.ResourceName }}
     Properties:
       Encrypted: true
-      Size: 100
+      Size: 50
```
r7vme: I'm rolling master docker disk size to 50GB, because we don't really need 100GB on master node. 100GB only needed for workers.

This prevents unnecessary delay (3-5 minutes caused by resizing) while upgrade.

service/controller/v12/templates/cloudformation/guest/instance.go
rossf7: Very good catch, thanks!

service/controller/v12/templates/cloudformation/guest/instance.go
rossf7: Yes makes sense to leave the volume at 50GB for masters. Great that it avoids the delay.

service/controller/v13/templates/cloudformation/guest/instance.go
```diff
@@ -19,25 +22,21 @@ const Instance = `{{define "instance"}}
...
     Type: AWS::EC2::Volume
-    DependsOn:
-    - {{ .Instance.Master.Instance.ResourceName }}
     Properties:
       Encrypted: true
-      Size: 100
+      Size: 50
       VolumeType: gp2
-      AvailabilityZone: !GetAtt {{ .Instance.Master.Instance.ResourceName }}.AvailabilityZone
+      AvailabilityZone: {{ .Instance.Master.AZ }}
```
rossf7: This is because the master resource now depends on the volumes?

r7vme: Yes, otherwise it has curcular error :(

r7vme:
thanks for review unfortunately this is only a part of the problem. The thing is that with m5 we in general affected by "disk ordering problem". Disks randomly can be nvme1 / nvme2 , so we need to extract disk label (that we set in CF).

This is already hacked in community
https://github.com/kubernetes-incubator/kube-aws/issues/1313
https://github.com/coreos/bugs/issues/2399#issuecomment-379730681

We also need to do some kind of hack. I'll work on that tomorrow.

r7vme:
> Also I've tested K8s 1.10.4 and it fixes the configmap problem. So upgrading from 1.10.2 is an option. WDYT?

Yes 1.10.4 makes total sense.

r7vme:
- Added 1.10.4
- Added NVMe udev rules (allows to extract disk labels and use `/dev/xvdh` and `/dev/xvdc`)

Tested in ginger. Test list added in description.

r7vme:
I'll be merging this on Monday.

service/controller/v12/templates/cloudconfig/master_format_var_lib_docker_service.go
```diff
@@ -8,7 +8,7 @@ ConditionPathExists=!/var/lib/docker
 
 [Service]
 Type=oneshot
-ExecStart=/bin/bash -c "([ -b "/dev/xvdc" ] && /usr/sbin/mkfs.xfs -f /dev/xvdc -L docker) || ([ -b "/dev/nvme1n1" ] && /usr/sbin/mkfs.xfs -f /dev/nvme1n1 -L docker)"
+ExecStart=/bin/bash -c "[ -e "/dev/xvdc" ] && /usr/sbin/mkfs.xfs -f /dev/xvdc -L docker"
```
r7vme: `-e` file exists (any type), because `/dev/xvdc` is symlink in case NVMe and block device in case `m3`

service/controller/v12/templates/cloudconfig/nvme_udev_hack.go
```diff
@@ -0,0 +1,26 @@
+package cloudconfig
+
+const NVMEUdevRule = `KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", PROGRAM="/opt/ebs-nvme-mapping /dev/%k", SYMLINK+="%c"
```
r7vme: udev rule that calls script on any events with nvme that will create/delete sylinks with name that was specified in EBS e.g. `/dev/xvdh`

service/controller/v12/templates/cloudconfig/nvme_udev_hack.go
```diff
@@ -0,0 +1,26 @@
...
+
+const NVMEUdevScript = `#!/bin/bash
+vol=$(nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
+vol=${vol#/dev/}
+if [[ -n "$vol" ]]; then
+    echo ${vol/xvd/sd} ${vol/sd/xvd}
+fi
+`
+
+const NVMEUdevTriggerUnit = `[Unit]
```
r7vme: This unit only necessary on the first boot, because udev rule was just added and we need to retrigger udev.

oogali:
@r7vme @rossf7 I'm excited that you've discovered my solution[1] for dealing with NVMe devices in AWS by way of the CoreOS and kube-aws projects, but as a courtesy, can you include a copy of the license and copyright notice[2] in your derivative work?

1: https://github.com/oogali/ebs-automatic-nvme-mapping
2: https://github.com/oogali/ebs-automatic-nvme-mapping/blob/master/LICENSE

r7vme:
Hi @oogali , adding licence [here](https://github.com/giantswarm/aws-operator/pull/1006)

PR #1017, opened 2018-06-14
Make decryption depend on encrypter

fgimenez:
There's some repeated boilerplate here in the templates, also in the `templateData` method. Maybe we could refactor, suggestions welcome!


service/controller/v13/templates/cloudconfig/decrypt_keys_assets_script.go
```diff
@@ -1,7 +1,95 @@
...
+        --data '{"ciphertext": "$(cat $encKey)"}' \
+        {{ .VaultAddress }}/v1/transit/decrypt/{{ .EncryptionKey }} | \
+        jq -r .data.plaintext | base64 -d > $f
+      mv -f $f ${encKey%.enc}
+    done;
+    echo done.
+}
+
+aws_login () {
+    pkcs7=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/pkcs7 | tr -d '\n')
```
rossf7: This is the EC2 metadata endpoint? For me, a comment would help for readability.

service/controller/v13/templates/cloudconfig/decrypt_keys_assets_script.go
```diff
@@ -1,7 +1,95 @@
 package cloudconfig
 
 const DecryptKeysAssetsScript = `#!/bin/bash -e
+{{ if eq .EncrypterType "vault" }}
```
rossf7: There is a lot of bash magic here and I see what you mean about the boilerplate with similar logic in decrypt_tls_assets_script.go

@corest @calvix Could you help here as I think you'll have better input that me.

fgimenez: Thanks, I'd like to get also input from @giantswarm/sig-security here.

fgimenez:
> For azure-operator we had the idea of having a Go binary that would download secrets from Key Vault.
> 
> In the end we're moving away from Key Vault so we didn't implement. But the same approach could help make the code easier to maintain.
> 
> I'd suggest doing this as a follow up clean up task rather than now.

Yes, makes total sense. We have already implemented all the login, encryption functionality and common functions in the vault encrypter package https://github.com/giantswarm/aws-operator/tree/master/service/controller/v13/encrypter/vault, adding a decryption method is very straightforward. Maybe we could extract this package to its own repo.

service/controller/v13/cloudconfig/master_template.go
```diff
@@ -337,3 +342,25 @@ func (e *MasterExtension) VerbatimSections() []k8scloudconfig.VerbatimSection {
...
 }
+
+func (e *MasterExtension) templateData() TemplateData {
+	var encrypterType string
+	var vaultAddress string
+	var encryptionKey string
+	v, ok := e.encrypter.(*vault.Encrypter)
+	if ok {
+		encrypterType = "vault"
+		encryptionKey, _ = v.EncryptionKey(context.TODO(), e.customObject)
```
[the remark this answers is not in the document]
fgimenez: Good point, in this case the context is not used and it's there just to satisfy the interface, but looks weird and not very future-proof. I've modified the master and worker extensions to have a context field, now it is injected from the `CloudConfig` object, PTAL.

service/controller/v13/templates/cloudconfig/decrypt_keys_assets_script.go
fgimenez: Sure, done.

service/controller/v13/cloudconfig/master_template.go
```diff
@@ -337,3 +342,25 @@ func (e *MasterExtension) VerbatimSections() []k8scloudconfig.VerbatimSection {
...
 	return newSections
 }
+
+func (e *MasterExtension) templateData() TemplateData {
+	var encrypterType string
+	var vaultAddress string
+	var encryptionKey string
+	v, ok := e.encrypter.(*vault.Encrypter)
+	if ok {
+		encrypterType = "vault"
```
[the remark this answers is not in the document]
fgimenez: Good point, I've added constants for this in #1016, will update this once it lands.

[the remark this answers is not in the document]
rossf7: Nice, thanks!

service/controller/v13/cloudconfig/master_template.go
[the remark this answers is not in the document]
rossf7: Thanks makes total sense but good to pass the current context.

service/controller/v13/cloudconfig/master_template.go
```diff
@@ -68,7 +72,9 @@ type RandomKeyTmplSet struct {
 
 type MasterExtension struct {
 	certs            legacy.CompactTLSAssets
+	ctx              context.Context
```
kopiczko: I think passing the context to the struct is a bad thing. E.g. ancelling this context can have unexpected consequences. I think context is designed to be request scope, so it should be only passed as a first argument to the function call.

service/controller/v13/encrypter/vault/vault.go
```diff
@@ -62,6 +62,8 @@ func NewEncrypter(c *EncrypterConfig) (*Encrypter, error) {
 
 		address: c.Address,
 		base:    base,
+		nonce:   "3c4caa13-dc08-f51f-51f8-02c4abb8d89f",
+		token:   "98b8b1aa-fa75-adba-7dd8-0499ba591bbd",
```
kopiczko: Very naive q: is it safe to keep it in the code?

fgimenez: Nope, just for testing purposes.

service/controller/v13/templates/cloudconfig/decrypt_tls_assets_script.go
```diff
@@ -1,7 +1,95 @@
...
+# if curl pkcs7 fails, exit with error logged
+token_exists () {
+  if [ -f $token_path ]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+token_is_valid() {
```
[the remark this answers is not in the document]
fgimenez: Done, there's a single template constant for the decryption script now.

service/controller/v13/cloudconfig/master_template.go
kopiczko: My suggestion is to pass the context to `Files` and then `templateData` method.

service/controller/v13/encrypter/vault/vault.go
fgimenez: Authentication from the container has been sorted, the hardcoded token has been removed.

service/controller/v13/cloudconfig/master_template.go
fgimenez: The problem I see with that approach is that `Files` is called while rendering the template, for the master template here https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_3_3/master_template.go#L1810. Pls correct me if I'm wrong but according to https://golang.org/pkg/text/template/#hdr-Arguments we can only use niladic methods on templates.

fgimenez:
> For the duplication in the bash scripts. The approach discussed of using a Go binary will help with that. It would be good to have an issue for that.

Done, created this issue for tracking https://github.com/giantswarm/giantswarm/issues/3440.

service/controller/v13/cloudconfig/master_template.go
kopiczko: You're right I missed that. IMO it's better to use context.TODO in such case.

kopiczko: That's way better!

fgimenez:
CI green, going with this.

PR #1025, opened 2018-06-20
Add public routes for accessing Vault ELB

fgimenez:
Without these routes we can't access Vault ELB endpoint from guest cluster nodes. This is required for using host cluster Vault as encryption backend, guest cluster nodes need to be able to access it to decrypt keys and TLS assets.


rossf7:
Thanks for confirming about updates.

> If at some point we want to migrate an existing cluster from KMS to Vault we would need anyway logic for doing changes in the keys

Yes if we decide to do that we can handle it as part of the migration task.

PR #1027, opened 2018-06-21
vendor: use updated e2eclients

rossf7:


integration/test/draining/main_test.go
```diff
@@ -48,7 +48,10 @@ func TestMain(m *testing.M) {
 		}
 	}
 
-	c = aws.NewClient()
+	c, err = aws.NewClient()
```
rossf7: Needed because of the error handling merged earlier in e2eclients.

PR #1028, opened 2018-06-22
Removes alerter service

JosephSalisbury:
See https://github.com/giantswarm/g8s-prometheus/pull/272 for removing alert.

The collector handles the VPC orphaning alerts nowadays, and the duplicate master checks don't make much sense with the move to cloudformation. This changeset removes the alerter service, as part of paving the road towards support for BYOC monitoring.


MarcelMue:
Just to get this straight in my head:
This was meant to detect duplicate resources of any kind. We now have a collector though which checks for orphaned resources.
So with cloud formation we should never have duplicate resources?

PR #1033, opened 2018-06-26
handle bound_iam_role_arn as string alice

fgimenez:
With the recent update of Vault we should handle `bound_iam_role_arn` as a string slice. Also we only manage this field now, no need to unmarshal/marshal the rest of them if we only change the IAM roles.


fgimenez:
Tested on axolotl

PR #1034, opened 2018-06-27
Delete unused resource sets

rossf7:
Towards giantswarm/giantswarm#3526

See https://github.com/giantswarm/giantswarm/issues/3526#issue-335816796 for which resource sets can be deleted.


rossf7:
OK changes affect integration tests too. I'll get that sorted after lunch.

PR #1038, opened 2018-07-02
Add registry domain parameter

fgimenez:
Defines registry domain parameter (initialized from installations) and sets it as value for k8scloudconfig.


fgimenez:
Tested on axolotl and gauss.

PR #1039, opened 2018-07-04
increase wait time for Vault connection available

fgimenez:
5 minutes is not always enough, increasing to 20.

For context, with Vault encryption we use the peering connection for accessing Vault from guest nodes, this resource is created as part of the guest stack, and the routes for the actual access are created in a host stack after the guest stack is done. We need to wait on guest cluster nodes for the routes to be in place in order to be able to decrypt the secrets.

### DIFF
--- a/service/controller/v13/templates/cloudconfig/decrypt_keys_assets_script.go
+++ b/service/controller/v13/templates/cloudconfig/decrypt_keys_assets_script.go
@@ -8,7 +8,7 @@ nonce_path=/var/nonce
 wait_for_vault_elb(){
     local service_name="$1"
     local state="${2:-active}"
-    for i in $(seq 20); do
+    for i in $(seq 80); do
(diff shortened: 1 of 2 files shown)


fgimenez:
Slowness on China is a factor, of course. In this case, route53 doesn't affect, the entry for host cluster vault is created beforehand. 

Recently creation is slower everywhere, looks like we are getting timeouts on `randomkeys.Search`, see https://gigantic.slack.com/archives/C2MS2VB37/p1530686756000192. I think this issue affects the timeouts for connecting to Vault from guests nodes on boot in axolotl, didn't see them in the first tests. This new timeout will cope with the `randomkeys.Search` delays.

rossf7:
Thanks for the explanation, makes sense its not Route 53.

I think we need to worry if cluster creation is slowing down. AFAIK we moved the encryption key secret to cluster-operator. We also dropped the resync period there to 1 min. This should help but we need to see what is slowing things down and maybe more  fixes are needed.

PR #1041, opened 2018-07-06
Fix s3object deletion for Vault encrypter

fgimenez:
Fixes check for deleted key on desired s3object state, we were using always `kms` logic, even for when using vault encrypter.

Also the vault encryption keys are now explicitly configured to allow deletion to prevent potential issues.


fgimenez:
Tested on axolotl

service/controller/v13/encrypter/vault/vault.go
```diff
@@ -99,6 +96,28 @@ func (e *Encrypter) CreateKey(ctx context.Context, customObject v1alpha1.AWSConf
 		return microerror.Mask(invalidHTTPStatusCodeError)
 	}
 
+	// we need to make the key erasable.
```
rossf7: Nitpicking but comment should start with a capital. Also could be deletable rather than erasable.

service/controller/v13/encrypter/vault/vault.go
```diff
@@ -174,20 +190,42 @@ func (e *Encrypter) CurrentState(ctx context.Context, customObject v1alpha1.AWSC
...
+		return "", microerror.Mask(err)
+	}
+	defer resp.Body.Close()
 
-	return clusterID, nil
+	if resp.StatusCode == http.StatusNotFound {
+		return "", microerror.Mask(keyNotFoundError)
+	} else if resp.StatusCode != http.StatusOK {
+		body, _ := ioutil.ReadAll(resp.Body)
+		return "", microerror.Maskf(invalidHTTPStatusCodeError, "want 200, got %d, response body: %q", resp.StatusCode, body)
```
rossf7: Again nitpicking but you could also use `http.StatusOK` in the error message.

service/controller/v13/encrypter/vault/vault.go
fgimenez: Sure, done.

service/controller/v13/encrypter/vault/vault.go
fgimenez: Good point, done.

PR #1042, opened 2018-07-08

rossf7:
Enables CRD sub resource feature gate.


rossf7:
Closing. Feature gate is enabled in master.

PR #1046, opened 2018-07-10
Update v13 changelog

fgimenez:
`v13` changelog got stale, this adds the missing entries according to https://github.com/giantswarm/aws-operator/commits/master/service/controller/v13

### DIFF
--- a/service/controller/v13/version_bundle.go
+++ b/service/controller/v13/version_bundle.go
@@ -17,6 +17,41 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Added CA public key into trusted user keys for SSO ssh.",
 				Kind:        versionbundle.KindAdded,
 			},
+			{
+				Component:   "aws-operator",
+				Description: "Fix volume attachments while upgrading.",
+				Kind:        versionbundle.KindFixed,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Add encrypter interface, make KMS code satisfy it and add Vault encrypter.",
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Enabled migration resource with credentials logic.",
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Add public for accessing Vault ELB when using Vault encrypter.",
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Improve logging for shouldScale method.",
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Add registry domain parameter and inject it into k8scloudconfig.",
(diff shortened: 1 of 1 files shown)


service/controller/v13/version_bundle.go
```diff
@@ -17,6 +17,41 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Added CA public key into trusted user keys for SSO ssh.",
 				Kind:        versionbundle.KindAdded,
 			},
+			{
+				Component:   "aws-operator",
+				Description: "Fix volume attachments while upgrading.",
```
rossf7: This should be Fixed as the message should be past tense. Would be good to align with `fmt`.

https://github.com/giantswarm/fmt/blob/master/releases/how_write_changelog_message.md

I find that page super helpful because I always forget all this! ;)

service/controller/v13/version_bundle.go
```diff
@@ -17,6 +17,41 @@ func VersionBundle() versionbundle.Bundle {
...
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Add public for accessing Vault ELB when using Vault encrypter.",
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Improve logging for shouldScale method.",
```
rossf7: Not sure we need a changelog for this. It's a low level change and not interesting to customers IMO.

service/controller/v13/version_bundle.go
```diff
@@ -17,6 +17,41 @@ func VersionBundle() versionbundle.Bundle {
...
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Add registry domain parameter and inject it into k8scloudconfig.",
```
rossf7: Maybe add docker registry to make this clearer?

service/controller/v13/version_bundle.go
```diff
@@ -17,6 +17,41 @@ func VersionBundle() versionbundle.Bundle {
...
+				Kind:        versionbundle.KindAdded,
+			},
+			{
+				Component:   "aws-operator",
+				Description: "Set version bundle version output via parameter.",
```
rossf7: stack output?

service/controller/v13/version_bundle.go
fgimenez: Done, thanks for the link!

service/controller/v13/version_bundle.go
fgimenez: Indeed, removed.

service/controller/v13/version_bundle.go
fgimenez: Good point, done.

service/controller/v13/version_bundle.go
fgimenez: Done, PTAL.

PR #1051, opened 2018-07-10
Fix version for v13

rossf7:
This should be a minor version not a patch. I messed this up when I added v13. Updating and 3.1.3 will be used for v12patch1.

See https://github.com/giantswarm/installations/pull/419 for the release index change.


rossf7:
Going with this and the release index change. I've confused the tests my changing the version. I'll make sure they are green on my next PR.

PR #1055, opened 2018-07-11
reconcile services

calvix:
As the annotation is now needed for guest cluster monitoring we need to be able to update old cluster and the that means enable reconciliation of the service to desired state

Also since the etcd port is different across providers and it needs to be part of annotation


calvix:
closing as changes need to go in v14

PR #1060, opened 2018-07-11
v14: add bridgezone resource

kopiczko:
Towards giantswarm/giantswarm#3314

**Please check https://github.com/giantswarm/aws-operator/pull/1058 first.**


kopiczko:
Upgrades work. Tested on ginger. They don't produce NS record in the host cluster zone, because host-post stack doesn't reconcile. This isn't a blocker because we have NS records in "intermediate" zone, but it's required for the final cleanup. This is a whole rabbit hole though and I want to address that it other PRs. Slack discussion reference: https://gigantic.slack.com/archives/C2MS2VB37/p1531909991000147

service/controller/v14/credential/credential.go
```diff
@@ -28,6 +28,24 @@ func GetARN(k8sClient kubernetes.Interface, obj interface{}) (string, error) {
 	return arn, nil
 }
 
+// GetDefaultARN is used only by the bridgezone resource. It should be removed
+// when the resource is removed.
+func GetDefaultARN(k8sClient kubernetes.Interface) (string, error) {
```
xh3b4sd: Any reason why there is no service like implementation which is configured with a k8s client? When this is for BYOC I thought we have the clients being transported via the context so I would have expected this mechanism to be used here.

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if !r.route53Enabled {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "route53 disabled, skipping execution")
```
xh3b4sd: The cancelation logs are inconsistent. I would like to see the usual pattern with a log for what is going on and a log for the cancelation. This is then more copy paste and not rephrasing everything over and over again. Like, don't make me think. When I can scan the pattern I know I am quicker and more efficient.

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+		return nil
+	}
+
+	baseDomain := key.BaseDomain(customObject)
+	intermediateZone := "k8s." + baseDomain
```
xh3b4sd: I feel like the domain construction is now just done everywhere. Maintaining this is no fun. How can we make this better?

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	var intermediateZoneID string
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "searching for intermediate zone in default guest account")
+
+		intermediateZoneID, err = r.findHostedZoneID(ctx, defaultGuest, intermediateZone)
+		if IsNotFound(err) {
```
xh3b4sd: When can this actually happen and how can we still ensure the cleanup of certain resources or configurations?

service/controller/v14/resource/cloudformation/current.go
```diff
@@ -67,7 +67,17 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		var hostedZoneNameServers string
 		if r.route53Enabled {
 			hostedZoneNameServers, err = sc.CloudFormation.GetOutputValue(stackOutputs, key.HostedZoneNameServers)
-			if err != nil {
+			// TODO This is for migration from v13 to v14. Remove this check when all guest clusters are managed by v14 or newer.
```
xh3b4sd: Can we track this with an issue please?

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "deletion of intermediate zone delegation ensured")
+	}
+
+	return nil
+}
+
+func (r *Resource) findHostedZoneID(ctx context.Context, client *route53.Route53, name string) (string, error) {
```
xh3b4sd: Would it make sense to have unit tests for this?

service/controller/v14/credential/credential.go
kopiczko: I just followed the pattern we already have https://github.com/giantswarm/aws-operator/pull/1060/files#diff-2a88f9bcdf0b0ee008cb4eae77ddc9b5R17

This function is here only for the migration period. As comment says it should be removed when `bridgezone` resource is removed.

service/controller/v14/resource/bridgezone/resource.go
kopiczko: This can happen when we have a brand new installation or when we are after the migration period.
In general this resource does changes only to the intermediate zone, so if it doesn't exist we don't have to care about the deletion.

kopiczko: I also wrote in the Resource godoc that when intermediate doesn't exist this resource becomes noop.

service/controller/v14/resource/bridgezone/resource.go
kopiczko: W/o mocking the whole client and making potential wrong assumptions this is the only part that is testable:

```
+			if hz.Name == nil || hz.Id == nil {
+				continue
+			}
+
+			hzName := *hz.Name
+			hzName = strings.TrimSuffix(hzName, ".")
+			hzID := *hz.Id
+
+			if hzName == name {
+				return hzID, nil
+			}
```

This is 3 LoC. I'm pretty confident it works as expected.

service/controller/v14/resource/cloudformation/current.go
kopiczko: I added point here https://github.com/giantswarm/giantswarm/issues/3314

```
Fix v14/resource/cloudformation/current.go TODO
````

File: service/controller/v14/resource/bridgezone/resource.go [hunk 7]
```diff
@@ -0,0 +1,415 @@
...
+//
+// New structure looks like:
+//
+//	installation.eu-central-1.aws.gigantic.io (host account)
+//	└── NS new_cluster.k8s.installation.eu-central-1.aws.gigantic.io (byoc guest account)
+//	    ├── A api.new_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//	    └── A ingress.new_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//
+// For the migration period for new clusters we need also to add delegation to
+// k8s.eu-central-1.aws.gigantic.io because of the AWS DNS caching issues.
```
TheoBrigitte: Missing `installation` in dns name should be > `k8s.installation.eu-central-1.aws.gigantic.io`

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+//	├── NS k8s.installation.eu-central-1.aws.gigantic.io (default guest account)
+//	│   ├── NS new_cluster.k8s.installation.eu-central-1.aws.gigantic.io (byoc guest account)
+//	│   ├── A api.old_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//	│   └── A ingress.old_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//	└── NS new_cluster.k8s.installation.eu-central-1.aws.gigantic.io (byoc guest account)
+//	    ├── A api.new_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//	    └── A ingress.new_cluster.k8s.installation.eu-central-1.aws.gigantic.io
+//
+// NOTE: In the code below k8s.eu-central-1.aws.gigantic.io zone is called
+// "intermediate" and new_cluster.k8s.eu-central-1.aws.gigantic.io zone is
```
TheoBrigitte: same remark about dns name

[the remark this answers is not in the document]
kopiczko: Thanks <3 Yes, this is because of viking where we don't have installation in the root zone name :(

service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -0,0 +1,415 @@
...
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "got final zone name servers and TTL from intermediate zone")
+	}
+
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring intermediate zone delegation")
+
+		upsert := route53.ChangeActionUpsert
+		ns := route53.RRTypeNs
+		ttl := int64(900)
```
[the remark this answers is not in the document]
kopiczko: I wouldn't say so. This is the only place we use this number. I personally don't like jumping around.

service/controller/v14/resource/bridgezone/resource.go
kopiczko: Done

kopiczko:
All green. Going.

PR #1082, opened 2018-07-25
vendor: update

kopiczko:


service/controller/v14/cluster_resource_set.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/giantswarm/operatorkit/controller/resource/metricsresource"
 	"github.com/giantswarm/operatorkit/controller/resource/retryresource"
 	"github.com/giantswarm/randomkeys"
+	_ "github.com/giantswarm/statusresource"
```
kopiczko: @xh3b4sd I just added this import so dep doesn't freak out. I also didn't want to change Gopkg.toml.

PR #1084, opened 2018-07-25
v14/resource/bridgezone: fix handling already deleted delegation

kopiczko:
Fixes https://github.com/giantswarm/giantswarm/issues/3838

### DIFF
--- a/service/controller/v14/resource/bridgezone/resource.go
+++ b/service/controller/v14/resource/bridgezone/resource.go
@@ -379,16 +379,16 @@ func (r *Resource) getNameServersAndTTL(ctx context.Context, client *route53.Rou
 	}
 
 	if len(out.ResourceRecordSets) == 0 {
-		return nil, 0, microerror.Maskf(notFoundError, "NS recrod %q for HostedZone %q not found", name, zoneID)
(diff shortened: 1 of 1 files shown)


service/controller/v14/resource/bridgezone/resource.go
```diff
@@ -388,7 +388,7 @@ func (r *Resource) getNameServersAndTTL(ctx context.Context, client *route53.Rou
 	rs := *out.ResourceRecordSets[0]
 
 	if strings.TrimSuffix(*rs.Name, ".") != name {
-		return nil, 0, microerror.Maskf(executionError, "expected NS recrod with name %q , found %q", name, *rs.Name)
+		return nil, 0, microerror.Maskf(notFoundError, "NS recrod %q for HostedZone %q not found", name, zoneID)
```
kopiczko: This works well when there is only one or zero delegations. Pretty hard to find in the test env.

rossf7: Nitpick and already there but would be nice to fix the typo on `record`.

kopiczko:
Tested on ginger

PR #1088, opened 2018-07-26
v14: changelog

kopiczko:


service/controller/v14/version_bundle.go
```diff
@@ -9,8 +9,8 @@ func VersionBundle() versionbundle.Bundle {
 		Changelogs: []versionbundle.Changelog{
 			{
 				Component:   "aws-operator",
-				Description: "Add your changes here.",
-				Kind:        versionbundle.KindChanged,
+				Description: "Added Route53 support for BYOC accoutns.",
+				Kind:        versionbundle.KindAdded,
```
kopiczko: Is this the only change?

rossf7: I don't know of any other changes and k8scc didn't change.

Also just a typo on accounts.

kopiczko:
@rossf7 I changed version here to v3.3.0 and I'm going to leave 4.0.0 in v15. Does that make sense for you?

rossf7:
Thanks, good call on the versioning.

I need to do a release for the IC migration. Depending on when K8s 1.11 is ready I can change v15 to v3.4.0.

PR #1091, opened 2018-07-26
v15: adapter refactor

kopiczko:
❗️ Follow up of #1090 ❗️ 

❗️ **This is already reviewed, don't waste the time on looking at the code** ❗️


kopiczko:
❗️ Follow up of #1090 ❗️ 

❗️ **This is already reviewed, don't waste the time on looking at the code** ❗️

PR #1095, opened 2018-07-30
Cancel resource for deletion when CF stacks are updating

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/3672


kopiczko:
I feel like this is stale. Feel free to reopen.

PR #1098, opened 2018-07-31
v15: remove encrypter dependency from cloudformation resource

kopiczko:


service/controller/v15/encrypter/vault/vault.go
```diff
@@ -319,68 +395,6 @@ func (e *Encrypter) Decrypt(key, ciphertext string) (string, error) {
 	return string(plaintext), nil
 }
```
kopiczko: Methods below replaced with `Ensure` ones. Logic is similar. They are just idempotent.

service/controller/v15/resource/cloudformation/create.go
```diff
@@ -29,18 +28,6 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 			return microerror.Mask(err)
 		}
```
kopiczko: All that hackery removed from `cloudformation` resource.

service/controller/v15/encrypter/spec.go
```diff
@@ -19,33 +18,18 @@ type EncryptionKeyState struct {
 }
```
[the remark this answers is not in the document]
kopiczko: I think we should move parts of `encrypter` to the `encryptionkey` resource. Here I'm trying to draw the line between stuff used by the `encryptionkey` to reconcile and other resource to encrypt things. My vision is to not have `Resource` interface at all.

service/controller/v15/encrypter/vault/vault.go
```diff
@@ -197,6 +201,78 @@ func (e *Encrypter) DesiredState(ctx context.Context, customObject v1alpha1.AWSC
...
+
+	role, err := e.getAWSAuthRole(p)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	combinedIAMRoleARNs := stringSlice(role.BoundIAMRoleARN).Add(iamRoleARNs...)
+
+	if len(combinedIAMRoleARNs) != len(role.BoundIAMRoleARN) {
+		e.logger.LogCtx(ctx, "level", "debug", "message", "adding authorized IAM role ARMs in the Vault API")
```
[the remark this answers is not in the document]
kopiczko: Too much Azure.

service/controller/v15/resource/cloudformation/create.go
kopiczko: I think it was a hack from the very beggining. Hard to say why it was designed like that.
I don't have ideas how to raise awareness of those things. We should just keep in mind that all we write should be idempotent. But that should be actually the company mindset.
I hope logging guides we have now will make it more visible and clear. I'm happy see more guidelines, but I don't think we can describe every case. At some point we just have to rely on each other.

kopiczko:
@rossf7 I would appreciate your review here. I think you are deeper into the topic maybe you spot something I missed.

service/controller/v15/cluster_resource_set.go
```diff
@@ -132,30 +132,33 @@ func NewClusterResourceSet(config ClusterResourceSetConfig) (*controller.Resourc
...
 		}
 
-		encrypterObject, err = vault.NewEncrypter(c)
+		e, err := vault.NewEncrypter(c)
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
-		encrypterRoleManager = encrypterObject.(encrypter.RoleManager)
+		encrypterObject = e
+		encrypterResource = e
```
rossf7: I get that we want to split the object and resource. As well as eventually removing the resource.

Still this looks weird to me and I'd appreciate some explanatory comments.

service/controller/v15/encrypter/kms/kms.go
```diff
@@ -115,7 +115,7 @@ func (k *Encrypter) DeleteKey(ctx context.Context, keyAlias string) error {
 	return nil
 }
 
-func (k *Encrypter) CurrentState(ctx context.Context, customObject v1alpha1.AWSConfig) (encrypter.EncryptionKeyState, error) {
+func (k *Encrypter) GetCurrentState(ctx context.Context, customObject v1alpha1.AWSConfig) (encrypter.EncryptionKeyState, error) {
```
rossf7: I like the renaming.

service/controller/v15/encrypter/spec.go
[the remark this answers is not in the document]
rossf7: As per my other comment. I'm fine with removing the Resource interface later but lets explain that in the comments.

service/controller/v15/encrypter/vault/string_slice.go
```diff
@@ -0,0 +1,42 @@
+package vault
+
+type stringSlice []string
```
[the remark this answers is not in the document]
rossf7: Good to extract this and nice to have tests.

kopiczko:
@rossf7 I would test that, but I have problems interacting with that env. Slack thread https://gigantic.slack.com/archives/C3S482Z98/p1533139664000402

service/controller/v15/encrypter/kms/kms.go
```diff
@@ -145,6 +145,20 @@ func (k *Encrypter) DesiredState(ctx context.Context, customObject v1alpha1.AWSC
...
+	k.logger.LogCtx(ctx, "level", "debug", "message", "endured authorized vault IAM roles")
+	return nil
+}
+
+func (k *Encrypter) EnsureDeletedAuthorizedIAMRoles(ctx context.Context, iamRoleARNs ...string) error {
+	k.logger.LogCtx(ctx, "level", "debug", "message", "ensuring deletion of authorized vault IAM roles")
+	k.logger.LogCtx(ctx, "level", "debug", "message", "no action for KMS backend")
+	k.logger.LogCtx(ctx, "level", "debug", "message", "endured deletion of authorized vault IAM roles")
+	return nil
+}
```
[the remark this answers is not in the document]
fgimenez: A comment here would help too IMO.

service/controller/v15/encrypter/vault/vault.go
```diff
@@ -197,6 +201,78 @@ func (e *Encrypter) DesiredState(ctx context.Context, customObject v1alpha1.AWSC
...
+	p := path.Join("auth", "aws", "role", decrypterVaultRole)
+
+	role, err := e.getAWSAuthRole(p)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	combinedIAMRoleARNs := stringSlice(role.BoundIAMRoleARN).Add(iamRoleARNs...)
+
+	if len(combinedIAMRoleARNs) != len(role.BoundIAMRoleARN) {
```
TheoBrigitte: One improvement here could be to have `stringSlice.Add` returning the number of added elements (so modifying the object in place instead of returning a new one).
And then just check if number of added elements is greater than 1.
No blocker just a suggestion.

kopiczko:
Closing in favour of https://github.com/giantswarm/aws-operator/pull/1170

PR #1101, opened 2018-08-02
v15: Cancel reconciliation on delete when stack is updating

rossf7:
Fixes https://github.com/giantswarm/giantswarm/issues/3672

Cancels reconciliation and preserves finalizers if the stack state is updating.

### DIFF
--- a/service/controller/v15/resource/cloudformation/delete.go
+++ b/service/controller/v15/resource/cloudformation/delete.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/controller"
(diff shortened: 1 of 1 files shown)


rossf7:
Tested on ginger.

rossf7:
Ack. Pawel looks like you're already on this. Do you want me to close?

It's a fix for giantswarm/giantswarm#3672 which is a PM that got assigned to me.

kopiczko:
I was unaware of https://github.com/giantswarm/giantswarm/issues/3672 but https://github.com/giantswarm/aws-operator/pull/1099 should cover that.

I will try to test & merge tomorrow.

rossf7:
Cool thanks I'll close this.

I tested on ginger and when the finalizer is kept then the deletion is repeated successfully even during a stack update. Since #1099 has this it should sort the PM.

PR #1102, opened 2018-08-02
Create multiple security groups when there more than 50 rules

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/3860


pipo02mix:
Closing it, we have chosen to increase SG max rules instead

PR #1104, opened 2018-08-03
integration, vendor: update e2etemplates

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/3892


integration/setup/setup.go
```diff
@@ -167,39 +172,96 @@ func WrapTestMain(c *awsclient.Client, g *framework.Guest, h *framework.Host, m
...
+func installAWSOperator(h *framework.Host) error {
+	var err error
 
-		awsResourceChartValuesEnv := os.ExpandEnv(e2etemplates.ApiextensionsAWSConfigE2EChartValues)
-		d := []byte(awsResourceChartValuesEnv)
+	var values string
+	{
+		c := chartvalues.AWSOperatorConfig{
+			Guest: chartvalues.AWSOperatorConfigGuest{
+				// TODO remove this setting
```
kopiczko: I should just remove it. It's always set to true in the tests.

integration/setup/setup.go
```diff
@@ -167,39 +172,96 @@ func WrapTestMain(c *awsclient.Client, g *framework.Guest, h *framework.Host, m
...
+
+func installAWSConfig(h *framework.Host, vpcPeerID string) error {
+	var err error
+
+	var values string
+	{
+		c := chartvalues.APIExtensionsAWSConfigE2EConfig{
+			CommonDomain: env.CommonDomain(),
+			// TODO(PK) can this be replaced with env.ClusterID?
+			ClusterName:          env.ClusterName(),
```
kopiczko: This is weird. Why it is ClusterName? Why there isn't value from `ClsuterID`. This shows getting rid of env vars is already useful.

integration/test/clusterstate/project.yaml
```diff
@@ -7,7 +7,6 @@ test:
     - "CIRCLECI=${CIRCLECI}"
     - "CLUSTER_NAME=${CLUSTER_NAME}"
     - "COMMON_DOMAIN=${COMMON_DOMAIN}"
-    - "GUEST_UPDATE_ENABLED=true"
```
kopiczko: Not used. CLUSTER_NAME maybe also can be removed.

integration/env/env.go
```diff
@@ -12,106 +12,110 @@ import (
...
+	getenv("GUEST_AWS_SECRET_ACCESS_KEY", &guestAWSAccessKeySecret)
+	getenv("GUEST_AWS_SESSION_TOKEN", &guestAWSAccessKeyToken)
+	getenv("HOST_AWS_ACCESS_KEY_ID", &hostAWSAccessKeyID)
+	getenv("HOST_AWS_SECRET_ACCESS_KEY", &hostAWSAccessKeySecret)
+	getenv("HOST_AWS_SESSION_TOKEN", &hostAWSAccessKeyToken)
+	getenv("IDRSA_PUB", &idRSAPub)
+	getenv("TESTED_VERSION", &testedVersion)
+	getenv("TEST_DIR", &testDir)
+	getenv("REGISTRY_PULL_SECRET", &registryPullSecret)
+	getenv("VAULT_TOKEN", &vaultToken)
```
kopiczko: I was so annoyed because of jumping around with those consts that I just did it like that. You can see this list here is ~3 times longer now.

kopiczko:
Chances are tests are failing right now. I'm requesting review so we can discuss approach in general.

integration/setup/setup.go
```diff
@@ -30,7 +30,7 @@ const (
 	credentialNamespace   = "giantswarm"
 )
 
-func HostPeerVPC(c *awsclient.Client, g *framework.Guest, h *framework.Host) error {
+func HostPeerVPC(c *awsclient.Client, g *framework.Guest, h *framework.Host) (string, error) {
```
kopiczko: When I look at that I feel like we should create `Setup` struct which would take all the clients and frameworks and provide install methods.

integration/env/env.go
xh3b4sd: - We tried to differentiate between common and provider specific env vars. Check for instance the `azure-operator`'s env package. My idea here was to have a common env package in `e2etests` but we should separate the code already to make the move more easy. 
- I think the approach how it was was just fine. I cannot comprehend why this should be better or why you changed it. I am rather against the new style. It is how it was in all other projects as well. May you want to elaborate a bit more on the problems you have so we can think how we can address your personal issues.

integration/setup/setup.go
xh3b4sd: The env var was called cluster name for some reason, which is just wrong IMO. It should just be cluster ID. There were also TODO comments to fix that naming.

kopiczko: But in the old template it was `${CLUSTER_NAME}` placeholder. And this had a different value than `env.ClusterID()`. Also `env.ClusterName() != env.ClusterID()` now.

integration/env/env.go
rossf7: +1 on separating common and provider. We need to be able to easily run e2e tests that don't have a guest cluster.

We also want to split logic in e2e-harness to make it more composable. See https://github.com/giantswarm/giantswarm/issues/3878 for the issue created by @MarcelMue. The refactoring going on is great and needed but we need to keep it aligned IMO.

cc @indyfree

integration/setup/setup.go
rossf7: @MarcelMue's idea from https://github.com/giantswarm/giantswarm/issues/3878 is to have a single config in `e2e-harness/pkg/framework` that takes in all clients.

Config will be optional so we only validate if we use host or guest functionality. This should let us align the setup across all projects. I think it's a good approach and could help here.

integration/env/env.go
kopiczko: I'll refactor it back. I wouldn't have an issue when adding a sing var. Copying common.go from azure-operator will make things easier.

integration/setup/setup.go
kopiczko: We can get rid of CLUSTER_NAME. It was set to CLUSTER_ID in case former was empty. And that always was the case.

integration/env/env.go
kopiczko: Prime example why I think current approach is inconvenient: https://github.com/giantswarm/aws-operator/pull/1104/commits/23493cafb8fca54efb01695c7a3c066bb1ec49dd

integration/setup/setup.go
```diff
@@ -168,43 +173,100 @@ func WrapTestMain(c *awsclient.Client, g *framework.Guest, h *framework.Host, m
...
 			return microerror.Mask(err)
 		}
+
 	}
 
-	o := func() error {
-		// NOTE we ignore errors here because we cannot get really useful error
-		// handling done. This here should anyway only be a quick fix until we use
-		// the helm client lib. Then error handling will be better.
-		framework.HelmCmd("delete --purge aws-config-e2e")
```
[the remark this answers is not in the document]
rossf7: Not sure either. Each test is in a fresh cluster so the delete shouldn't be needed.

Could just be because the tests here are older. I think the approach in `azure-operator` with the `Resources` function that installs everything is the right one.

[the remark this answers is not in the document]
fgimenez: This is useful for local executions, you need proper teardown in order to rerun.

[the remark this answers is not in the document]
rossf7: Ah that makes sense. Being able to run tests locally helps a lot when changing them.

The `teardown` logic should cover deleting these resources. When running locally could we run `teardown` before `setup`?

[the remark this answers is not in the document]
fgimenez: You are right, we should rely on the teardown logic and not remove resources on setup. I guess the reason why this was added here is the current teardown approach https://github.com/giantswarm/aws-operator/blob/master/integration/teardown/teardown.go#L30-L60 if something goes wrong removing one of the first resources the rest of them can be left behind because of the early return. We could change this IMO.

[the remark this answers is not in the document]
rossf7: Totally agree. If we can make `setup` and `teardown` both idempotent that would improve things a lot IMO.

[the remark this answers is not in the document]
kopiczko: Addressed here https://github.com/giantswarm/aws-operator/pull/1112

kopiczko:
PTALA. This works now https://circleci.com/workflow-run/e24d9b53-7414-4dd2-8242-b667706a15bc

PR #1113, opened 2018-08-13
integration/teardown: fix undefined err

kopiczko:


kopiczko:
I didn't know e2e code isn't event compiled when e2e tests aren't ran.

kopiczko:
This one looks good now.

PR #1125, opened 2018-08-22
Set v15 version to 4.0.0

rossf7:
Updates version to match release index.


rossf7:
See https://gigantic.slack.com/archives/C2MS2VB37/p1534925919000100

rossf7:
Merging to unblock folk and because change is minor. I'll watch for any issues with e2e.

PR #1127, opened 2018-08-23
add tenant cluster operator AWS IAM Policy

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/3675

Every tenant cluster make use of the same IAM Policy which is documented in our [doc](https://github.com/giantswarm/docs-content/blob/master/content/guides/prepare-aws-account-for-guest-clusters/index.md#3-permissions-setup), order to have a single place of truth which can also be edited, I am writing this down in this repo (as it was done for Azure https://github.com/giantswarm/azure-operator/pull/291).


marians:
Should we name it to fit the "tenant cluster" term? The file is now called `guest.json`.

PR #1128, opened 2018-08-23
Add SG rule aim to the description field

pipo02mix:
Towards https://github.com/giantswarm/vodafone/issues/357


service/controller/v15/templates/cloudformation/guest/security_groups.go
```diff
@@ -10,13 +10,15 @@ const SecurityGroups = `{{define "security_groups" }}
       SecurityGroupIngress:
       {{ range $v.MasterSecurityGroupRules }}
       -
+        Description: {{ .Description }}
```
rossf7: <3

pipo02mix:
I tested and it worked fine. PTAL (I will wait for the green but I had to update branch twice after all green was there before :) )

pipo02mix:
I am merging beside some test fails. It passed correctly for all before update branch a couple of times. The failures are flapping (some times one fails and then it passes) and the reason is API not up.

PR #1130, opened 2018-08-24
v14patch1: copy v14

rossf7:
Adding patch release to disable ingress controller resources in k8scloudconfig. This is for migrating IC to chart-operator on K8s 1.10 clusters.


rossf7:
Copy PR. Going with this.

PR #1133, opened 2018-08-24
Adds v14patch1 to disable ingress controller in k8scloudconfig

rossf7:
This is so we can migrate IC to chart-operator on k8s 1.10 clusters. Change is also added to v15 for k8s 1.11 but it will take customers time to move.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,18 +2,15 @@
 
 
 [[projects]]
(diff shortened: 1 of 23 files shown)


rossf7:
7/8 e2e tests passed. Failure was for the current release not WIP. Good enough for me so going with this.

PR #1137, opened 2018-08-28
add credential-default to helm chart

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/3815

Add `credential-default` which is now required by aws-operator to operate tenant cluster.


TheoBrigitte:
I noticed that also, my guess is this is a trick to enforce a certain order when deploying the chart.

PR #1138, opened 2018-08-29
v16: copy v15

rossf7:
Copy PR with new v16 WIP release. I know there are some open PRs but we need an active K8s 1.11 release.


rossf7:
New code isn't active so just going with the build step. I'll make sure full CI passes when v16 is wired up.

PR #1139, opened 2018-08-29
controller/v16: update imports and packages

rossf7:


service/controller/v16/version_bundle.go
```diff
@@ -1,4 +1,4 @@
-package v15
+package v16
```
fgimenez: Shouldn't the `Version` in this file be bumped too? Not sure if this is done later.

rossf7: Good to check. I'll do this in the next PR when I wire up v16.

To me doing it this way makes sense. As these changes were just done by the bash script.

PR #1142, opened 2018-08-29
Remove descriptions for SG subtle to be affected

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/4109


pipo02mix:
Should I need it to create as a patch?

PR #1154, opened 2018-09-10
Add description for security group rules for master node

pipo02mix:
Towards https://github.com/giantswarm/vodafone/issues/357


pipo02mix:
Tested with upgrades even. PTAL

PR #1157, opened 2018-09-11
integration/setup: try harder to delete host VPC

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4059


kopiczko:
OK, let's merge this and I will work on teardown before setup in other PR.

integration/setup/teardown.go
```diff
@@ -44,13 +44,13 @@ func teardown(config Config) error {
...
-	}
-
 	err = deleteHostPeerVPC(config)
 	if err != nil {
-		return microerror.Mask(err)
+		errors = append(errors, microerror.Mask(err))
+	}
+
+	if len(errors) > 0 {
+		return microerror.Mask(errors[0])
```
tuommaki: Would it make sense to log rest of the `errors` before returning? I believe they end up being swallowed otherwise and this could lead to real issue being hidden by some warning'a'like error overriding it WDYT?

kopiczko: For me it even makes sense to not return error at all in the teardown. We don't do any actions based on that anyway.

kopiczko: Done.

kopiczko:
PTALA

kopiczko:
e2e tests passed here https://circleci.com/workflow-run/01a8f65a-f6c3-4d50-9eb9-623dca44e6e8

PR #1162, opened 2018-09-13
update vendor

r7vme:
in particular `certs` library


r7vme:
Ok e2e passed. Merging.

PR #1167, opened 2018-09-18
integration/setup: perform teardown before execution

kopiczko:
Fixes https://github.com/giantswarm/giantswarm/issues/4059.


kopiczko:
2 tests failed, but I'm not so worried this is general change and tests don't fail on introduced code.

PR #1169, opened 2018-09-20
Disable CoreDNS K8scc templates

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902


fgimenez:
Good catch, I've added an entry similar to what we have in `v15` for `nginx-ingress-controller`, PTAL

service/controller/v17/version_bundle.go
```diff
@@ -12,6 +12,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Add security group descriptions for master nodes.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "cloudconfig",
+				Description: "Removed CoreDNS related components (will be managed by chart-operator).",
```
kopiczko: Nit: I would skip `will be`. They should be already managed at the time we release this version.

rossf7: To me its not clear without the will be but `(now managed by chart-operator)' is shorter and clear.

kopiczko: As I see it's only done in v15.

fgimenez: Good catch, fixed.

fgimenez:
For the record this is not working, when we remove the k8scc CoreDNS manifests Calico networking fails. See https://github.com/giantswarm/giantswarm/issues/4242 for details.

fgimenez:
Tested on ginger.

PR #1170, opened 2018-09-20
v17: remove encrypter dependency from cloudformation resource

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/3783.

Successor of https://github.com/giantswarm/aws-operator/pull/1098.


kopiczko:
Oudated. I need to refresh my memory when I go back to it.

PR #1188, opened 2018-09-27
Bake k8scloudconfig into Docker image

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2816 (option 2 from here https://github.com/giantswarm/giantswarm/issues/2816#issuecomment-424759971)

Looks creepy as we bring all unused k8scloudconfig directories, but there is no other way to do this with `dep`, or I don't know any other way.


corest:
If we're ok with this, I'll provide same PRs for other operators.

Gopkg.toml
```diff
@@ -145,7 +145,10 @@ required = [
   name = "k8s.io/client-go"
   version = "kubernetes-1.10.4"
 
-
+[[prune.project]]
+  name = "github.com/giantswarm/k8scloudconfig"
+  unused-packages = false
```
kopiczko: I wonder if we can drop this line, i.e. `unused-packages = false`.

corest: Nope, I tried that. The only possible way to bypass that - add go file into each folder and import packages somewhere.

kopiczko: That makes sense. It should be better when we have release-operator in place anyway.

PR #1190, opened 2018-09-28
Fix v18 imports and bump version

fgimenez:


service/controller/v18/version_bundle.go
```diff
@@ -50,6 +50,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "aws-operator",
-		Version: "4.2.0",
+		Version: "4.3.0",
```
rossf7: For next PR v17 changelog entries should be removed.

fgimenez: Sure, done PTAL.

PR #1191, opened 2018-09-28
wire v18, bump version and initialize changelog

fgimenez:


rossf7:
The version needs to be wired in the resource sets.

https://github.com/giantswarm/aws-operator/blob/master/service/controller/cluster.go
https://github.com/giantswarm/aws-operator/blob/master/service/controller/drainer.go

PR #1193, opened 2018-09-28
Quick fix for m3 case of NVME timeouts unit

r7vme:
Fixes: https://github.com/giantswarm/giantswarm/issues/4290

The problem that unit will be in failed state on `m3` instances (no NVME), this does not cause any issues, just triggers alert about failed unit.

@tuommaki suggested following improvement to ignore errors in case if no nvme folder.

Shall we accept such exception and make this small improvement w/o new patch releases?


kopiczko:
I'm not sure here. Will you then make sure all currently installed versions have the change added manually?

r7vme:
Merging.

PR #1194, opened 2018-09-28
include memleak patch

fgimenez:

Using k8scc `v_3_6_1` (which includes the 1.11 mem leak fix) in both `v17` and `v18`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -312,7 +312,7 @@
     "v_3_6_0",
     "v_3_6_1"
   ]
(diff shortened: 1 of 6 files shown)


rossf7:
Can you check the changelog from the patch release.

https://github.com/giantswarm/aws-operator/blob/master/service/controller/v16patch1/version_bundle.go

Anything we need to add?

fgimenez:
@rossf7 Good point, I've added all the changelog entries from `v16patch1` PTAL.

PR #1198, opened 2018-10-05
Vendor e2e-harness and use time consts from backoff

indyfree:


indyfree:
Going ahead with this as ClusterState Test seems to be flapping and my change is unrelated. See e.g. here: https://circleci.com/gh/giantswarm/aws-operator/12310

PR #1202, opened 2018-10-06
Adds AMIs for all regions

JosephSalisbury:
Got image id for `us-east-1` from https://stable.release.core-os.net/amd64-usr/1745.4.0/coreos_production_ami_hvm.txt


JosephSalisbury:
@giantswarm/sig-operator decided to be a bit proactive here, and added all AMI IDs for all regions, so we don't have to deal with this again with the next region

RFR

JosephSalisbury:
Tested on `gauss`

PR #1203, opened 2018-10-06
Collect info from host routes

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/4333


pipo02mix:
It has changed with the new AZ implementation and not more worthy

PR #1221, opened 2018-10-25
circleci: allow approval before build

kopiczko:


.circleci/config.yml
```diff
@@ -151,85 +151,70 @@ workflows:
...
-          filters:
-            branches:
-              only: master
           requires:
-          - build
+          - master
       - e2eTestCurPRUpdate:
           requires:
-          - hold
+          - pr
```
kopiczko: This is

```yaml
workflows:
  version: 2
  build_e2e:
    jobs:
      - build
      - hold:
          type: approval
      - master:
          requires:
            - build
          filters:
            branches:
              only: master
      - pr:
          requires:
            - build
            - hold
          filters:
            branches:
              ignore: master



      - e2eTestCurMasterClusterState:
          requires:
            - master
      - e2eTestWIPMasterClusterState:
          requires:
            - master
      - e2eTestCurPRClusterState:
          requires:
            - pr
      - e2eTestWIPPRClusterState:
          requires:
            - pr



      - e2eTestCurMasterDraining:
          requires:
            - master
      - e2eTestWIPMasterDraining:
          requires:
            - master
      - e2eTestCurPRDraining:
          requires:
            - pr
      - e2eTestWIPPRDraining:
          requires:
            - pr



      - e2eTestCurMasterScaling:
          requires:
            - master
      - e2eTestWIPMasterScaling:
          requires:
            - master
      - e2eTestCurPRScaling:
          requires:
            - pr
      - e2eTestWIPPRScaling:
          requires:
            - pr



      - e2eTestCurMasterUpdate:
          requires:
          - master
      - e2eTestCurPRUpdate:
          requires:
          - pr
```

kopiczko: screen shot 2018-10-25 at 12 50 38

🎉

.circleci/config.yml
```diff
@@ -151,85 +151,70 @@ workflows:
...
+      - e2eTestCurPRScaling:
+          requires:
+            - pr
+      - e2eTestWIPPRScaling:
+          requires:
+            - pr
 
 
 
       - e2eTestCurMasterUpdate:
```
kopiczko: @xh3b4sd should we run these in WIP as well?

.circleci/config.yml
MarcelMue: You will need to define `pr` or `master` jobs here as well: https://github.com/giantswarm/aws-operator/pull/1221/files#diff-1d37e48f9ceff6d8030570cd36286a61R25

.circleci/config.yml
xh3b4sd: There should be a comment on the other definition that this makes no sense to be run in WIP. The test takes the current and wip version anyway in order to perform the upgrade. That is why the upgrade test is only there once and not twice.

kopiczko: To clarify. "There should be a comment" means the comment should be there but it disappeared or it should be added?

kopiczko:
I'm going with that.

.circleci/config.yml
kopiczko: Works also for master. As you can see hold is not green and tests are running.

screen shot 2018-10-25 at 17 30 07

PR #1233, opened 2018-10-31
vendor: update

kopiczko:


service/controller/v18/templates/cloudformation/guest/nat_gateway.go
```diff
@@ -4,6 +4,8 @@ const NatGateway = `{{define "nat_gateway"}}
   {{- $v := .Guest.NATGateway }}
   NATGateway:
     Type: AWS::EC2::NatGateway
+    DependsOn:
```
MarcelMue: Related to vendor?

kopiczko: No. I'm removing.

kopiczko:
Green enough for vendor update. 10/12 passing.

PR #1239, opened 2018-11-06
v18: add changelog about waiting for CF stacks to be deleted

kopiczko:


service/controller/v18/version_bundle.go
```diff
@@ -7,6 +7,11 @@ import (
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
+			{
+				Component:   "aws-operator",
+				Description: "Fixed a bug with removing CR before all Cloud Formation stacks are deleted.",
```
kopiczko: Just saying but I find it very weird to construct sentences like this. Also `fixed` is repeated from kind.
With imperative it would be `Wait for all Cloud Formation to be deleted before removing finalizers from the CR.`.

kopiczko:
```suggestion
				Description: "Fixed a bug with removing CR finalizers before all Cloud Formation stacks are deleted.",
```

service/controller/v18/version_bundle.go
```diff
@@ -7,6 +7,11 @@ import (
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
+			{
+				Component:   "aws-operator",
+				Description: "Fixed a bug with removing CR finalizers before all Cloud Formation stacks are deleted.",
+				Kind:        versionbundle.KindFixed,
+			},
```
kopiczko: I'm still waiting for the review here. Doesn't it make sense to add it to the changelog?

cc @teemow

PR #1247, opened 2018-11-07
update k8scloudconfig to 3.7.1

teemow:
this makes sure

* that the kube-proxy is installed before we wait for calico to be installed and ready.
* the pod priority class for calico is correct.


service/controller/v18patch1/version_bundle.go
```diff
@@ -8,33 +8,18 @@ func VersionBundle() versionbundle.Bundle {
...
 			{
-				Component:   "calico",
-				Description: "Updated to 3.2.3. Also the manifest has proper resource limits to get QoS policy guaranteed.",
-				Kind:        versionbundle.KindChanged,
-			},
-			{
-				Component:   "kubernetes",
-				Description: "Enabled admission plugins: DefaultTolerationSeconds, MutatingAdmissionWebhook, ValidatingAdmissionWebhook.",
+				Component:   "kube-proxy",
+				Description: "Now gets installed and upgraded before Calico.",
```
kopiczko: I don't know how to conform this message with the past tense BS.

teemow: :man_shrugging:

PR #1251, opened 2018-11-08
Wire v19

tuommaki:


tuommaki:
> The drainer wiring is missing.

Good catch! Fixed.

tuommaki:
E2e unstable. I take this as a flapping as the error message at least in some of them is same as mentioned in Slack discussions as well (`worker nodes are still not found`). Going with this.

PR #1253, opened 2018-11-08
v19/resource/migration: improve logs

kopiczko:
https://github.com/giantswarm/giantswarm/issues/4571

It looks like we update a CR that is newer than the one we observe in
a next reconciliation loop. These logs will allow to confirm that.

They also are aligned with our current standards.


kopiczko:
> I think the pointer magic is actually build to fail.

:D I think it's clear. You take the pointer, and return only error. It's like JSON marshaling.

> I find it also weird to have the other version now logged as well

I'm out of data to explain things. Maybe we can drop this from log when we figure actual cause. And in general `migration` resource should be removed at some point in the fuuuuture.

PR #1257, opened 2018-11-09
Improve legacy tenant subnet CIDR detection

tuommaki:
When checking AWSConfig CR if there's a tenant subnet already allocated
in legacy CIDR field, improve the check so that if it's a proper CIDR
then use it and otherwise proceed with new subnet allocation.

This way we can either use empty string or "DEPRECATED" etc. in legacy
CIDR field for migration period.

### DIFF
--- a/service/controller/v19/resource/ipam/create.go
+++ b/service/controller/v19/resource/ipam/create.go
@@ -39,15 +39,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		//
 		//     https://github.com/giantswarm/giantswarm/issues/4192
 		//
-		if key.CIDR(customObject) != "" {
+		_, presentSubnet, err := net.ParseCIDR(key.CIDR(customObject))
+		if err == nil {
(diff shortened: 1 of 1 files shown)


service/controller/v19/resource/ipam/create.go
```diff
@@ -39,15 +39,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		//
 		//     https://github.com/giantswarm/giantswarm/issues/4192
 		//
-		if key.CIDR(customObject) != "" {
```
kopiczko: I think we should check for `deprecated` string explicitly. CIDR in wrong format may be caused by other problems in the stack.

FYI this is how it was done before https://github.com/giantswarm/kubernetesd/commit/4d8d9efcbfd4b621c4eb916c343ed08fab0e6645#diff-6ca759f68647d79bdade44f011a47ff9R138

I don't mind it being in upper case though.

tuommaki:
Ok. Got it wrong. I'll redo this.

PR #1265, opened 2018-11-09
Adds cache

JosephSalisbury:
Towards reducing build times, this PR adds support for saving and restoring the go cache between jobs, to significantly reduce overall build times.

See https://circleci.com/docs/2.0/caching/ for CircleCI documentation.

First run (https://circleci.com/gh/giantswarm/aws-operator/14466)
- restoring cache: 00:00
- build: 18:33
- saving cache: 01:08
Total: 19:41 (1181 seconds)

Second run (https://circleci.com/gh/giantswarm/aws-operator/14467)
- restoring cache: 01:06
- build: 05:06
- saving cache: 00:00
Total: 06:12 (372 seconds)

So, ~70% decrease in total time.


.circleci/config.yml
```diff
@@ -32,11 +32,26 @@ jobs:
         chmod +x ./architect
         ./architect version
 
+    - run: |
+        date +"%Y" > /tmp/year
+        date +"%m" > /tmp/month
+        date +"%d" > /tmp/day
+    - restore_cache:
+        keys:
```
JosephSalisbury: Cache keys are prefix matched in order, so we attempt to use a go cache from today, this month, and then this year in order.
This does mean the first build of a day / month / year will be somewhat slower, but I expect it will still be faster overall.

.circleci/config.yml
```diff
@@ -32,11 +32,26 @@ jobs:
...
+        - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}-{{ checksum "/tmp/day" }}
```
teemow: Can you describe why you are restoring three keys?

kopiczko: +1

kopiczko: Also why checksums? What is wrong with `go-cache-v1-2018-11-12`? What is v1?

.circleci/config.yml
```diff
@@ -32,11 +32,26 @@ jobs:
...
+        - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}
+        - go-cache-v1-{{ checksum "/tmp/year" }}
+
     - run: ./architect build
```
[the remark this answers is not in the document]
JosephSalisbury: Architect already has support for mounting the go cache from the container to the host, this is the CircleCI configuration for persisting the cache, so we won't be able to add this to architect, no.

.circleci/config.yml
JosephSalisbury: From https://circleci.com/docs/2.0/caching/#using-keys-and-templates, you can't use the value of an environment variable as part of a key, but you can use the checksum of the contents of a file, hence the checksum to get a date format for the key.

I've added `v1` in case we need to update the key format in the future (recommended in docs).

JosephSalisbury: CircleCI uses prefix matching to retrieve the cache, and will try each key in order until it finds one that matches.

So, if we have a cache for the specific day of the build, we'll retrieve that. If we don't (e.g: first build of the day), we'll fallback to the cache from the start of the month (and so on for the first build of the year).

So, we are only restoring one cache, with fallbacks to less fresh caches in case of a miss.

.circleci/config.yml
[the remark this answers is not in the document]
teemow: ok. then we should add this to all go projects and at best add it to a guide how to setup new projects.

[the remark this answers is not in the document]
JosephSalisbury: Plan is to add a ticket to @giantswarm/sig-releng, wanted to see if it would work first :D

.circleci/config.yml
teemow: Why not just use one key which is independent of date?

JosephSalisbury: Caches are immutable, so we'd end up with a cache that is super out of date, and not get any benefit from it.

PR #1272, opened 2018-11-13
set etcd dns ttl to 60s

teemow:
to avoid etcd dns issues (with calico) during upgrades we have to lower
the ttl to at least 60s.


tuommaki:
One thing: Should this be also present in `v19`?

teemow:
@tuommaki thanks for the reminder. added to v19 too.

PR #1276, opened 2018-11-14
Add NAT GW for each availability zone

tuommaki:
In order to be able to route traffic when some AZs go down, each AZ must
have a NAT GW present.

### DIFF
--- a/service/controller/v19/templates/cloudformation/guest/nat_gateway.go
+++ b/service/controller/v19/templates/cloudformation/guest/nat_gateway.go
@@ -10,7 +10,7 @@ const NatGateway = `{{define "nat_gateway"}}
     Properties:
       AllocationId:
         Fn::GetAtt:
-        - NATEIP
+        - {{ .NATEIPName }}
         - AllocationId
       SubnetId: !Ref {{ .PublicSubnetName }}
       Tags:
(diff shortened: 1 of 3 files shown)


service/controller/v19/templates/cloudformation/guest/nat_gateway.go
```diff
@@ -11,19 +12,20 @@ const NatGateway = `{{define "nat_gateway"}}
         Fn::GetAtt:
         - NATEIP
         - AllocationId
-      SubnetId: !Ref PublicSubnet
+      SubnetId: !Ref {{ .PublicSubnetName }}
       Tags:
         - Key: Name
-          Value: {{ $v.ClusterID }}
-  NATEIP:
```
MarcelMue: Why did this line get removed? I don't quite get why we get the attribute a few lines above then.

tuommaki: `NATEIP` was a hard coded resource name that is now rendered from `{{ .NATEIPName }}`. In practice, resulting CF template will have `NATEIP00`, `NATEIP01` etc.

MarcelMue: Okay, then line 12 in this file is probably wrong?

tuommaki: Yes, good catch! Fixing

PR #1279, opened 2018-11-15
Remove quotes around AvailabilityZones

tuommaki:
.Values.Installation.V1.Provider.AWS.AvailabilityZones is a list so when
there are quotes around it, its internal [ and ] get interpreted as part
of a string value -> remove them.


tuommaki:
Wrong approach, doesn't work.

PR #1280, opened 2018-11-15
Expand AZ list to configmap

tuommaki:


xh3b4sd:
BTW when I say "no strong feelings" I want to say that I have a lot of strong feelings, but they are all awful. :D

tuommaki:
Scream

PR #1281, opened 2018-11-15
integration: fix certs deletion during tenant cluster deletion

kopiczko:
This fixes recreatecluster tests.

### DIFF
--- a/integration/setup/tenant_cluster.go
+++ b/integration/setup/tenant_cluster.go
@@ -25,26 +25,23 @@ const (
 	crNamespace = "default"
 )
 
-func EnsureTenantClusterCreated(ctx context.Context, id string, config Config) error {
-	err := InstallAWSConfig(ctx, id, config)
+func EnsureTenantClusterCreated(ctx context.Context, id string, config Config, wait bool) error {
+	config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating tenant cluster %#q", id))
+
+	err := ensureAWSConfigInstalled(ctx, id, config)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	err = InstallCertConfigs(ctx, id, config)
+	err = ensureCertConfigsInstalled(ctx, id, config)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	{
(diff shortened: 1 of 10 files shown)


integration/setup/tenant_cluster.go
```diff
@@ -192,45 +236,7 @@ func ensureHostVPC(ctx context.Context, config Config) (string, error) {
...
-			},
-		}
-
-		values, err = chartvalues.NewAPIExtensionsAWSConfigE2E(c)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-	}
-
-	err = config.Release.Install(ctx, key.AWSConfigReleaseName(id), release.NewStableChartInfo("apiextensions-aws-config-e2e-chart"), values, crExistsCondition(ctx, id, config))
```
kopiczko: This went up. The only change is `config.Release.Install` replaced with `config.Release.EnsureInstalled`.

integration/test/ipam/provider.go
```diff
@@ -57,52 +57,36 @@ func NewProvider(config ProviderConfig) (*Provider, error) {
...
 	}
 
-	err := setup.InstallAWSConfig(ctx, id, setupConfig)
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	err = setup.InstallCertConfigs(ctx, id, setupConfig)
+	wait := false
+	err := setup.EnsureTenantClusterCreated(ctx, id, setupConfig, wait)
```
kopiczko: Now with `wait` we can reuse those which is cleaner.

integration/test/recreatecluster/recreate_cluster_test.go
```diff
@@ -17,24 +17,18 @@ func Test_Recreate_Cluster(t *testing.T) {
...
-
-		config.Logger.LogCtx(ctx, "level", "debug", "message", "deleted tenant cluster")
 	}
 
 	{
-		config.Logger.LogCtx(ctx, "level", "debug", "message", "creating tenant cluster")
-
-		err := setup.EnsureTenantClusterCreated(ctx, env.ClusterID(), config)
+		wait := true
+		err := setup.EnsureTenantClusterCreated(ctx, env.ClusterID(), config, wait)
```
kopiczko: This call was failing because it tried to reinstall existing certs.

PR #1283, opened 2018-11-15
circleci: disable ClusterState tests

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4692

They seem to be failing constantly with `worker nodes are still not found`.
They are disabled temporary to not annoy people when they are being fixed.


kopiczko:
Everything green on the first run :)

PR #1288, opened 2018-11-16
terminate the master after detaching of volumes

teemow:
once the volumes are detached from the master we can terminate the
instance.

the instance would be terminate by the cloudformation too, but only at the
very end. and since k8s keeps the lost master node around as long as it is
only "Stopped" we need to manually terminate the master.


service/controller/v19/resource/cloudformation/update.go
```diff
@@ -50,14 +50,27 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 			force := false
 			shutdown := true
 			wait := true
+			instance := ""
+
 			for _, v := range volumes {
 				for _, a := range v.Attachments {
+					instance = a.InstanceID
```
kopiczko: I think we have docker volumes also in worker instances so this is just a random instance. Am I missing something?

xh3b4sd: I also think we should ensure to get the instance ID of the master instance we want. Only taking any is not a sound approach, especially when there is such an implementation detail and later something changes and everything breaks and nobody knows why.

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -50,14 +50,27 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
...
+
+			i := &ec2.TerminateInstancesInput{
+				InstanceIds: []*string{
+					aws.String(instance),
+				},
+			}
+			_, err = sc.AWSClient.EC2.TerminateInstances(i)
+			if err != nil {
+				return microerror.Mask(err)
+			}
```
xh3b4sd: Initializing the termination here makes sense. I never liked the EBS service implementation manipulating the actual instance state. For the style guide it would be awesome to scope it in its own `{ ... }` block.

service/controller/v19/resource/cloudformation/update.go
teemow: The search is for instances with docker AND etcd volume. And this is only the master. I didn't touch it and yes I know getting the instanceId is ugly here. All in all I found this code really implicit (master shutdown being 'hidden' in volume detachment) and would like to change this at some point to have proper and explicit handling of the master instance.

teemow: To be clear I don't think that I am making it worse here.

service/controller/v19/resource/cloudformation/update.go
teemow: Why?

xh3b4sd: Because the team writes code like that in a lot of places. With the short and simplistic variable names like `i` for input and `o` for output when working with the AWS SDK it is always the same easy way when scoped. You do not need to change the code when it is extended. You do not need to think about other weird names and make the code different everywhere and thus harder to scan.

teemow: > Because the team writes code like that in a lot of places. 

Thats not an argument :smile: 

I understood you would like to scope the local variables. Makes sense. How do you distinguish between function and scope then? When to use what? (naturally I'd rather make it a function then scoping this inside the function)

xh3b4sd: When you want to have a dedicated function go ahead. Otherwise I would appreciate the scope. There is no "rule" to prefer one over the other.

teemow: I've refactored this into a dedicated method and also made sure to only terminate one stopped master there. Still not the most beautiful thing. PTAL

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+			},
+		}
+
+		result, err := sc.AWSClient.EC2.DescribeInstances(i)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if len(result.Reservations) > 1 {
+			return microerror.Maskf(moreThanOneOldMasterError, "Expected one stopped old master. Found %d stopped masters with the name %s.", len(result.Reservations), key.MasterInstanceName)
```
kopiczko:
```suggestion
			return microerror.Maskf(executionFailedError, "expected at most one master instance with name %#q in state %#q but got %d", instanceName, instanceState, len(result.Reservations))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		}
+
+		if len(result.Reservations) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("No old master found with state 'stopped' and name %s", key.MasterInstanceName(customObject)))
```
kopiczko:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find master instance with name %#q in state %#q", instanceName, instanceState))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+			return nil
+		}
+
+		if len(result.Reservations[0].Instances) > 1 {
+			return microerror.Maskf(moreThanOneOldMasterError, "Expected one stopped old master. Found %d stopped masters with the name %s.", len(result.Reservations), key.MasterInstanceName)
```
kopiczko:
```suggestion
			return microerror.Maskf(executionFailedError, "expected at most one master instance with name %#q in state %#q but got %d", instanceName, instanceState, len(result.Reservations))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		}
+
+		if len(result.Reservations[0].Instances) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("No old master found with state 'stopped' and name %s", key.MasterInstanceName(customObject)))
```
kopiczko:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("did not find master instance with state `stopped` and name %#q", key.MasterInstanceName(customObject)))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+
+// Terminates the master instance of the cluster.
+//
+// To detect the old master instance we find the instance by its name and
+// the instance state "stopped". Within the upgrade process the master first
+// gets stopped and its volumes get detached. This function makes sure that
+// the stopped instance is also terminated.
+func (r *Resource) terminateOldMasterInstance(ctx context.Context, obj interface{}) error {
+	var result ec2.DescribeInstancesOutput
+
```
kopiczko:
```suggestion
	instanceName := key.MasterInstanceName(customObject)
	instanceState := "stopped"

```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		return microerror.Mask(err)
+	}
+
+	{
+		i := &ec2.DescribeInstancesInput{
+			Filters: []*ec2.Filter{
+				{
+					Name: aws.String("tag:Name"),
+					Values: []*string{
+						aws.String(key.MasterInstanceName(customObject)),
```
kopiczko:
```suggestion
						aws.String(instanceName),
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+					},
+				},
+				{
+					Name: aws.String("instance-state-name"),
+					Values: []*string{
+						aws.String("stopped"),
```
kopiczko:
```suggestion
						aws.String(instanceState),
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		}
+
+		if len(result.Reservations[0].Instances) > 1 {
+			return microerror.Maskf(moreThanOneOldMasterError, "Expected one stopped old master. Found %d stopped masters with the name %s.", len(result.Reservations), key.MasterInstanceName)
+		}
+
+		if len(result.Reservations[0].Instances) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("No old master found with state 'stopped' and name %s", key.MasterInstanceName(customObject)))
+			return nil
+		}
```
kopiczko:
```suggestion
		}

		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found master instance with name %#q in state %#q", instanceName, instanceState))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		return microerror.Mask(err)
+	}
+
+	sc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	{
+		i := &ec2.DescribeInstancesInput{
```
kopiczko:
```suggestion
		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding master instance with name %#q in state %#q", instanceName, instanceState))

		i := &ec2.DescribeInstancesInput{
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+		}
+
+		if len(result.Reservations[0].Instances) < 1 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("No old master found with state 'stopped' and name %s", key.MasterInstanceName(customObject)))
+			return nil
+		}
+	}
+
+	{
+		i := &ec2.TerminateInstancesInput{
```
kopiczko:
```suggestion
		instanceID := *result.Reservations[0].Instances[0].InstanceId

		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("terminating master instance with ID %#q, instanceID))

		i := &ec2.TerminateInstancesInput{
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+			InstanceIds: []*string{
+				aws.String(*result.Reservations[0].Instances[0].InstanceId),
```
kopiczko:
```suggestion
				aws.String(instanceID),
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -276,3 +287,85 @@ func shouldUpdate(currentState, desiredState StackState) bool {
...
+			},
+		}
+		_, err := sc.AWSClient.EC2.TerminateInstances(i)
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
kopiczko:
```suggestion
		}

		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("terminated master instance with ID %#q, instanceID))
```

service/controller/v19/resource/cloudformation/update.go
```diff
@@ -35,29 +36,39 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 		}
 
 		if stackStateToUpdate.ShouldUpdate && !stackStateToUpdate.ShouldScale {
```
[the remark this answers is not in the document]
teemow: This is a good point and should be discussed in https://github.com/giantswarm/giantswarm/issues/4543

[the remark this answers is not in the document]
pipo02mix: Ideally we would reconcile the number of nodes before the upgrade. Not sure how to do it. In the ticket pointed by Timo I said until we have autoscaler integrated SE should first align number the nodes and after perform the upgrade

PR #1301, opened 2018-11-22
Fix race condition when checking allocated subnets from CRs

tuommaki:
When pre-v19 and v19+ clusters are created within short period of time
and AWSConfig for v19+ cluster gets reconciliation before pre-v19 one,
there was a race that missing Status field for pre-v19 CR lead to
overlapping allocation of tenant subnet.

When gathering present subnet allocations from AWSConfig CRs, use legacy
CIDR field from Spec section when one in Status section is empty (or
missing).

### DIFF
--- a/service/controller/v19/resource/ipam/create.go
+++ b/service/controller/v19/resource/ipam/create.go
@@ -204,7 +204,13 @@ func getAWSConfigSubnets(g8sClient versioned.Interface) ([]net.IPNet, error) {
 	for _, ac := range awsConfigList.Items {
 		cidr := key.ClusterNetworkCIDR(ac)
 		if cidr == "" {
-			continue
+			// To prevent race condition when pre-v19 and v19+ clusters are
+			// created within short period of time and v19+ CR gets picked
+			// first. The pre-v19 CR might not have Status section yet.
+			cidr = key.CIDR(ac)
+			if cidr == "" {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


tuommaki:
When the CRD gets changed (and legacy field removed) the key function used here `key.CIDR` doesn't compile anymore and to fix that problem you would find this piece of code. Do you think this needs TODO despite of that?

service/controller/v19/resource/ipam/create.go
```diff
@@ -204,7 +204,13 @@ func getAWSConfigSubnets(g8sClient versioned.Interface) ([]net.IPNet, error) {
 	for _, ac := range awsConfigList.Items {
 		cidr := key.ClusterNetworkCIDR(ac)
 		if cidr == "" {
-			continue
+			// To prevent race condition when pre-v19 and v19+ clusters are
+			// created within short period of time and v19+ CR gets picked
+			// first. The pre-v19 CR might not have Status section yet.
+			cidr = key.CIDR(ac)
```
MarcelMue: Is this the value from `Spec` or from `Status`? I assume `Spec`?

tuommaki: `Spec`

MarcelMue: Ack all good then.

PR #1303, opened 2018-11-22
increase-master-root-volume-size

calvix:
towards https://github.com/giantswarm/giantswarm/issues/4620


calvix:
> LGTM if this is an ephemeral volume.

Yep, this is ephemeral volume and the CF update went smoothly. No issues left or right.

calvix:
`/var/log` will be moved to separate EBS that will be also encrypted, there is a story coming from VF

PR #1306, opened 2018-11-23
change default storage class for multiaz

teemow:
make sure that pods and volumes get scheduled into the same az.

activate the volume binding mode 'WaitForFirstConsumer' to the storage
class which will delay the binding and provisioning of a
PersistentVolume until a Pod using the PersistentVolumeClaim is created.


pipo02mix:
I would like to test before what happens to the existing PVC when you change the storage class properties.

PR #1307, opened 2018-11-23
fix security group for multiple nat gateways

teemow:
in case we have multiple nat gateways we need to add multiple security
group rules. this fixes the templating and removes old adapter code.


teemow:
@calvix awesome thanks! that explains that the control plane couldn't access the tenant api anymore :heart:

PR #1308, opened 2018-11-23
Reduce number of AWS API calls in ELB metrics collector

tuommaki:
Instead of getting tags for each ELB separately group all ELB names into
one elb.DescribeTags query and work from there on.

ELB instance health stauses are still fetched with N requests because
AWS SDK doesn't allow specification of ELB names in
elb.DescribeInstanceHealthInput.

Towards https://github.com/giantswarm/giantswarm/issues/4759


service/collector/elb.go
```diff
@@ -52,6 +52,12 @@ type ELB struct {
 	installationName string
 }
 
+type loadBalancer struct {
+	instancesOutOfService float64
+	name                  string
+	tags                  map[string]string
```
kopiczko: I personally prefer to keep structs "ready to be exported". In this case all The fields should be capitalized. Then it's easier to reason about the struct within package.

tuommaki: Changed.

tuommaki:
Tested on `ginger` and this seems to work as expected.

tuommaki:
All was green here. I just merged another PR so not gonna re-run tests. Going with this.

PR #1317, opened 2018-11-27
circleci: restore ClusterState tests

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4692.

It should be fixed with
https://github.com/giantswarm/aws-operator/pull/1288.


kopiczko:
All green on first run. That makes me happy :)

PR #1320, opened 2018-11-27
Vendor update e2e-harness

tuommaki:


tuommaki:
Requires also other dependences to be updated. This is taken care in here: https://github.com/giantswarm/aws-operator/pull/1318

PR #1323, opened 2018-11-29
Only take etcd port from key

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/2383


corest:
@MarcelMue is this still a valid thing? Seems to be outdated

MarcelMue:
The general change is still valid and on my personal board. I didn't get this change out because I got stuck in azure. Going to close here and try again in the future.

PR #1327, opened 2018-12-03
Add policy in workers to allow modify and list hostzones

pipo02mix:


pipo02mix:
> Direction looks good. For my interest is this prep so we can use external-dns with AWS tenant clusters?

Yeah [I created a ticket](https://github.com/giantswarm/giantswarm/issues/4795) after discussion with Puja to plan some changes at IC level. We have thought to move the logic to create the ELB and DNS out of the aws operator as we do in Azure. So `external-dns` will take care of creating the DNS recordsets and nginx controller service (with the annotation) will provision the ELB.

As there is a customer that already want to use external DNS I thought to make this addition to contribute with the epic.

PR #1342, opened 2018-12-13
update autoscaler iam policy to scale to 0

teemow:
the autoscaler needs additional permissions to scale an ASG to 0 nodes.

see
https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws#scaling-a-node-group-to-0


tuommaki:
What would be the scenario where tenant cluster workers would be scaled to 0? Something useful for customers who are otherwise running with less than three workers?

teemow:
I guess there are not many use cases in clusters at the moment since there is only one node pool at the moment. But it is supported upstream and probably there have been use-cases why this was implemented. Eg nodes running batch jobs.

PR #1345, opened 2018-12-14
fix autoscaler iam policy condition

teemow:
use `autoscaling:ResourceTag` instead of `aws:RequestTag` and the scope
of the policy is correctly mapped to ASGs of the cluster itself.


teemow:
Tested on a cluster and this one works correctly.

PR #1346, opened 2018-12-17
Describe ELB tags in batches (for metrics)

tuommaki:
AWS API has a limit for number of ELBs to process in the scope of one
DescribeTags request. In order to efficiently process over ELBs when
there are many of them, perform DescribeTags requests in concurrent
batches.


tuommaki:
Testing on `gauss` with multiple clusters and this seems to work now:
screenshot from 2018-12-17 13-02-02

service/collector/elb.go
```diff
@@ -139,31 +143,75 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
 	{
-		i := &elb.DescribeTagsInput{}
-		for _, l := range loadBalancers {
-			i.LoadBalancerNames = append(i.LoadBalancerNames, l.LoadBalancerName)
+		// AWS API has a limit for maximum number of LoadBalancerNames in
+		// single Describe request so it must be done in batches of
+		// maxELBsInOneBatch. In order to not spend so much time on this,
+		// perform requests concurrently and synchronize them with errgroup.
+		errGroup, _ := errgroup.WithContext(context.Background())
+		tagOutputs := make([]*elb.DescribeTagsOutput, len(loadBalancers)/maxELBsInOneDescribeTagsBatch+1)
```
JosephSalisbury: A comment for the math here would be helpful, I don't follow the why

MarcelMue: Comment could be helpful but it's just calculating the number of output badges we expect.

TheoBrigitte: Example: when you have 42 load balancers, and you process them by batches of 20. You'll have 3 batches.

Hence:
```go
a := len(loadBalancers)            // 42
b := maxELBsInOneDescribeTagsBatch // 20
c := a/b+1                         // 3
```

tuommaki:
For the record: CI green, E2E green, manual testing on `geckon` green. I'll add one more comment for explanation though. :+1:

service/collector/elb.go
tuommaki: Exactly. Added also comment telling the same. :slightly_smiling_face:

service/collector/elb.go
```diff
@@ -139,31 +143,77 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
 	var lbs []loadBalancer
 	{
-		i := &elb.DescribeTagsInput{}
-		for _, l := range loadBalancers {
-			i.LoadBalancerNames = append(i.LoadBalancerNames, l.LoadBalancerName)
+		// AWS API has a limit for maximum number of LoadBalancerNames in
+		// single Describe request so it must be done in batches of
+		// maxELBsInOneBatch. In order to not spend so much time on this,
+		// perform requests concurrently and synchronize them with errgroup.
+		errGroup, _ := errgroup.WithContext(context.Background())
```
kopiczko:
```suggestion
		errGroup, _ := errgroup.WithContext(context.TODO())
```

Should be taken care of in https://github.com/giantswarm/giantswarm/issues/4909

service/collector/elb.go
kopiczko: That's the best microoptimization I've seen in a while :D

service/collector/elb.go
```diff
@@ -139,31 +143,77 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
+		// Allocate list for ELB tag description results. Length is number of
+		// AWS API batches + 1 for possible left over from uneven number.
+		tagOutputs := make([]*elb.DescribeTagsOutput, len(loadBalancers)/maxELBsInOneDescribeTagsBatch+1)
+
+		i := 0
+		tagInput := &elb.DescribeTagsInput{}
+		for _, lb := range loadBalancers {
+			if len(tagInput.LoadBalancerNames) == maxELBsInOneDescribeTagsBatch {
+				copyTagInput := tagInput
```
kopiczko: AFAIU only a pointer to the same struct is copied here. I guess it wasn't intended.

service/collector/elb.go
```diff
@@ -139,31 +143,77 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
+				errGroup.Go(func() error {
```
kopiczko: `i` should be copied here. I guess `DescribeTags` will take so much time that all of the outputs will land in the last index.

service/collector/elb.go
```diff
@@ -139,31 +143,77 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
+
+				tagInput = &elb.DescribeTagsInput{}
+				i++
+			}
+
+			tagInput.LoadBalancerNames = append(tagInput.LoadBalancerNames, lb.LoadBalancerName)
+		}
+
+		// Last batch.
+		if len(tagInput.LoadBalancerNames) > 0 {
```
kopiczko: This is ugly and I believe it can be removed.

service/collector/elb.go
```diff
@@ -120,16 +125,18 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
 		return microerror.Mask(err)
 	}
 
-	var loadBalancers []*elb.LoadBalancerDescription
+	var loadBalancerNames []*string
```
tuommaki: Only `LoadBalancerName` is used so to make further logic in this function nicer, extract name out here.

service/collector/elb.go
kopiczko: Ack. It's no longer necessary when using `append`.

service/collector/elb.go
```diff
@@ -139,31 +146,70 @@ func (e *ELB) collectForAccount(ch chan<- prometheus.Metric, awsClients clientaw
...
 		}
 
-		for _, d := range o.TagDescriptions {
-			lb := loadBalancer{
-				Name: *d.LoadBalancerName,
-				Tags: make(map[string]string),
+		// Extract tags from responses and create further loadBalancer types
+		// based on these.
+		for _, o := range tagOutputs {
+			if o == nil {
```
kopiczko: I think when we have sucha condition this means a bug in the code above. I'd prefer to have a panic and fix it.

tuommaki: Oh, true. With earlier implementation this was necessary due to different cases whether `numLBs % batchSize == 0` was true or not. I'll remove the check so it will panic in case it behaves wrong.

tuommaki:
Tested the latest version successfully on `ginger`. E2E has been green at the times. There are currently couple flapping errors, but those won't test metrics gathering code anyway so I proceed here.

PR #1353, opened 2019-01-10
add-vpc-s3-endpoint

calvix:
add VPC endpoint for S3, but only for s3 bucket from the guest cluster account

### DIFF
--- a/service/controller/v21/templates/cloudformation/guest/vpc.go
+++ b/service/controller/v21/templates/cloudformation/guest/vpc.go
@@ -23,4 +23,13 @@ const VPC = `{{define "vpc"}}
       Tags:
         - Key: Name
           Value: {{ $v.ClusterID }}
+  VPCS3Endpoint:
+    Type: 'AWS::EC2::VPCEndpoint',
+    Properties:
+      VpcId: !Ref VPC
+      RouteTableIds:
+        {{- range $v.RouteTableNames }}
+        - !Ref .ResourceName
+        {{end}}
+      Policy: !Ref VPCS3EndpointPolicy
 {{end}}`
--- a/service/controller/v21/adapter/guest_iam_policies.go
+++ b/service/controller/v21/adapter/guest_iam_policies.go
@@ -14,6 +14,7 @@ import (
 type GuestIAMPoliciesAdapter struct {
 	ClusterID         string
 	EC2ServiceDomain  string
+	GuestAccountID    string
 	KMSKeyARN         string
 	MasterRoleName    string
 	MasterPolicyName  string
@@ -30,6 +31,7 @@ func (i *GuestIAMPoliciesAdapter) Adapt(cfg Config) error {
 
 	i.ClusterID = clusterID
 	i.EC2ServiceDomain = key.EC2ServiceDomain(cfg.CustomObject)
+	i.GuestAccountID = cfg.GuestAccountID
 	i.MasterPolicyName = key.PolicyName(cfg.CustomObject, key.KindMaster)
 	i.MasterProfileName = key.InstanceProfileName(cfg.CustomObject, key.KindMaster)
 	i.MasterRoleName = key.RoleName(cfg.CustomObject, key.KindMaster)
--- a/service/controller/v21/adapter/guest_vpc.go
+++ b/service/controller/v21/adapter/guest_vpc.go
@@ -16,6 +16,7 @@ type GuestVPCAdapter struct {
 	HostAccountID    string
 	PeerVPCID        string
 	PeerRoleArn      string
+	RouteTableNames  []RouteTableName
 }
 
 func (v *GuestVPCAdapter) Adapt(cfg Config) error {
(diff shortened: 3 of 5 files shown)


mazzy89:
I went through the `adapter` codebase to grasp some knowledge about it.

The last thing that is required to do as long as I understood would be to retrieve and pass through `GuestAccountID` here https://github.com/giantswarm/aws-operator/blob/78020856b0202ce0225a4b6869e9eb5c000b6e15/service/controller/v21/resource/cloudformation/main_stack.go#L43. 

Correct?

Q. Which are the concepts here of `Host` and `Guest`?

calvix:
Yes, 
> Q.: Which are the concepts here of Host and Guest?

This our old naming for `Control Plane` (Host) and `Tenant Cluster` (Guest)

service/controller/v21/templates/cloudformation/guest/vpc.go
```diff
@@ -23,4 +23,26 @@ const VPC = `{{define "vpc"}}
       Tags:
         - Key: Name
           Value: {{ $v.ClusterID }}
```
kopiczko: I completely miss why we need this endpoint. I'd appreciate some more information in the PR desc <3

calvix: https://github.com/giantswarm/giantswarm/issues/4927

customer requested this feature, it allows access s3 without going over public network

service/controller/v21/templates/cloudformation/guest/vpc.go
```diff
@@ -23,4 +23,26 @@ const VPC = `{{define "vpc"}}
...
+      VpcId: !Ref VPC
+      RouteTableIds:
+        {{- range $v.RouteTableNames }}
+        - !Ref {{ .ResourceName }}
+        {{- end}}
+      ServiceName: 'com.amazonaws.{{ $v.Region }}.s3'
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Sid: "{{ $v.ClusterID }}-vpc-s3-endpoint-policy-bucket"
```
kopiczko: What is `Sid`? I don't see it in examples. Some arbitrary name?

mazzy89: `Sid` is an unique ID  that you apply to the policy to be uniquely identified and displayed in the UI. it is simply useful for displaying. It can be left blank. It is not mandatory

calvix: Its inline policy name, its not required, but it helps explain what is that policy for

service/controller/v21/templates/cloudformation/guest/vpc.go
```diff
@@ -23,4 +23,26 @@ const VPC = `{{define "vpc"}}
...
+            Principal : "*"
+            Effect: "Allow"
+            Action: "*"
+            Resource: "arn:{{ $v.RegionARN }}:s3::{{ $v.GuestAccountID }}:*"
+          - Sid: "{{ $v.ClusterID }}-vpc-s3-endpoint-policy-object"
```
mazzy89: The current policy seems not correct because according the aws arn guideline it should not contain the `GuestAccountID` as showed here https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

A good policy could be this:

```yaml
PolicyDocument:
        Version: "2012-10-17"
        Statement:
          - Sid: "{{ $v.ClusterID }}-vpc-s3-endpoint-policy-bucket"
            Principal: "*"
            Effect: "Allow"
            Action: "s3:*"
            Resource:
              - "arn:{{ $v.RegionARN }}:s3:::*"
              - "arn:{{ $v.RegionARN }}:s3:::*/*"
```

calvix: this is a problem

mazzy89: Actually thinking for 2 more secs this Condition doesn't cover all the cases :-(

calvix:
So in the end we wont need guestAccountID as for some reason S3 ARN policy does not allow use of accountID ( one of 2 service which don't allow it)

We need to have it like that.

service/controller/v21/templates/cloudformation/guest/vpc.go
```diff
@@ -23,4 +23,26 @@ const VPC = `{{define "vpc"}}
...
+          - Sid: "{{ $v.ClusterID }}-vpc-s3-endpoint-policy-bucket"
+            Principal : "*"
+            Effect: "Allow"
+            Action: "s3:*"
+            Resource: "arn:{{ $v.RegionARN }}:s3:::*"
+          - Sid: "{{ $v.ClusterID }}-vpc-s3-endpoint-policy-object"
+            Principal : "*"
+            Effect: "Allow"
+            Action: "s3:*"
+            Resource: "arn:{{ $v.RegionARN }}:s3:::*/*"
```
kopiczko: What is the actual difference between those two statements besides `Sid`?

mazzy89: One applies on buckets, the other one applies to objects inside the buckets.

See for instance actions like `s3:GetObject`. This must apply to objects of the bucket. While action like `s3:ListBucket` applies to the  bucket itself

PR #1354, opened 2019-01-11
Vendor update apiextensions & statusresource

tuommaki:
Update apiextensions and statusresource in order to get k8s labels for nodes
listed in CR status.


tuommaki:
Rolling with this. Seems to be already present in Azure & KVM.

tuommaki:
Oh but wait, it was already there. Nevermind... I'll wake up now.

PR #1357, opened 2019-01-15
Lock default security group

mazzy89:
This PR locks down the default Security Group created once created the VPC. By default it is blank which means every connection is allowed. As stated here https://github.com/giantswarm/giantswarm/issues/4671 it is not assigned to anything but it is good practice to lock it down.

The PR must be still tested but it would be nice to have the first round of reviews on my first PR 😇

### DIFF
--- a/service/controller/v21/templates/cloudformation/guest/security_groups.go
+++ b/service/controller/v21/templates/cloudformation/guest/security_groups.go
@@ -143,22 +143,4 @@ const SecurityGroups = `{{define "security_groups" }}
       ToPort: -1
       SourceSecurityGroupId: !Ref MasterSecurityGroup
 
-  VPCDefaultSecurityGroupIngress:
-    Type: AWS::EC2::SecurityGroupIngress
-    Properties:
-      GroupId !GetAtt VPC.DefaultSecurityGroup
-      Description: Lockdown Default Security Group ingress traffic.
-      IpProtocol: -1
-      FromPort: -1
-      ToPort: -1
-      SourceSecurityGroupId: !Ref VPC.DefaultSecurityGroup
-
-  VPCDefaultSecurityGroupEgress:
-    Type: AWS::EC2::SecurityGroupEgress
-    Properties:
-      GroupId !GetAtt VPC.DefaultSecurityGroup
-      Description: Lockdown Default Security Group egress traffic.
-      IpProtocol: -1
-      CidrIp: 127.0.0.1/32
-
 {{ end }}`
--- a/service/controller/v21/templates/cloudformation/guest/security_groups.go
+++ b/service/controller/v21/templates/cloudformation/guest/security_groups.go
@@ -143,4 +143,22 @@ const SecurityGroups = `{{define "security_groups" }}
       ToPort: -1
       SourceSecurityGroupId: !Ref MasterSecurityGroup
 
+  VPCDefaultSecurityGroupIngress:
+    Type: AWS::EC2::SecurityGroupIngress
+    Properties:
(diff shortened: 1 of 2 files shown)


kopiczko:
Please also run e2e test. They need manual approval. See the check `ci/circleci: build_e2e/hold Pending — Your job is on hold on CircleCI!` and click on `Details`.

service/controller/v21/templates/cloudformation/guest/security_groups.go
```diff
@@ -143,4 +143,22 @@ const SecurityGroups = `{{define "security_groups" }}
...
       SourceSecurityGroupId: !Ref MasterSecurityGroup
 
+  VPCDefaultSecurityGroupIngress:
+    Type: AWS::EC2::SecurityGroupIngress
+    Properties:
+      GroupId !GetAtt VPC.DefaultSecurityGroup
+      IpProtocol: -1
+      FromPort: -1
+      ToPort: -1
+      SourceSecurityGroupId: !Ref VPC.DefaultSecurityGroup
```
calvix: I would definitely add `CidrSource: "127.0.0.1/32"`, same for the egress.
Also adding `Description` would be beneficial as it will also tell customers a reason why this is locked. 
`Description: "Lockdown default security Group."`

mazzy89: After some research it makes sense to add the loopback addr for the egress rule as stated by @calvix  . Regarding the ingress rule I'd say is ok to work like that. Please advice if you don't agree with that.

Reference:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#w2ab1c21c10d105d104c15b4

calvix: Its indeed redundant, yet it's much more easy to read :)

also `!Ref VPC.DefaultSecurityGroup` this might not work as there is no such resource named `VPC.DefaultSecurityGroup` in CloudFormation template so Reference that might not work.

mazzy89: `VPC.DefaultSecurityGroup` it should work. `VPC` is the name of block which reference the VPC object. `DefaultSecurityGroup` is a special property name in CloudFormation as here described https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html#w2ab1c21c10d105d174c11b4

calvix: Oh special property, Nice.

mazzy89: I agree with the readability however I don't suggest to use the loopback address for the ingress rule because for the fact that in case that two or more EC2 instances have assigned the default security group then they can't communicate anymore between them. This security ingress rule instead is a little bit less restricted for instances inside the VPC but at the same time very safe for whatever thing which run outside.

mazzy89:
The PR has been tested and it works. The only mismatch I could have seen is that in the Ingress Rule the description is not reported as showed in the screenshots below. Not sure about the reason. However in my experience I'm not new to these kind of mismatches with CloudFormation. Any of you guys have any idea?

image
image
image
image

tuommaki:
>  Any of you guys have any idea?

No clue. Looks weird, but could it be a bug due to self-pointing security group? If it otherwise works correctly, I guess this is just an aesthetic issue :man_shrugging:

mazzy89:
By removing the Ingress Rule the CloudTrail err is gone. I think we could live without description in the default Ingress Rule. Going to run the last e2e tests and if there are no objections I'm going to merge this.

PR #1369, opened 2019-01-28
Fix encryption certificates

mazzy89:
This PR aims to resolve the issue https://github.com/giantswarm/giantswarm/issues/5179

During cluster creation an error is reported that both master and worker templates are invalid due to unrecognised characters.

The issue is that when the certificates are encoded in base64 with the function `RenderFileAssetContent` then the result is not valid.

The solution resides on encoding in base64 directly rather than passing the asset content to the function.

The questions we could try to reply to make things clear are:

1. Why this issue now? Why has not been found out before? Any regression in any library or any changes?
2. Why the function `RenderFileAssetContent`? The creation of the template invalidates in case of certs breaks the the certs?

cc @corest

### DIFF
--- a/service/controller/v22/cloudconfig/master_template.go
+++ b/service/controller/v22/cloudconfig/master_template.go
@@ -2,6 +2,7 @@ package cloudconfig
 
 import (
 	"context"
+	"encoding/base64"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/certs"
@@ -200,6 +201,7 @@ func (e *MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
 		},
 	}
 
+	certsMeta := []k8scloudconfig.FileMetadata{}
 	{
 		certFiles := certs.NewFilesClusterMaster(e.ClusterCerts)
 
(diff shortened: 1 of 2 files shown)


corest:
Pls fix `fmt`. 
The issue is, that encrypted certificate is not a valid go template. But it shouldn't even go through templating as it is for sure doesn't require any

mazzy89:
@corest the `fmt` is already fixed. VSCode has `fmt` included.

> The issue is, that encrypted certificate is not a valid go template. But it shouldn't even go through templating as it is for sure doesn't require any.

Then it makes sense. Even tough I'd expect the template function to be a bit more robust and validate the sanity of the template at the level.

mazzy89:
Yeah good point @corest. Should I make a copy of the v22 folder rename it v22patch and then put the changes there right?

corest:
@mazzy89 no, `v22` is not a part of any active release. But `v21` is. So we will need this in `v22` as well as in `v21patch1`. Or maybe we can even patch it in place.   I think you can ask your buddy about the current release process. Otherwise, I can explain it to you in the call. But all release magic requires time, so you can discuss priorities and process with your team

mazzy89:
@corest gotcha. thank you. I'll ask then to @calvix

mazzy89:
e2e failed 😫let's go to debug again

PR #1371, opened 2019-01-29
Only try S3 bucket deletion when it's empty

tuommaki:
Instead of trying bucket deletion on every run despite of whether there are
objects or not, only try it when there are no objects left.


tuommaki:
Bucket being not found should not occur, but in case it does in some edge case, it's fine to ignore it and continue instead of blocking. When we clearly know whether bucket contains objects or not, I don't find any excuse to behave accordingly.

For me personally correctness of API usage has very high priority. This is no different from our customers using our API or configuring their workloads correctly. I would like to have this so I'll merge. Ping me if you have any issues with this code in your PR.

PR #1375, opened 2019-01-30
Update ipam dependency to fix subnet allocation

MarcelMue:
This pulls https://github.com/giantswarm/ipam/pull/20 through to aws-operator.

The underlying problem is that subnets do not choose a suitable starting point in all cases. This should fix any cluster creation issues we saw in `seal` and `gauss`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -363,11 +363,11 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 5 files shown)


MarcelMue:
I deployed this to `gauss` and started a cluster with 2 AZs which went fine.

PR #1403, opened 2019-02-12
Switch to coreos 1967.5.0 in v17patch2 and v21patch1

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5334


corest:
This was tested, works fine

PR #1413, opened 2019-02-21
Vendor audit policy

mazzy89:
Update vendor an add version bundle due to the recent change in k8scloudconfig for the Audit policy

cc https://github.com/giantswarm/installations/pull/652


mazzy89:
Tested yesterday; spin up a new cluster and no issues.

PR #1414, opened 2019-02-21
Increase mmapfs default limit in the nodes (from k8scloudconfig)

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/5418


pipo02mix:
Tested on ginger successfully

PR #1434, opened 2019-03-06
use-proper-hostname

calvix:
towards https://github.com/giantswarm/giantswarm/issues/5460 and fixing also https://github.com/giantswarm/giantswarm/issues/5383

  few other sources mention that hostname for aws k8s needs to be fixed like here: https://blog.scottlowe.org/2018/09/28/setting-up-the-kubernetes-aws-cloud-provider/


calvix:
result:
```
Name:               ip-10-1-2-62.eu-central-1.compute.internal
Roles:              worker
Labels:             aws-operator.giantswarm.io/version=4.8.0
                    beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=m4.xlarge
                    beta.kubernetes.io/os=linux
                    failure-domain.beta.kubernetes.io/region=eu-central-1
                    failure-domain.beta.kubernetes.io/zone=eu-central-1a
                    giantswarm.io/provider=aws
                    ip=10.1.2.62
                    kubernetes.io/hostname=ip-10-1-2-62.eu-central-1.compute.internal
                    node-role.kubernetes.io/worker=
                    role=worker
```

teemow:
Can't we fix this via DHCP? Having this in a systemd unit feels wrong. Also there should be https://github.com/coreos/coreos-metadata which has the hostname already. But doesn't exist on our machines.

mazzy89:
As here https://github.com/giantswarm/giantnetes-terraform/pull/215#issuecomment-470034428

imo it is not the way to go. in the upstream projects like kops the way is to leave unchanged the dns hostname and use the flag `--hostname-override` to set the correct dns hostname coming from EC2 metadata.

calvix:
The thing is that we want to fix it only for AWS, if we go with `--hostname-override` we need to  adjust it for all platforms as the source of kubelet args is in k8scloudconfig and that is provider independent so far.

`coreos-metadata` is by default on all AWS CoreOS images, its just disabled.

mazzy89:
Can't we add a conditional on that flag? something like

```
{{if eq .Provider "aws"}} 
--hostname-override=
{{ end }}
```

mazzy89:
ok if it is completely provider independent, then I don' see any other way to achieve that

calvix:
tested, even conformance test is green

PR #1527, opened 2019-03-25
Pin calico-kube-controllers to master instance

tuommaki:
In order to reduce unnecessary disruptions and churn, pin
calico-kube-controllers POD to master instance so it doesn't get continuously
rescheduled during upgrade.


tuommaki:
For the record, this was for networking PM: https://github.com/giantswarm/giantswarm/issues/5584

PR #1528, opened 2019-03-25
Make kubelet volume same size as docker volume

mazzy89:
Make kubelet volume same size as docker volume

When the customer (or us) will change the value in the CR for the docker volume size then this will trigger the change also in the kubelet volume.

### DIFF
--- a/service/controller/v25/adapter/guest_launch_configuration.go
+++ b/service/controller/v25/adapter/guest_launch_configuration.go
@@ -70,24 +70,6 @@ func (l *GuestLaunchConfigAdapter) Adapt(config Config) error {
 		}
 	}
 
-	{
-		cur := config.StackState.WorkerKubeletVolumeSizeGB
-		def := defaultEBSVolumeSize
-
-		if cur == "" {
-			cur = "0"
-		}
-
-		curi, err := strconv.Atoi(cur)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-
-		if curi <= 0 {
-			config.StackState.WorkerKubeletVolumeSizeGB = def
-		}
-	}
-
 	l.WorkerBlockDeviceMappings = []BlockDeviceMapping{
 		{
 			DeleteOnTermination: true,
@@ -104,7 +86,10 @@ func (l *GuestLaunchConfigAdapter) Adapt(config Config) error {
 		{
 			DeleteOnTermination: true,
 			DeviceName:          kubeletEBSVolumeMountPoint,
-			VolumeSize:          config.StackState.WorkerKubeletVolumeSizeGB,
+			// TL;DR; kubelet volume same size as docker volume 
+			// this is a temporary solution that should stay around until Node Pools story is ready.
+			// See here for furhter info https://github.com/giantswarm/giantswarm/issues/5582#issuecomment-476170597
+			VolumeSize:          config.StackState.WorkerDockerVolumeSizeGB,
 			VolumeType:          defaultEBSVolumeType,
 		},
 	}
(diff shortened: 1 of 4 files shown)


service/controller/v25/adapter/guest_launch_configuration.go
```diff
@@ -102,9 +84,12 @@ func (l *GuestLaunchConfigAdapter) Adapt(config Config) error {
 			VolumeType:          defaultEBSVolumeType,
 		},
 		{
+			// TL;DR; kubelet volume same size as docker volume
+			// this is a temporary solution that should stay around until Node Pools story is ready.
+			// See here for furhter info https://github.com/giantswarm/giantswarm/issues/5582#issuecomment-476170597
 			DeleteOnTermination: true,
 			DeviceName:          kubeletEBSVolumeMountPoint,
-			VolumeSize:          config.StackState.WorkerKubeletVolumeSizeGB,
+			VolumeSize:          config.StackState.WorkerDockerVolumeSizeGB,
```
calvix: this could be potentially dangerous when dockerVolumeSizeGb is not properly set, that can be true for very old clusters who were upgraded to the latest version but the value was never set ( at least I think this can happen as we don't have proper migration policies when CRD changes are introduced)

What about something safe like in kvm-operator
https://github.com/giantswarm/kvm-operator/blob/master/service/controller/v21/resource/deployment/worker_deployment.go#L181-L184
https://github.com/giantswarm/kvm-operator/blob/master/service/controller/v21/key/key.go#L274-L287

mazzy89: I thought that if the value is then never set there is a fallback for the `config.StackState.WorkerDockerVolumeSizeGB` which is defaultEBSVolumeSize and this value could have been propagated also to kubelet volume during the reconciliation. but it seems not then I'll go with that way. thank you @calvix

calvix: TBH I am not sure how `config.StackState.WorkerDockerVolumeSizeGB` is initialised, if its just copying values from CRD than we might hit issue, if there is some logic than it might be OK

mazzy89: I've pushed a fix but unfortunately, it does not work as expected and I think I'm missing something at the level of how those values are initialized. 

A test is failing and I don't have a clue of why it is failing even though I've debugged it, I could not pull out a rabbit from the hat.

```
--- FAIL: Test_AdapterLaunchConfiguration_RegularFields (0.00s)
    --- FAIL: Test_AdapterLaunchConfiguration_RegularFields/basic_matching,_all_fields_present (0.00s)
        guest_launch_configuration_test.go:134: unexpected BlockDeviceMappings, got [{true /dev/xvdh 250 gp2} {true /dev/xvdf 100 gp2} {true /dev/xvdg 100 gp2}], want [{true /dev/xvdh 250 gp2} {true /dev/xvdf 100 gp2} {true /dev/xvdg 250 gp2}]
```

When `DockerVolumeSizeGB` is defined (https://github.com/giantswarm/aws-operator/blob/06fc2458747170e076a7b40a143732c1c139129b/service/controller/v25/adapter/guest_launch_configuration_test.go#L35) in the CR and it is equal to `250` then I'd expect that the kubelet volume would have the same value (https://github.com/giantswarm/aws-operator/blob/06fc2458747170e076a7b40a143732c1c139129b/service/controller/v25/adapter/guest_launch_configuration_test.go#L60). However this is not happening and it's not the case. The kubelet volume fallback to the default value for the docker which is `100`. Something from the initialization logic, it escapes from my understanding.

mazzy89:
Tested and it works. when the `dockerVolumeSizeGb` is not set then by changing the `dockerVolumeSizeGb` both of the volumes docker and kubelet are correctly resized.

PR #1535, opened 2019-04-02
Make sure instances in service are lower than max when upgrading

MarcelMue:
Fixes the issue that we can't upgrade one node clusters right now. As the minimum instances available is the same as the number of instances, aws will simply refuse our changes.


MarcelMue:
Do we want to backport this as well?
There are older one node clusters which are now stuck.

service/controller/v25/adapter/guest_auto_scaling_group.go
```diff
@@ -49,8 +49,18 @@ func (a *GuestAutoScalingGroupAdapter) Adapt(cfg Config) error {
...
-	a.MaxBatchSize = workerCountRatio(currentDesiredMinWorkers, asgMaxBatchSizeRatio)
-	a.MinInstancesInService = workerCountRatio(currentDesiredMinWorkers, asgMinInstancesRatio)
+	a.MaxBatchSize = strconv.Itoa(workerCountRatio(currentDesiredMinWorkers, asgMaxBatchSizeRatio))
+
+	minInstancesInService := workerCountRatio(currentDesiredMinWorkers, asgMinInstancesRatio)
+	if minInstancesInService != 0 && minInstancesInService == maxWorkers {
+		// MinInstancesInService must be less than the autoscaling group's MaxSize.
+		// This should only ever be an issue if the min and max workers are exactly one.
+		// If this is the case then the cluster will most likely go down while the new worker
+		// is coming up and the old one is removed.
```
kopiczko: This comment makes me think if we can't make the condition stronger and check if min and max workers are qual to `1`.

tomahawk28:
In unittest `TestWorkerCountRatioMaxBatchSize`, we are still using `asgMaxBatchSizeRatio`, which fixed to `0.3`, and I don't think we should use a value in the package, can we add one more field in test struct for this?

service/controller/v25/adapter/guest_auto_scaling_group.go
MarcelMue: I changed this condition PTAL

MarcelMue:
@tomahawk28 I can do that in a separate PR only in the latest version. I don't want to backport that all the way like I do with the changes here. Is that okay for you?

service/controller/v23/adapter/guest_auto_scaling_group.go
```diff
@@ -49,8 +49,18 @@ func (a *GuestAutoScalingGroupAdapter) Adapt(cfg Config) error {
 	a.ASGMinSize = minWorkers
 	a.ASGType = key.KindWorker
 	a.ClusterID = key.ClusterID(cfg.CustomObject)
-	a.MaxBatchSize = workerCountRatio(currentDesiredMinWorkers, asgMaxBatchSizeRatio)
-	a.MinInstancesInService = workerCountRatio(currentDesiredMinWorkers, asgMinInstancesRatio)
+	a.MaxBatchSize = strconv.Itoa(workerCountRatio(currentDesiredMinWorkers, asgMaxBatchSizeRatio))
+
+	minInstancesInService := workerCountRatio(currentDesiredMinWorkers, asgMinInstancesRatio)
+	if minWorkers == 1 && maxWorkers == 1 && minInstancesInService == maxWorkers {
```
kopiczko:
```suggestion
	if minWorkers == 1 && maxWorkers == 1 && minInstancesInService == 1 {
```

service/controller/v23/adapter/guest_auto_scaling_group.go
```diff
@@ -49,8 +49,18 @@ func (a *GuestAutoScalingGroupAdapter) Adapt(cfg Config) error {
...
+		// MinInstancesInService must be less than the autoscaling group's MaxSize.
+		// This should only ever be an issue if the min and max workers are exactly one.
+		// If this is the case then the cluster will most likely go down while the new worker
+		// is coming up and the old one is removed.
+		minInstancesInService = minInstancesInService - 1
```
kopiczko:
```suggestion
		minInstancesInService = 0
```

service/controller/v23/adapter/guest_auto_scaling_group.go
kopiczko: I also think we can remove the third part of the condition all together. Is it possible that minInstancesInService  isn't 0 or 1 in that case?

MarcelMue: It currently is always 1 when this is the case, but yeah the condition is not really needed. I removed the third part and will test again to make sure that everything works correctly.

MarcelMue:
Tested again on ginger by upgrading from `7.2.0` to `7.3.0` and works like a charm.

PR #1536, opened 2019-04-04
Allow change of master instance type while scaling

tuommaki:
There are cases where customer doesn't want upgrades on a cluster, but master
instance type should get changed and it should not require full upgrade nor
should it block cluster scaling. Therefore allowing scaling when there's
pending cluster scaling request allows the master instance type to be changed
"on-the-fly" without disturbing workload.

:exclamation:  This doesn't change the functionality of `v25`. From quick look that doesn't obey `Updates.Enabled` flag anymore and therefore performs the instance type change when one is applied to CR (so effectively it doesn't need any change for this).


mazzy89:
So i can now start to test it.

mazzy89:
tested. it works like a charm. thank you @tuommaki

PR #1539, opened 2019-04-15
v26: copy v25

rossf7:
Adding new WIP version for use with k8s 1.14.


rossf7:
Copy PR. Build is green. Going with this.

PR #1541, opened 2019-04-15
Wire v26

rossf7:


service/controller/v26/version_bundle.go
```diff
@@ -41,6 +36,6 @@ func VersionBundle() versionbundle.Bundle {
 			},
 		},
 		Name:    "aws-operator",
-		Version: "4.9.0",
+		Version: "5.0.0",
```
rossf7: @MarcelMue FYI. I've gone with major bump so you can use for k8s 1.14.

rossf7:
Tested on `gauss`.

PR #1563, opened 2019-04-30
Wire new network allocator svc to IPAM resources

tuommaki:
This change wires shared network allocator service instance to all IPAM
resources. This change doesn't make use of new service yet, but is a
preparation step to next convert all resources to use that service.


MarcelMue:
Please take care, I am creating a new versionbundle!
So new copy PRs and so on coming.

PR #1572, opened 2019-04-30
Make decision factors visible when determining scaling (and update)

tuommaki:
I am investigating a postmortem where aws-operator is applying CF stack update
due to changes in scaling settings while CR object contents don't have any
differences between current and desired state. This then leads to AWS API
returning validation error because nothing changes from its perspective either.

Exposing the variables that lead to specific outcome helps to understand why
the decision was made.

Towards https://github.com/giantswarm/giantswarm/issues/5721


service/controller/clusterapi/v27/detection/detection.go
```diff
@@ -46,11 +46,11 @@ func (d *Detection) ShouldScale(ctx context.Context, cr v1alpha1.AWSConfig) (boo
 	}
 
 	if !cc.Status.TenantCluster.TCCP.ASG.IsEmpty() && cc.Status.TenantCluster.TCCP.ASG.MaxSize != legacykey.ScalingMax(cr) {
-		d.logger.LogCtx(ctx, "level", "debug", "message", "detected the tenant cluster should scale due to scaling max changes")
+		d.logger.LogCtx(ctx, "level", "debug", "message", "detected the tenant cluster should scale due to scaling max changes: cc.Status.TenantCluster.TCCP.ASG.MaxSize is %d while legacykey.ScalingMax(cr) is %d", cc.Status.TenantCluster.TCCP.ASG.MaxSize, legacykey.ScalingMax(cr))
```
kopiczko: I would skip `cc` in those messages like:
```suggestion
		d.logger.LogCtx(ctx, "level", "debug", "message", "detected the tenant cluster should scale due to scaling max changes: .Status.TenantCluster.TCCP.ASG.MaxSize is %d while legacykey.ScalingMax(cr) is %d", cc.Status.TenantCluster.TCCP.ASG.MaxSize, legacykey.ScalingMax(cr))
```

Also using function names from key is cryptic but I think it's ok first step.

I also started discussion to address that: https://gigantic.slack.com/archives/C2MS2VB37/p1556637615023500

tuommaki: Why would you remove `cc` when that's reference to our internal context? That `.Status...` prefix makes me intuitively think about CR object which this is not.

kopiczko: > makes me intuitively think about CR object

That was my intention. I think those paths are the same. I'd prefer to see decision made based on the CR. That is more understandable for someone not up to date with the codebase.

tuommaki: > I think those paths are the same. I'd prefer to see decision made based on the CR.

Those are quite different from CR. The paths in CR are:
* `.Status.Cluster.Scaling.DesiredCapacity`
* `.Spec.Cluster.Scaling.Max`

Also: `cc` stands for `controllercontext` which is the internal variable that's used to hold state from cloud provider. I don't understand how the message even ideally could be about having difference between CR vs. CR?

Maybe we should hang on this as I think we are talking about different things possibly?

kopiczko: Oh yeah. Not everything in cc comes from the CR status. Fine then. Let's go!

PR #1580, opened 2019-05-03
update e2esetup

TheoBrigitte:
Update e2esetup with latest changes.

Where `k8s.SetupConfig.K8sClient` changed to `k8s.SetupConfig.Clients`


TheoBrigitte:
Need some review here.

PR #1581, opened 2019-05-03
Fix malformed policy

mazzy89:
~~**Disclaimer**: this is just a **draft** and it is not meant to be merged because the change has been put intentionally in the wrong bundle just to show off to the team a proposal of changes after investigation of the PM https://github.com/giantswarm/giantswarm/issues/5872~~

**Axolotl** is affected by an issue. None can't create a cluster there due to an issue in CF. The issue is reported in the PM. 

KMS is not available in China. Despite that, the worker policy is created with the action `kms:decrypt` and the KMS ARN is passed anyway to the CF template. However, this ARN is malformed because it is generated with the Vault encrypter which returns as key not an ARN but the Cluster ID. So we get an error like the following from CF:

```
Resource 29h9p must be in ARN format or "*". (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: 29f780bf-6827-11e9-b906-456f9fdc775b)
```

I've proposed here a solution that could prevent to create the policy action `kms:decrypt` in the worker policy. I'm not sure though about any side effects that we could have by creating policy without kms action.


MarcelMue:
I would recommend just testing this out in china - `giraffe` is marked as a test env, so maybe see what happens?

I can't really tell what impact the change will have.

mazzy89:
Ok I could test it in `giraffe`. Apart this I'd hear some opinions from some veteran of the operator about the fix. could it be done better?

mazzy89:
If I had to create a bundle, should I create it as a patch for `v25`? hence `v25patch1` right?

service/controller/legacy/v25/adapter/guest_iam_policies.go
```diff
@@ -30,7 +30,11 @@ func (i *GuestIAMPoliciesAdapter) Adapt(cfg Config) error {
 	i.WorkerProfileName = key.InstanceProfileName(cfg.CustomObject, key.KindWorker)
 	i.WorkerRoleName = key.RoleName(cfg.CustomObject, key.KindWorker)
 	i.RegionARN = key.RegionARN(cfg.CustomObject)
-	i.KMSKeyARN = cfg.TenantClusterKMSKeyARN
+
+	if cfg.TenantClusterKMSKeyARN != "" {
+		i.KMSKeyARN = cfg.TenantClusterKMSKeyARN
+	}
```
xh3b4sd: When you check the templates you see that there is already a condition to cover this. So IMO this change is not necessary.

service/controller/legacy/v25/resource/tccp/create.go
```diff
@@ -318,7 +318,10 @@ func (r *Resource) newTemplateBody(ctx context.Context, cr v1alpha1.AWSConfig, t
 				VersionBundleVersion: key.VersionBundleVersion(cr),
 			},
 			TenantClusterAccountID: cc.Status.TenantCluster.AWSAccountID,
-			TenantClusterKMSKeyARN: cc.Status.TenantCluster.Encryption.Key,
+		}
+
+		if r.encrypterBackend == encrypter.KMSBackend {
+			c.TenantClusterKMSKeyARN = cc.Status.TenantCluster.Encryption.Key
 		}
```
xh3b4sd: I don't have a better solution for this. I just checked other conditions towards the backend types and they are all handled in the resource. Good enough for now.

xh3b4sd:
When we want to patch v25 then we need to create a patch from it. I would ask why we can't put the fix in the latest version and upgrade customer workloads.

service/controller/legacy/v25/adapter/guest_iam_policies.go
mazzy89: Yeah I was thinking the same. It's already handled there so it should be enough to have it in the CF template

mazzy89:
As discussed for the moment I'm going to test this in giraffe as it is. then we'll create the chanages in the right bundle

mazzy89:
I've tested and the stack `cluster-7czgc-guest-main` is created normally. The error mentioned here is not present after this change.

However another error is present in the `host` stack.

image

I don't have the capacity today to look further due to my ops duty. I'll check it tomorrow

mazzy89:
image

I've found the issue and the reason is that the VPC Peering is created with the `axolotl` CP and not the `giraffe` CP. so here the issue.

Not sure though how it could happen need to debug it

mazzy89:
The issue here is that the VPC peering ID during the creation of the VPC Peering is of `axolotl` and not `giraffe`. Not sure though how it gets set up.

By the code, it seems that it is retrieved by https://github.com/giantswarm/aws-operator/blob/a4c6e37fb7dd1d5f2c50e2d786cdd2c4d0dced5a/service/controller/legacy/v27/key/key.go#L393 but not sure how it is set up. cc @giantswarm/sig-operator ?

xh3b4sd:
The latest change in the `aws-operator` is that the CP VPC ID is configured on deployment via installation repo settings. The VPC IDs in `giraffe` and `axolotl` are the same, and I am not sure this is correct. 

* https://github.com/giantswarm/installations/blob/master/axolotl/draughtsman-configmap-values.yaml#L46
* https://github.com/giantswarm/installations/blob/master/giraffe/draughtsman-configmap-values.yaml#L30

mazzy89:
@xh3b4sd ♥️ thanks a lot. we found out the issue. cleared out

mazzy89:
I've updated the draughtsman config values and the aws-operator configmap has been updated accordingly as @xh3b4sd pointed during our call. The configmap is mounted correctly and made available inside the container. The VPC Peer ID in the `config.yaml` looks correct.

Despite all of this, a new cluster running the command:

```sh
gsctl create cluster -o giantswarm -n "Salvo - test PM" -r 8.1.0
```

fails because the VPC Peering is created with the wrong VPC Accepter ID.

xh3b4sd:
Which one is the wrong accepter ID and which one should it be?

mazzy89:
Wrong VPC Peer ID: `vpc-50a3dc34` (axolotl)

Correct VPC Peer ID: `vpc-0d1a9a8f1fc1ae8d9` (giraffe)

mazzy89:
Once this [fix](https://github.com/giantswarm/aws-operator/pull/1621) is in, i'll merge it and continue to test.

mazzy89:
Still, China is showing itself to be pretty nasty 😷 Without even nominating how many times i need to retries a command before it success w/o timeout.

After the last fixes, CloudFormation is now all green, instances are up. Though the cluster does not come up.

This ~1h after created the cluster.

```sh
➜ kgno
Unable to connect to the server: EOF
```

I could not ssh on the machines because they asked for password. Something fishy is happening there around the authentication. Not sure why `giantswarm` user does not work. I've pushed to installations my ssh to be created in the tenant clusters so I can troubleshoot better.

operator does not show any error in the logs. (i've tried for error by grep'ing - `|grep error`)

Not sure if in China it needs more time to create a cluster. I'll leave the cluster running to see if it goes up.

xh3b4sd:
You cannot SSH into machines because your key is not there. The keys are propagated from the AWSConfig CRs. We need to check why this does not work, as your key seems to be in the installations repo.

mazzy89:
I've added the key after I've created the cluster because I've realized it was missing from there. Nevertheless is not mandatory that the key is in the CRD. Indeed i can access other installastions where my key is not there because we use the giantswarm user through the sso. Anyway I'll propagate the change to giraffe so my key will be listed there and I can access to continue the troubleshotting.

mazzy89:
Some updates here to break the radio silence. I've spent the entire day to figure out one other issue in giraffe. I've realized that the vault token has probably expired. I have then to roll out the entire installations. Unfortunately, in China this must be done manually and then update manually the Route53 endpoint. I'll do this as next and then hope to get back to troubleshoot the real problem

mazzy89:
Rollout out `giraffe` with new Vault tokens after cluster-service reported issues.

I've tried to create a new cluster.

I've found out that aws-operator reports this error at the bootstrap

```
➜ stern aws-operator -ngiantswarm -t|grep error
+ aws-operator-5cf764b5b6-kndmc › aws-operator
aws-operator-5cf764b5b6-kndmc aws-operator 2019-05-13T23:03:11.924020971Z {"caller":"github.com/giantswarm/aws-operator/service/controller/legacy/v27/encrypter/vault/vault.go:470","level":"error","message":"invalid HTTP status code, want 200, got 400, response body: \"{\\\"errors\\\":[\\\"missing client token\\\"]}\\n\"","time":"2019-05-13T23:03:11.923879+00:00"}
```

CF is green, instances are created. Trying to list nodes `k get nodes` does not work, a symptom that API is down and cluster overall down.

I've tried to access the machines. Still not possible. By looking at the `awsconfigs` CR, my ssh user is not listed there. It is in the draughtsman ConfigMap, though. 

```sh
kdcm draughtsman-values-configmap -ndraughtsman|grep salvo
```

Maybe another issue with propagating ssh user property to the operator. Not sure though. I haven't investigated it.

I'll try to figure out that error.

xh3b4sd:
The keys should be set in https://github.com/giantswarm/kubernetesd/blob/master/service/creator/cluster.go#L132-L139.

mazzy89:
Oh then it make sense. the `kubernetesd` configmap is not up to date.

mazzy89:
Another issue here that prevent the cluster to be healthy is that the CP public route tables don't have any entries for the VPC Peering connection to the TC. 

That's important because Vault must be routable and accessible from the TC to run `decrypt-tls-assets` systemd

mazzy89:
Just to make this issue a bit more clear. China is the only case where we use Vault to decrypt asset to Vault must be routable from the TC

mazzy89:
https://github.com/giantswarm/installations/pull/747 - this PR should address the issue mentioned in here https://github.com/giantswarm/aws-operator/pull/1581#issuecomment-492641079

mazzy89:
tested https://github.com/giantswarm/installations/pull/747 and unfortunately, it addresses only the Vault routing. 

China is still broken. 🤬

I've investigated on a new TC created with the fix from https://github.com/giantswarm/installations/pull/747 and I've noticed that the API server does not come up due to the following error:

```
E0515 17:08:58.896995    2129 pod_workers.go:190] Error syncing pod 9f5c7929a1121cb33f50821cec43b85e ("k8s-api-server-ip-172-19-163-27.cn-north-1.compute.internal_kube-system(9f5c7929a1121cb33f50821cec43b85e)"), skipping: failed to "StartContainer" for "k8s-api-server" with CrashLoopBackOff: "Back-off 5m0s restarting failed container=k8s-api-server pod=k8s-api-server-ip-172-19-163-27.cn-north-1.compute.internal_kube-system(9f5c7929a1121cb33f50821cec43b85e)"
E0515 17:08:58.923811    2129 kubelet.go:2244] node "ip-172-19-163-27.cn-north-1.compute.internal" not found
W0515 17:08:58.951347    2129 cni.go:213] Unable to update cni config: No networks found in /etc/cni/net.d
```

mazzy89:
Created another cluster and now the reason why it can't spin up is:

```
15 19:17:37 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: Starting etcd3...
May 15 19:17:38 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: Started etcd3.
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.274783 I | etcdmain: etcd Version: 3.3.10
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.275819 I | etcdmain: Git SHA: 27fc7e2
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.276190 I | etcdmain: Go Version: go1.10.4
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.276568 I | etcdmain: Go OS/Arch: linux/amd64
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.276900 I | etcdmain: setting maximum number of CPUs to 2, total number of available CPUs is 2
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.277322 W | etcdmain: found invalid file/dir lost+found under data dir /var/lib/etcd (Ignore this if you are upgrading etcd)
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.277726 I | embed: peerTLS: cert = /etc/etcd/server-crt.pem, key = /etc/etcd/server-key.pem, ca = , trusted-ca = /etc/etcd/server-ca.pem, client-cert-auth = true, crl-file =
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal docker[14775]: 2019-05-15 19:17:39.278292 C | etcdmain: tls: failed to find any PEM data in certificate input
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: etcd3.service: Main process exited, code=exited, status=1/FAILURE
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: etcd3.service: Failed with result 'exit-code'.
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: etcd3.service: Service has no hold-off time, scheduling restart.
May 15 19:17:39 ip-172-19-163-81.cn-north-1.compute.internal systemd[1]: etcd3.service: Scheduled restart job, restart counter is at 450.
```

mazzy89:
I spoke with @corest and we agreed that this PR solve the issue that it was opened for. So going to merge it and another fix for China is coming as well.

PR #1583, opened 2019-05-06
Improve logging

tuommaki:


tuommaki:
Next in the row: Flexbox and some animations to logs :rocket:

PR #1643, opened 2019-05-15
Ensure VPCPeerID during operator creation

MarcelMue:
Doesn't solve flapping on AWS e2e but atleast makes them run again.

Problem was that the VPC ID is now passed as a command line argument into aws-operator.


Gopkg.toml
```diff
@@ -61,7 +61,7 @@ required = [
   name = "github.com/giantswarm/e2eclients"
 
 [[constraint]]
-  branch = "master"
+  branch = "fix-aws"
```
MarcelMue: TODO

PR #1653, opened 2019-05-17
Use desired capacity to count number nodes

pipo02mix:
Cluster custom resource does not reflect the real state anymore, using the desired capacity works even if the autoscaling is active.


pipo02mix:
Tested on gauss

PR #1655, opened 2019-05-19
Refactor decryptor systemd units

corest:
Fix China decrypter units provisioning

This PR introduces new service `vault-aws-authorizer`, which takes care about aouthorizing into vault via aws auth backend, then running `decrypt-tls-assets` and then `decrypt-keys-assets`. The reason behind that is that there is some bug in vault, which makes decrypter units running into race condition, when keys assets decrypter gets tls assets content


corest:
In general we should rewrite that bash into go code, specially if we're going to move to vault decrypter for all envs. But I'll leave that for responsible team. Now I'm just fixing shit

corest:
I would suggest looking into moving from kms to vault decryption. In that way, we will not need to cover both cases. But that's hard to test, yes. Merging, for now, to make it work for customers

PR #1697, opened 2019-06-20
RBAC: Add statuses to cluster.k8s.io types' resource lists

tuommaki:
aws-operator modifies status part of cluster.k8s.io types for e.g. IPAM
purposes, so it must have permission for them.


tuommaki:
Works. Going with this.

PR #1698, opened 2019-06-20
Temporarily take worker AZs instead of all AZs

tuommaki:
To proceed with CP stack creation, take temporarily worker AZs instead of
cluster AZs when creating ASG.


tuommaki:
Tested this last week already. Going.

PR #1720, opened 2019-07-03
Prepare for dynamically changing cluster availability zones

tuommaki:
This change adapts AZ specific subnet allocation to be deterministic which then
allows any combination of cluster AZ allocation to dynamically be made during
tenant cluster lifecycle.

**NOTE:** This now changes the CloudFormation resource naming for most of the
network related resources which completely breaks compatibility with older
clusters. This is necessary though because older index based naming scheme does
not work when there is non-consecutive set of allocated AZs.


service/controller/clusterapi/v29/adapter/guest_instance.go
```diff
@@ -77,8 +77,8 @@ func (i *GuestInstanceAdapter) Adapt(config Config) error {
 			return microerror.Maskf(notFoundError, "CustomObject has no availability zones")
 		}
 
-		i.Master.AZ = zones[0].Name
-		i.Master.PrivateSubnet = key.PrivateSubnetName(0)
+		i.Master.AZ = key.MasterAvailabilityZone(config.CustomObject)
+		i.Master.PrivateSubnet = key.PrivateSubnetName(i.Master.AZ)
```
tuommaki: This does tie master to specified AZ but rest of the code in this PR doesn't have necessary support for it.

service/controller/clusterapi/v29/adapter/guest_route_tables.go
```diff
@@ -17,17 +17,19 @@ type GuestRouteTablesAdapter struct {
 }
 
 func (r *GuestRouteTablesAdapter) Adapt(cfg Config) error {
+	workerAZs := key.SortedWorkerAvailabilityZones(cfg.MachineDeployment)
+
 	r.HostClusterCIDR = cfg.ControlPlaneVPCCidr
 	r.PublicRouteTableName = RouteTableName{
 		ResourceName: "PublicRouteTable",
-		TagName:      key.RouteTableName(cfg.CustomObject, suffixPublic, 0),
+		TagName:      key.RouteTableName(cfg.CustomObject, suffixPublic, workerAZs[0]),
```
tuommaki: These magical "first AZ" entries made me worry a bit, but we need to tackle these together with overhaul for keeping stable master AZ as well.

service/controller/clusterapi/v29/adapter/adapter.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 	StackState                      StackState
 	TenantClusterAccountID          string
 	TenantClusterKMSKeyARN          string
+	TenantClusterAvailabilityZones  []string
```
tuommaki: This value comes from `controllercontext` that has all distinct cluster AZs (master AZ + AZs from all `MachineDeployments`).

service/controller/clusterapi/v29/adapter/guest_auto_scaling_group.go
```diff
@@ -64,14 +64,9 @@ func (a *GuestAutoScalingGroupAdapter) Adapt(cfg Config) error {
...
-	if err != nil {
-		return microerror.Mask(err)
-	}
-
-	for i, az := range azs {
-		a.PrivateSubnets = append(a.PrivateSubnets, key.PrivateSubnetName(i))
-		a.WorkerAZs = append(a.WorkerAZs, az.Name)
+	for _, az := range cfg.TenantClusterAvailabilityZones {
+		a.PrivateSubnets = append(a.PrivateSubnets, key.PrivateSubnetName(az))
+		a.WorkerAZs = append(a.WorkerAZs, az)
```
tuommaki: I'm thinking if we maybe want to remove distinction between Master and Worker AZs and move towards Cluster AZs?

service/controller/clusterapi/v29/key/common.go
```diff
@@ -10,100 +10,66 @@ func IsDeleted(getter DeletionTimestampGetter) bool {
...
-	// resource name for first AZ.
-	if idx < 1 {
-		return "VPCPeeringRoute"
+// SanitizeCFResourceName filters out all non-ascii alphanumberics from input
+// string.
+//
+// Example: SanitizeCFResourceName("abc-123") == "abc123"
+// Example2: SanitizeCFResourceName("Dear god why? щ（ﾟДﾟщ）") == "Deargodwhy"
+//
+func SanitizeCFResourceName(v string) string {
```
MarcelMue: Is this really necessary? Shouldn't we already have full control over the input?

tuommaki: What do you mean?

MarcelMue: We sorted this while chatting.

tuommaki:
For the record the target here:

When [this](https://github.com/giantswarm/aws-operator/pull/1751) PR is merged, then usage of `key.StatusAvailabilityZones()` should be replaced by use of `controllercontext.Status.TenantCluster.AvailabilityZones`. Currently these are the locations:
```$ grep -nR 'key.StatusAvailabilityZones' service/controller/clusterapi/
service/controller/clusterapi/v29/adapter/guest_subnets.go:32:  zones, err := key.StatusAvailabilityZones(cfg.MachineDeployment)
service/controller/clusterapi/v29/adapter/guest_auto_scaling_group.go:67:       azs, err := key.StatusAvailabilityZones(cfg.MachineDeployment)
service/controller/clusterapi/v29/adapter/guest_instance.go:67:         zones, err := key.StatusAvailabilityZones(config.MachineDeployment)
service/controller/clusterapi/v29/resource/cpf/create.go:138:           azs, err := key.StatusAvailabilityZones(cc.Status.TenantCluster.TCCP.MachineDeployment)
service/controller/clusterapi/v29/resource/tcnp/create.go:277:  azs, err := key.StatusAvailabilityZones(md)
```

That way the allocated subnets for existing AWS resources are kept as-is and when there are AZs dropped and added later, the existing resources don't get recreated.

service/controller/clusterapi/v29/adapter/guest_subnets.go
```diff
@@ -44,30 +44,30 @@ func (s *GuestSubnetsAdapter) Adapt(cfg Config) error {
...
 		snet := Subnet{
 			AvailabilityZone:    az.Name,
 			CIDR:                az.Subnet.Public.CIDR,
 			Name:                snetName,
 			MapPublicIPOnLaunch: false,
 			RouteTableAssociation: RouteTableAssociation{
-				Name:           key.PublicSubnetRouteTableAssociationName(i),
-				RouteTableName: "PublicRouteTable",
+				Name:           key.SanitizeCFResourceName(key.PublicSubnetRouteTableAssociationName(az.Name)),
+				RouteTableName: key.SanitizeCFResourceName(key.PublicRouteTableName(key.MasterAvailabilityZone(cfg.CustomObject))),
```
tuommaki: :question: Should this be hooked to master's AZ or should this be for corresponding AZ?

service/controller/clusterapi/v29/adapter/adapter.go
```diff
@@ -48,6 +50,7 @@ type Config struct {
 	StackState                      StackState
 	TenantClusterAccountID          string
 	TenantClusterKMSKeyARN          string
+	TenantClusterAvailabilityZones  []controllercontext.ContextStatusTenantClusterAvailabilityZone
```
tuommaki: Is it ok to use this type from `controllercontext` or should this type be duplicated & copied to one in here? I don't think there will ever be import loops in this case so from that point of view this should be fine.

service/controller/clusterapi/v29/controllercontext/cluster_az.go
```diff
@@ -0,0 +1,9 @@
+package controllercontext
+
+import "net"
+
+type ContextTenantClusterAvailabilityZone struct {
```
MarcelMue: I created this as a shared struct between `cc` `status` and `spec`. This made me slightly more sane and allows for some neat refactorings in follow up PRs.

service/controller/clusterapi/v29/adapter/guest_subnets.go
MarcelMue: We still need to clear this up but this is a follow up step IMO.

MarcelMue:
This is a pretty big PR, it is tested with the golden file and manually. I will continue to improve this code, want to get this in for AZ subnet management though.

PR #1721, opened 2019-07-03
Fix typo in collector Set type godoc

sslavic:


sslavic:
Some e2e tests are not passing, but they are also named WIP; having in mind also that this is only godoc change I'll merge the PR.

PR #1741, opened 2019-07-09
Add two util functions to work with network addresses

tuommaki:
CalculateParent() returns next big network for given network.

Filter() is basic filter function for slice of networks.


tuommaki:
Dropped in favor of https://github.com/giantswarm/ipam/pull/24

PR #1808, opened 2019-07-25
push to docker

TheoBrigitte:
Towards: giantswarm/giantswarm#6557

Use the new CircleCI job to push docker image with version tag


.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ e2eTest: &e2eTest
         when: always
 
 orbs:
-  architect: giantswarm/architect@0.1.1
+  architect: giantswarm/architect@dev:b94d01c
```
TheoBrigitte: Note: update this before merging

.circleci/config.yml
```diff
@@ -230,14 +230,24 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+      - architect/push-to-docker:
+          username: "${QUAY_USERNAME}"
+          password: "${QUAY_PASSWORD}"
```
kopiczko: Why do we need to pass that when they are read from env var in the job?

.circleci/config.yml
```diff
@@ -230,14 +230,24 @@ workflows:
...
+          registry: "quay.io"
+          image: "quay.io/giantswarm/aws-operator"
```
kopiczko: I don't like that repetition. We should take registry name from the image.

.circleci/config.yml
TheoBrigitte: Passing those values here make it clear and explicit which values are used by the orb, instead of reading env vars from within the orb.

.circleci/config.yml
TheoBrigitte: I don't know if this would be valid for other repositories like aliyun. Those are 2 differents values and should stay this way for less confusion IMO:
`registry` is the url used for login
`image` is the docker image name

kopiczko: We checked and it's valid.

.circleci/config.yml
TheoBrigitte: We decided to pass env var names instead of the value, for safety (password value won't ever be exposed).

.circleci/config.yml
```diff
@@ -230,14 +230,23 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+      - architect/push-to-docker:
+          image: "quay.io/giantswarm/aws-operator"
+          username_envar: "QUAY_USERNAME"
+          password_envar: "QUAY_PASSWORD"
```
kopiczko: At some point it would be cool to rename those. But at least it's clear where they are used.

TheoBrigitte:
Tests are broken since a long time, going with this.

kopiczko:
> Tests are broken since a long time, going with this.

When images are pulled correctly in tests this is not connected.

PR #1829, opened 2019-08-01
DisableVersionBundleSelection flag propagation

TheoBrigitte:
Towards: giantswarm/giantswarm#6663

Propagate the new `DisableVersionBundleSelection` flag to every resources sets.


kopiczko:
Are you sure we need that everywhere? Maybe we can just not register all those bundles?

TheoBrigitte:
Closing as we are going with different and simplet approach.

PR #1850, opened 2019-08-05
Rename templates for aws-operator charts

rossf7:
See https://gigantic.slack.com/archives/C2MS2VB37/p1565002376113900.

The numeric prefixes are not needed as Helm takes care of creating resources in the correct order. Gets the naming straight since this is the SOTA operator.


rossf7:
Tested on `ginger`.

I'll merge in the morning because `axolotl` is in the slow zone.

PR #1880, opened 2019-08-14
integration: install credential-default Secret during setup

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/6736.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -60,8 +60,8 @@ required = [
   branch = "master"
   name = "github.com/giantswarm/e2eclients"
 
-[[override]]
-  branch = "chartvalues"
+[[constraint]]
+  branch = "master"
   name = "github.com/giantswarm/e2etemplates"
 
 [[constraint]]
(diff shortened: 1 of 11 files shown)


Gopkg.toml
```diff
@@ -60,8 +60,8 @@ required = [
   branch = "master"
   name = "github.com/giantswarm/e2eclients"
 
-[[constraint]]
-  branch = "master"
+[[override]]
+  branch = "chartvalues"
```
kopiczko: TODO: revert

service/service.go
```diff
@@ -93,6 +94,10 @@ func New(config Config) (*Service, error) {
 		}
 	}
 
+	{
+		i, _ := dynamic.NewForConfig(restConfig)
```
tfussell: What is this for? Doesn't seem to be being used and it's causing the e2e test build to fail. Sorry if this is still a WIP. Keeping an eye on these changes as I'm trying to get `aws-operator` tests to work.

kopiczko: Yeah this is some leftover.

PR #1940, opened 2019-09-17
Add internal elb for ingress

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6913

Added only for legacy as ingress will be removed in clusterapi

### DIFF
--- a/service/controller/legacy/v30/adapter/guest_load_balancers.go
+++ b/service/controller/legacy/v30/adapter/guest_load_balancers.go
@@ -35,8 +35,10 @@ type GuestLoadBalancersAdapter struct {
 	ELBHealthCheckUnhealthyThreshold int
 	IngressElbHealthCheckTarget      string
 	IngressElbName                   string
+	IngressInternalElbName           string
 	IngressElbPortsToOpen            []GuestLoadBalancersAdapterPortPair
 	IngressElbScheme                 string
+	IngressInternalElbScheme         string
 	MasterInstanceResourceName       string
 	PublicSubnets                    []string
 	PrivateSubnets                   []string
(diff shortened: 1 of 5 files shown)


corest:
@giantswarm/sig-operator sry about pinging, but need to move here and this blocks me

MarcelMue:
Why is this not added in the `cluster-api` branch? Is it not needed there?

corest:
@MarcelMue from the first message

> Added only for legacy as ingress will be removed in clusterapi

service/controller/legacy/v30/adapter/guest_load_balancers.go
```diff
@@ -111,6 +113,14 @@ func (a *GuestLoadBalancersAdapter) Adapt(cfg Config) error {
 	}
 	a.IngressElbScheme = externalELBScheme
 
+	// Ingress internal load balancer settings.
+	ingressInternalElbName, err := key.InternalLoadBalancerName(cfg.CustomObject.Spec.Cluster.Kubernetes.IngressController.Domain, cfg.CustomObject)
```
MarcelMue: For some reason the names we generate here are too long for e.g. our E2E sometimes.
Not entirely certain why this happens or if it's just an E2E problem.
Screenshot from 2019-09-18 20-08-12

corest: Crap
Maybe it needs to be shortened to IC
That’s anyway going to be removed with clusterapi

MarcelMue: There is only one per CF stack, right? 
Can we just name it `ingress-<clusterID>-internal`?

corest: That should be fine, yes

corest: I don't get this. Is there some crazy stuff on CI only? LB name for internal ingress is `ingress-<cluster_id>-internal`, which is <32

MarcelMue: Yes, I noticed this as well. Checking. I think the issue is that the clusterID in ci is way longer.

I will fix this, no need to worry for you.

PR #1943, opened 2019-09-18
Refactor whitelisting with private API whitelist

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6924

After sig-operator/security syncs, I've decided to add whitelisting for internal LB as well. 

1. With this PR I'll just bring it into WIP controllers and refactor a bit existing whitelisting structures on controller level

---

further work:
Add separate security group for internal LB
  - if private whitelisting enabled - allow only tenant cluster cidr + configured subnets
  - if private whitelisting disabled - allow all the private networks


corest:
I'd love to have those two whitelists better structured, but it would require refactoring all the existing controllers. 

That's why private LB mostly duplicated, only at the controller level I've added a bit of context regarding public/private whitelisting.

corest:
Before moving even more deep towards this direction, I need some thoughts towards this direction

service/controller/legacy/v30/adapter/guest_security_groups_test.go
```diff
@@ -188,14 +188,16 @@ func TestAdapterSecurityGroupsRegularFields(t *testing.T) {
...
+		description                   string
+		customObject                  v1alpha1.AWSConfig
+		apiWhitelistingEnabled        bool
+		apiWhitelistSubnets           string
+		elasticIPs                    []*ec2.Address
+		hostClusterCIDR               string
+		expectedError                 bool
+		expectedRules                 []securityGroupRule
+		privateAPIWhitelistingEnabled bool
+		privateAPIWhitelistSubnets    string
```
pipo02mix: Is this var used? has sense to create a test case for private too?

corest: Test will be there with actual usage in code. For now just wiring things

main.go
```diff
@@ -152,8 +152,10 @@ func mainE(ctx context.Context) error {
...
-	daemonCommand.PersistentFlags().Bool(f.Service.Installation.Guest.Kubernetes.API.Security.Whitelist.Enabled, false, "Enable or disable guest cluster k8s API whitelisting.")
-	daemonCommand.PersistentFlags().String(f.Service.Installation.Guest.Kubernetes.API.Security.Whitelist.SubnetList, "", "Subnet list for guest cluster k8s API whitelisting.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Installation.Guest.Kubernetes.API.Security.PrivateWhitelist.Enabled, false, "Enable or disable guest cluster k8s private API whitelisting.")
+	daemonCommand.PersistentFlags().String(f.Service.Installation.Guest.Kubernetes.API.Security.PrivateWhitelist.SubnetList, "", "Subnet list for guest cluster k8s private API whitelisting.")
+	daemonCommand.PersistentFlags().Bool(f.Service.Installation.Guest.Kubernetes.API.Security.Whitelist.Enabled, false, "Enable or disable guest cluster k8s public API whitelisting.")
+	daemonCommand.PersistentFlags().String(f.Service.Installation.Guest.Kubernetes.API.Security.Whitelist.SubnetList, "", "Subnet list for guest cluster k8s public API whitelisting.")
```
kopiczko: How about having `Whitelist.Public` and `Whitelist.Private`?

service/controller/clusterapi/v30/adapter/spec.go
```diff
@@ -45,10 +45,10 @@ const (
 	httpsPort = 443
 )
 
-// APIWhitelist defines guest cluster k8s api whitelisting.
+// APIWhitelist defines guest cluster k8s public/private api whitelisting.
 type APIWhitelist struct {
-	Enabled    bool
-	SubnetList string
+	Private Whitelist
+	Public  Whitelist
```
kopiczko: Don't we need to update CF templates now?

main.go
corest: Refactored flags with this Private/Public split

service/controller/clusterapi/v30/adapter/spec.go
corest: The template didn't change. It just receives rules, which are processed before passing into template

PR #1951, opened 2019-09-20
Duplicate etcd record backport to v29patch1

MarcelMue:
Porting the changes made in https://github.com/giantswarm/aws-operator/pull/1941 inside the legacy controller from `v30` back to `v29patch1`

cc @corest


MarcelMue:
The change is a fix for routing to etcd from the controlplane. 
Previously It was made to only be accessible from the private hosted zone inside the tenant cluster. But we need to have access from the controlplane for monitoring and some etcd backups. Therefor this PR adds a second Recordset to ensure this again.

PR #1955, opened 2019-09-23
Add internal whitelisting implementation

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6924

If there is no whitelisting - allow all private subnets
If there is whitelisting - allow control-plane cidr, tenant cidr + configured subnets

### DIFF
--- a/service/controller/clusterapi/v30/resource/tccp/testdata/case-0-basic-test.golden
+++ b/service/controller/clusterapi/v30/resource/tccp/testdata/case-0-basic-test.golden
@@ -214,7 +214,7 @@ Resources:
       LoadBalancerName: 8y5ck-api-internal
       Scheme: internal
       SecurityGroups:
-        - !Ref MasterSecurityGroup
+        - !Ref APIInternalELBSecurityGroup
       Subnets:
         - !Ref PrivateSubnetEuCentral1b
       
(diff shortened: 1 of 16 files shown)


service/controller/clusterapi/v30/resource/tccp/testdata/case-0-basic-test.golden
```diff
@@ -528,6 +528,37 @@ Resources:
       Tags:
         - Key: Name
           Value: 8y5ck-etcd-elb
+  APIInternalELBSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: 8y5ck-internal-api
+      VpcId: !Ref VPC
+      SecurityGroupIngress:
+
```
MarcelMue: Been thinking about it but the rendering here doesn't make my eyes particularly happy.
It's fine for now as it is consistent at least, but I might try and play around with it in a follow up.

corest: I’m not much into details on this. Can I improve that somehow?

MarcelMue: I played around briefly but can't make it better - so all good.

PR #1957, opened 2019-09-23
Add internal LB in ASG too

pipo02mix:
As we have an internal ingress controller too, we need to add it into the auto scaling group, otherwise, LB instances will be empty

### DIFF
--- a/service/controller/legacy/v29patch1/templates/cloudformation/tccp/auto_scaling_group.go
+++ b/service/controller/legacy/v29patch1/templates/cloudformation/tccp/auto_scaling_group.go
@@ -20,6 +20,7 @@ const AutoScalingGroup = `
       LaunchConfigurationName: !Ref {{ $v.ASGType }}LaunchConfiguration
       LoadBalancerNames:
         - !Ref IngressLoadBalancer
+        - !Ref IngressInternalLoadBalancer
       HealthCheckGracePeriod: {{ $v.HealthCheckGracePeriod }}
       MetricsCollection:
         - Granularity: "1Minute"
--- a/service/controller/legacy/v29patch1/templates/cloudformation/tccp/record_sets.go
+++ b/service/controller/legacy/v29patch1/templates/cloudformation/tccp/record_sets.go
@@ -37,7 +37,7 @@ const RecordSets = `
       Name: 'api.{{ $v.ClusterID }}.k8s.{{ $v.BaseDomain }}.'
       HostedZoneId: !Ref 'InternalHostedZone'
       Type: A
-  EtcdInternalRecordSet:
+  EtcdRecordSet:
     Type: AWS::Route53::RecordSet
     Properties:
       AliasTarget:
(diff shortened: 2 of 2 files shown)


pipo02mix:
Tested manual, I will check in gauss now

service/controller/legacy/v29patch1/templates/cloudformation/tccp/record_sets.go
```diff
@@ -47,7 +47,7 @@ const RecordSets = `
       Name: '{{ $v.EtcdDomain }}.'
       HostedZoneId: !Ref 'InternalHostedZone'
       Type: A
-  EtcdRecordSet:
+  EtcdExternalRecordSet:
     Type: AWS::Route53::RecordSet
```
pipo02mix: I have realized it was breaking the upgrade as it was trying to update the record set with same name but in different zone

```
[Tried to create resource record set [name='etcd.0cuut.k8s.ginger.eu-central-1.aws.gigantic.io.', type='A'] but it already exists]
```
We need to keep the internal the same

service/controller/legacy/v29patch1/templates/cloudformation/tccp/record_sets.go
```diff
@@ -37,7 +37,7 @@ const RecordSets = `
       Name: 'api.{{ $v.ClusterID }}.k8s.{{ $v.BaseDomain }}.'
       HostedZoneId: !Ref 'InternalHostedZone'
       Type: A
-  EtcdInternalRecordSet:
+  EtcdRecordSet:
```
MarcelMue: Can we keep this as `EtcdInternalRecordSet` Or will this also break?

service/controller/legacy/v29patch1/templates/cloudformation/tccp/auto_scaling_group.go
```diff
@@ -20,6 +20,7 @@ const AutoScalingGroup = `
       LaunchConfigurationName: !Ref {{ $v.ASGType }}LaunchConfiguration
       LoadBalancerNames:
         - !Ref IngressLoadBalancer
+        - !Ref IngressInternalLoadBalancer
```
MarcelMue: We should then port this to `v30` as well - I will take care.

service/controller/legacy/v29patch1/templates/cloudformation/tccp/record_sets.go
pipo02mix: the problem is the existing `8.4.0` have the `EtcdRecordSet ` pointing to the internal, the order of CF is adding first the other recordset and fails because it exists. Not sure how to solve it without keeping the name

MarcelMue: Okay, I see. Let's keep the name for now and see if we can solve it otherwise later on.

service/controller/legacy/v29patch1/templates/cloudformation/tccp/auto_scaling_group.go
pipo02mix: I can do it in a follow-up PR

1.

pipo02mix:
Tested on ginger, it works :)

PR #1959, opened 2019-09-24
Change recordet names

pipo02mix:
I have realized it was breaking the upgrade (from `8.4.0` to `8.4.1`) as it was trying to update the recordset with the same name but in a different zone
```
[Tried to create resource record set [name='etcd.0cuut.k8s.ginger.eu-central-1.aws.gigantic.io.', type='A'] but it already exists]
```
We need to keep the internal the same


pipo02mix:
testing in ginger

PR #1963, opened 2019-09-30
migrate classic ELB to network ELB for node pools

calvix:
changes required to migrate classic ELB to network ELB for node pools
few notes:

* ConnectionIdleTimeout for classic ELB is no longer configurable for network LB and its set to 300s
* Security groups are not supported for load balancers with type 'network' 
* For TCP health checks, the supported values are 10 and 30 seconds. ( set to 10s)
* Custom health check timeouts are not supported for health checks for target groups with the TCP protocol 

towards  https://github.com/giantswarm/giantswarm/issues/6883

TODO:
- [x] clean code for security groups 
- [x] sort out security group for internal and public API LB


calvix:
There is still some small workaround internal API security group but rest should be reviewable so please take a look

service/controller/clusterapi/v31/resource/tccpingresstargetgroups/create.go
```diff
@@ -0,0 +1,87 @@
...
+		ingressInsecureTargetGroup, err := cloudFormation.GetOutputValue(outputs, IngressInsecureTargetGroupIDs)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		ingressSecureTargetGroup, err := cloudFormation.GetOutputValue(outputs, IngressSecureTargetGroupIDs)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		cc.Status.TenantCluster.TCCP.IngressTargetGroupIDs = []string{ingressInsecureTargetGroup, ingressSecureTargetGroup}
+		fmt.Printf("IngressLoadbalancerTargetGroups: XXSSAA  %#v\n", cc.Status.TenantCluster.TCCP.IngressTargetGroupIDs)
```
tuommaki: Debug leftover? Zap! :wink:

service/controller/clusterapi/v31/cloudconfig/tccp.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	g8sv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
 	"github.com/giantswarm/certs"
-	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v_4_8_0"
+	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v_4_9_0"
```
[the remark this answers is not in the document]
calvix: Yes

service/controller/clusterapi/v31/cloudconfig/tccp.go
```diff
@@ -54,17 +54,27 @@ func (t *TCCP) Render(ctx context.Context, cr cmav1alpha1.Cluster, clusterCerts
...
+		// this is due to problem that network lb does not change source IP
+		// and packet sent from master via network load-balancer to master
+		// has same source and destination IP and the operating system will not forward the packet back to LB
+		// result is that connection does not work
+		params.Cluster.Kubernetes.API.Domain = Localhost
+		params.Cluster.Etcd.Domain = Localhost
 		params.DisableEncryptionAtREST = true
 		// Ingress controller service remains in k8scloudconfig and will be
 		// removed in a later migration.
+		params.DisableCalico = true
```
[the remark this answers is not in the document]
calvix: bad comment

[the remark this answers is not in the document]
calvix: fixed

service/controller/clusterapi/v31/resource/tccp/template/template_main_load_balancers.go
```diff
@@ -4,132 +4,159 @@ const TemplateMainLoadBalancers = `
...
+      HealthCheckEnabled: true
+      HealthCheckIntervalSeconds: {{ $v.ELBHealthCheckInterval }}
+      HealthCheckPort: {{ .PortInstance }}
+      HealthCheckProtocol: TCP
+      HealthyThresholdCount: {{ $v.ELBHealthCheckHealthyThreshold }}
+      Name: {{ .TargetGroupName }}
+      Port: {{ .PortInstance }}
+      Protocol: TCP
+      TargetGroupAttributes:
+      - Key: proxy_protocol_v2.enabled
```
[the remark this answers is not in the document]
calvix: No, we had proxy protocol before and this is just a new way how to define it. (Before there was a special policy attached to LB)

I tested ingress and it works as expected. nginx IC has proxy-protocol enabled and LB is sending traffic as expected.

service/controller/clusterapi/v31/resource/tccpingresstargetgroups/create.go
calvix: yep, leftover :p

service/controller/clusterapi/v31/cloudconfig/tccp.go
```diff
@@ -54,17 +54,27 @@ func (t *TCCP) Render(ctx context.Context, cr cmav1alpha1.Cluster, clusterCerts
...
 		params.DisableIngressControllerService = false
 		params.EtcdPort = key.EtcdPort
 		params.Extension = &MasterExtension{
 			awsConfigSpec: cmaClusterToG8sConfig(t.config, cr, labels),
 			baseExtension: baseExtension{
-				cluster:       cr,
-				encrypter:     t.config.Encrypter,
-				encryptionKey: cc.Status.TenantCluster.Encryption.Key,
+				calicoMTU:      params.Cluster.Calico.MTU,
```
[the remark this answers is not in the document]
calvix: Not sure what would need to be discussed, We had configurable MTU since a long time I am jsut passing the value to the right place here.

I know that we want to have it configurable per cluster in the future as some customer asks for it.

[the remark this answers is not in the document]
tuommaki: No, nothing _to_ be discussed, but I was referring to that discussion on specific customer needs, but yeah. All good here. :+1:

tuommaki:
Will you ping again after final modifications and I glance them over? So far this lgtm.

calvix:
@tuommaki Cool, there should be one more change to move internal API security rules to the master security group and that should be it. I will ping again.

service/controller/clusterapi/v31/resource/tccpingresstargetgroups/create.go
```diff
@@ -0,0 +1,86 @@
+package tccpingresstargetgroups
```
xh3b4sd: I am not sure what your reasoning was but we have already a resource to fetch the TCCP stack outputs which are then put into the controller context. See https://github.com/giantswarm/aws-operator/blob/master/service/controller/clusterapi/v31/resource/tccpoutputs/create.go. I think this can simply be extended so we do not need to have another resource for it.

service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
```diff
@@ -3,6 +3,10 @@ Description: Tenant Cluster Control Plane Cloud Formation Stack.
 Outputs:
   DockerVolumeResourceName:
     Value: DockerVolume8Y5CK78968
+  IngressInsecureTargetGroupsID:
+    Value: !Ref IngressInsecureTargetGroup
+  IngressSecureTargetGroupsID:
+    Value: !Ref IngressSecureTargetGroup
```
xh3b4sd: Is it absolutely necessary to put these information into the stack outputs? We also have another approach where we have a dedicated resource to fetch information from the AWS API separately. Reason for this in several cases is to have information accessible even during upgrades. Not sure which approach would be best here.

service/controller/clusterapi/v31/resource/tccpoutputs/create.go
```diff
@@ -3,7 +3,6 @@ package tccpoutputs
 import (
 	"context"
 	"fmt"
-
```
xh3b4sd: ❤️

service/controller/clusterapi/v31/resource/tccpingresstargetgroups/create.go
calvix: The problem here was that the resource also needs to be used in machineDeployment controller and this value was not propagated, I previously had it in this resource and it was nil at the place where I need to use it.

xh3b4sd: I see. You want the TCCP outputs in the MachineDeployment controller. Then I would configure the `toCluster` function in the `tccpoutputs` resource, extend it with the two new values and wire it in the MachineDeployment controller. This would be more straight forward because we then only have one resource each for all the outputs. I feel like having a different resource increases the complexity of the approach and we do not need to do this.

calvix:
@giantswarm/team-firecracker guys can you take a another look? this should be the final version

service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
```diff
@@ -457,6 +504,12 @@ Resources:
         FromPort: 2379
         ToPort: 2379
         CidrIp: 10.1.0.0/16
+      -
+        Description: Allow traffic from Cluster CIDR to 2379 for network-lb health checks.
```
xh3b4sd: Does `Cluster` mean Tenant Cluster? Could we clarify this? We also talk about Control Plane in the other rules.

calvix: Good point I will improve comments.

PR #1967, opened 2019-10-02
Pick up tenant cluster OIDC config from Cluster CR

tuommaki:
Starting with v31 pick up tenant cluster OIDC config directly from Cluster CR's
provider extension instead of `aws-operator's` configuration. This allows to
use cluster specific OIDC config instead of installation wide.

Towards https://github.com/giantswarm/giantswarm/issues/5588


tuommaki:
@pipo02mix What's the easiest / best way to test this?

tuommaki:
> I guess there is some extra wiring in kubernetesd or friends present for backward compatibility.

Since Node Pools release is a breaking release (no upgrades from earlier clusters), what would need backwards compatibility?

pipo02mix:
> What's the easiest / best way to test this?

can we create a test cluster for customer so they can verify it? I guess we can do in our envs and give them the api endpoint

tuommaki:
> can we create a test cluster for customer so they can verify it?

Since this is bundled with Node Pools, not yet (need to finish api & clients first), but I'll check the resulting apiserver flags from master node and verify it that way and then we test this later with customer when Node Pools are ready for customer preview.

tuommaki:
Tested this on `gauss`:
```
      apiserver    
      --oidc-client-id=oidc-clientid
      --oidc-issuer-url=https://oidc-issuer:8082/foo/bar/issue
      --oidc-username-claim=oidc-username     
      --oidc-groups-claim=oidc-groups
```

PR #1975, opened 2019-10-09
Change image.repository to image.name

rossf7:
See https://github.com/giantswarm/fmt/pull/75. Updating since aws-operator should be SotA (State of the Art).


rossf7:
Tested on `ginger`.

rossf7:
Tested manually. Going with this.

PR #1986, opened 2019-10-15
helm/aws-operator: fix rbac for {clusters,machinedeployments}/status

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/7276

We noticed that when testing flattened version. We don't know why it
works for non-flattened one.


MarcelMue:
Should also be added for machinedeployments.

PR #2007, opened 2019-10-23
Add IAMRoleManager role for kiam

corest:

Calling it `IAMRoleManager` to not make specific mention of the kiam

### DIFF
--- a/service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
+++ b/service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
@@ -69,6 +69,30 @@ Resources:
             Condition:
               StringEquals:
                 autoscaling:ResourceTag/giantswarm.io/cluster: "8y5ck"
+  IAMRoleManagerRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      RoleName: 8y5ck-IAMRoleManager-Role
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          Effect: "Allow"
+          Principal:
+            AWS: !Ref MasterRole.Arn
+          Action: "sts:AssumeRole"
+  IAMRoleManagerRolePolicy:
+    Type: "AWS::IAM::Policy"
+    Properties:
+      PolicyName: 8y5ck-IAMRoleManager-Policy
+      Roles:
+        - Ref: "IAMRoleManagerRole"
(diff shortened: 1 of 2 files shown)


pipo02mix:
I am hardening the policy we define for customers AWS accounts. I would really appreciate if we can follow a name pattern so we can define regex for our GS roles. 

What about `gs-` prefix?

corest:
@pipo02mix I'm fine with prefixes. But was it discussed anywhere? Documented?

pipo02mix:
No, it was my idea after discussions with customers.

corest:
@giantswarm/sig-operator ptal

corest:
Later use. This one will be used by default app `kiam` (tested)
And next PR will add the same `Route53Manager` role for default app `external-dns`

service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
```diff
@@ -69,6 +69,30 @@ Resources:
             Condition:
               StringEquals:
                 autoscaling:ResourceTag/giantswarm.io/cluster: "8y5ck"
+  IAMManagerRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      RoleName: 8y5ck-IAMManager-Role
```
kopiczko: I guess it should be `{{ $v.ClusterID }}-IAMManager-Role`

service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
```diff
@@ -69,6 +69,30 @@ Resources:
...
+        Version: "2012-10-17"
+        Statement:
+          Effect: "Allow"
+          Principal:
+            AWS: !GetAtt MasterRole.Arn
+          Action: "sts:AssumeRole"
+  IAMManagerRolePolicy:
+    Type: "AWS::IAM::Policy"
+    Properties:
+      PolicyName: 8y5ck-IAMManager-Policy
```
kopiczko: And `{{ $v.ClusterID }}-IAMManager-Policy`.

service/controller/clusterapi/v31/resource/tccp/testdata/case-0-basic-test.golden
kopiczko: It's a test file 🤦‍♂

PR #2015, opened 2019-10-24
Removes chart aws-operator-chart from legacy branch

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/7312

The `aws-operator-chart` chart is only used by operators deployed with GitHub, so is not needed in the `legacy` branch.


JosephSalisbury:
See https://github.com/giantswarm/giantswarm/issues/7312#issuecomment-546335659, this needs more work for E2E tests to be green.

PR #2030, opened 2019-10-29
Route53 recordsets cleanup resource

corest:

As we now bring external-dns into default apps, any dns records, managed by external-dns, blocks CF deletion (because of hosted zone can't be deleted if there are any records except of NS and SOA).

This resource deletes all the records, which are not managed by CF on both private and public hosted zones


service/controller/resource/cleanuprecordsets/delete.go
```diff
@@ -0,0 +1,115 @@
...
+				route53Changes = append(route53Changes, route53Change)
+
+				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("found non-managed record set %#q in hosted zone %#q for tenant cluster %#q", *rr.Name, *hostedZone.Id, key.ClusterID(&cr)))
+			}
+		}
+
+		if len(route53Changes) > 0 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleteting non-managed record sets in hosted zone %#q for tenant cluster %#q", *hostedZone.Id, key.ClusterID(&cr)))
+
+			chageRecordSetInput := &route53.ChangeResourceRecordSetsInput{
```
rossf7:
```suggestion
			changeRecordSetInput := &route53.ChangeResourceRecordSetsInput{
```

service/controller/resource/cleanuprecordsets/delete.go
```diff
@@ -0,0 +1,115 @@
...
+		return microerror.Mask(err)
+	}
+
+	var hostedZones []*route53.HostedZone
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding hosted zones for tenant cluster %#q", key.ClusterID(&cr)))
+
+		hostedZonesInput := &route53.ListHostedZonesByNameInput{}
+
+		o, err := cc.Client.TenantCluster.AWS.Route53.ListHostedZonesByName(hostedZonesInput)
```
kopiczko: IIRC there is some paging. But I may be wrong.

service/controller/resource/cleanuprecordsets/delete.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		for _, zone := range o.HostedZones {
+			baseDomain := fmt.Sprintf("%s.", key.TenantClusterBaseDomain(cr))
```
kopiczko: This doesn't look right to me. For new style clusters we don't have `k8s.*` zone. Each cluster has separate `CLUSTER_ID.k8s.*` zone in the BYOC account. This zone should be deleted by the CF stack. If dynamic DNS is configured correctly (i.e. creating records in the new zone) then this resources should not be needed.

corest: I didn’t get this

CF can’t delete hosted zone if there are records not from CF. Who does cleanup then?

kopiczko: > CF can’t delete hosted zone if there are records not from CF.

I didn't know this. Then we need the resource.

But still we shouldn't create any records in `k8s` zone. We should create them in `CLUSTER_ID.k8s` zone.

corest: I’m again not sure about what you’re talking here. Resources created in clusterID.k8s zones (outside the stack) 
This resource cleane those resources in clusterID.k8s zone

service/controller/resource/cleanuprecordsets/delete.go
corest: Will check this

service/controller/resource/cleanuprecordsets/delete.go
```diff
@@ -0,0 +1,115 @@
...
+		return microerror.Mask(err)
+	}
+	cc, err := controllercontext.FromContext(ctx)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	var hostedZones []*route53.HostedZone
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("finding hosted zones for tenant cluster %#q", key.ClusterID(&cr)))
```
kopiczko: We should have cluster ID injected in the logs one way or another. I wouldn't print it in log messages.

service/controller/key/cluster.go
```diff
@@ -115,6 +115,17 @@ func MasterInstanceType(cluster v1alpha1.Cluster) string {
 	return clusterProviderSpec(cluster).Provider.Master.InstanceType
 }
 
+func ManagedRecordSets(cluster v1alpha1.Cluster) []string {
+	tcBaseDomain := TenantClusterBaseDomain(cluster)
+	return []string{
+		fmt.Sprintf("%s.", tcBaseDomain),
+		fmt.Sprintf("\\052.%s.", tcBaseDomain), // \\052 - `*` wildcard record
```
kopiczko: Is wildcard record created by external-dns?

service/controller/key/cluster.go
```diff
@@ -115,6 +115,17 @@ func MasterInstanceType(cluster v1alpha1.Cluster) string {
...
+		fmt.Sprintf("api.%s.", tcBaseDomain),
+		fmt.Sprintf("etcd.%s.", tcBaseDomain),
+		fmt.Sprintf("internal-api.%s.", tcBaseDomain),
```
kopiczko: Also can't we list services and check annotations to discover entries that need to be removed? That way we won't need to extend this list in case we add something else.

service/controller/resource/cleanuprecordsets/delete.go
corest:
```
Route 53 returns up to 100 items in each response
```

It should be fine with this. We don't have hundreds of clusters

service/controller/key/cluster.go
corest: Nope, wildcard created by aws-operator. It is the same as in azure-operator. external-dns doesn't support multiple records per services/ingress

service/controller/key/cluster.go
corest: I've looked into that option. But we anyway need to know hosted zone to remove the record. Also there might be other apps adding records based on different annotations. So from that point of view we would need to extend list with annotations in the same way as  list of managed recordsets

service/controller/resource/cleanuprecordsets/delete.go
corest: What is wrong with this way of logging? How can I do this in a different way?

kopiczko: Just skip ` for tenant cluster %#q` part of the log.

service/controller/key/cluster.go
kopiczko: But you can check if the cluster base domain matches with the value of annotation. Wouldn't that be sufficient?

service/controller/key/cluster.go
kopiczko: Won't it be deleted by the CF stack then?

corest: I meant it is created by CF stack in aws-operator. So it will be deleted by CF stack.

service/controller/key/cluster.go
corest: But you need to know about those annotations. Fo external-dns it will be something like `external-dns-foo`,  so you'll look into this in operator. Then we add later another app which makes it 'other-app-foo-annotation' - what to do - update operator again? Cleaning route53 directly makes it app independent. Relying on annotations brings scope we don't know how will look later

service/controller/resource/cleanuprecordsets/delete.go
corest: But in which way `finding hosted zones` message is informative then?

kopiczko: Cluster ID is part of the reconciled object name which is already logged.

service/controller/key/cluster.go
kopiczko: Then it shouldn't be listed here correct?

service/controller/key/cluster.go
kopiczko: Okay. I don't know enough about external-dns but I bet it's discoverable. Otherwise how could external-dns work. One way or another it's not independent. I won't block on that though. Let's sort the wildcard and go.

service/controller/key/cluster.go
corest: The list here is the list of managed by operator records. Not by apps. So cleanup delete all the others which are not in list.

kopiczko: Ack. That makes sense. Thanks for bearing with me.

service/controller/key/cluster.go
corest: It is out of external-dns (or any other app) scope. It is just about bringing recordsets to the state, managed by CF

PR #2047, opened 2019-10-31
main, service: move vesionbundle to pkg/project

kopiczko:

I relalized versionbundle package is only used in two files:

- main.go
- service/service.go

Resource sets should use `porject.BundleVersion()` anyway.

Therefore I think it makes sense to move all that stuff to the project package.


service/service.go
```diff
@@ -317,7 +317,7 @@ func New(config Config) (*Service, error) {
 			Name:           project.Name(),
 			Source:         project.Source(),
 			Version:        project.Version(),
-			VersionBundles: versionbundle.NewSlice(),
+			VersionBundles: []versionbundle.Bundle{project.NewVersionBundle()},
```
kopiczko: `NewSlice` is dropped because we don't have import conflict anymore.

PR #2051, opened 2019-10-31
Remove component updates from changelog

tfussell:
I didn't have a clear picture of the final versioning we would use for the node pools release when I originally added these. `aws-operator` `v5.5.0` released in GS v9.0 already includes these updates so I think it doesn't make sense to have them here as well.


kopiczko:
FYI this changelog won't show up till you release next version.

PR #2057, opened 2019-11-04
vendor: update versionbundle

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/7511


kopiczko:
Tested on gauss.

PR #2065, opened 2019-11-13
Add SHA verification of the ignition

corest:


corest:
It makes sense to make this verification as a part of cloudconfig operator

PR #2077, opened 2019-11-25
Use apps/v1 for deployment

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7355

I'm unclear why its started happening now but this fixes deployment of the PR from @anvddriesch  https://github.com/giantswarm/aws-operator/pull/2068

Since we need to upgrade the API version for K8s 1.16 this updates the helm chart.

```
status:
  appVersion: ""
  reason: 'Upgrade "aws-operator-remove-adapter" failed: Deployment.apps "aws-operator-remove-adapter"
    is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":"aws-operator",
    "version":"7.0.0-8ffb529e4a11578553a7f802bfc8fdf20cd4deb3"}: `selector` does not
    match template `labels`'
```

### DIFF
--- a/helm/aws-operator/templates/deployment.yaml
+++ b/helm/aws-operator/templates/deployment.yaml
@@ -12,7 +12,7 @@ spec:
   selector:
     matchLabels:
       app: {{ .Values.project.name }}
-      version: {{ .Values.project.version }
+      version: {{ .Values.project.version }}
   strategy:
     type: RollingUpdate
   template:
--- a/helm/aws-operator/templates/deployment.yaml
+++ b/helm/aws-operator/templates/deployment.yaml
@@ -12,6 +12,7 @@ spec:
   selector:
     matchLabels:
       app: {{ .Values.project.name }}
+      version: {{ .Values.project.version }
   strategy:
     type: RollingUpdate
   template:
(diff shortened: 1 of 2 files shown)


helm/aws-operator/templates/deployment.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
 spec:
   replicas: 1
   revisionHistoryLimit: 3
+  selector:
+    matchLabels:
+      app: {{ .Values.project.name }}
+      version: {{ .Values.project.version }
```
sslavic: Good point to have version included. I've been doing these kind of chances in other repos/projects, as preparation for k8s 1.16 where extesnsion/v1beta1 is dropped - have to review them again myself, but please check as well when reviewing dep upgrade PRs, if version or more is needed in selector there too.

rossf7: @sslavic Will do I noticed this when debugging deployment issues on `ginger`. For VOO operators I think we do need both labels.

helm/aws-operator/templates/deployment.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
 spec:
   replicas: 1
   revisionHistoryLimit: 3
+  selector:
+    matchLabels:
+      app: {{ .Values.project.name }}
+      version: {{ .Values.project.version }}
```
rossf7: The selector has to be specified when using `apps/v1` and this is what tiller was complaining about. 

I think for VOO operators we need to include the version label.

rossf7:
Tested on `ginger`.

PR #2086, opened 2019-12-06
Copy v30 to v30patch1

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/7302


tfussell:
We generally merge version bundle copy/wiring without approval, but I wanted to increase visibility since this is AFAIK our first release on the `thiccc` branch.

PR #2090, opened 2019-12-13
legacy: upgrade to k8s 1.16

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7675


sslavic:
@rossf7 I've reverted all debug commits, and switched to statusresource branch with /status init fix https://github.com/giantswarm/statusresource/pull/45; e2e tests are still passing and running in parallel again (see [circleci](https://circleci.com/workflow-run/ed9fc906-f0c0-4f20-8abd-7677165aa666))

rossf7:
> I've reverted all debug commits, and switched to statusresource branch with /status init fix giantswarm/statusresource#45; e2e tests are still passing and running in parallel again

@sslavic From my POV this looks good to go. To be safe I think we could test the `statusresource` changes in azure-operator and / or kvm-operator.

integration/setup/config.go
```diff
@@ -63,6 +65,10 @@ func NewConfig() (Config, error) {
 	{
 		c := k8sclient.ClientsConfig{
 			Logger: logger,
+			SchemeBuilder: k8sclient.SchemeBuilder{
+				corev1alpha1.AddToScheme,
+				providerv1alpha1.AddToScheme,
```
kopiczko: Don't we need `infrastructurev1alpha2.AddToScheme` here?

integration/setup/tenant_cluster.go
```diff
@@ -71,6 +71,11 @@ func EnsureTenantClusterDeleted(ctx context.Context, id string, config Config, w
 
 	config.Logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleting tenant cluster %#q", id))
 
+	err = config.Guest.Initialize()
```
kopiczko: Is this something new?

integration/setup/config.go
rossf7: No because this is legacy. All we need from the CAPI types is to create the new AWSCluster CRD so the `ipam` resource does not error.

`err := config.CPCRDClient.EnsureCreated(ctx, infrastructurev1alpha2.NewAWSClusterCRD(), b)`

integration/setup/tenant_cluster.go
rossf7: Without this the guest framework clients don't work. It needs to read the secret generated by cert-operator in the minikube cluster.


https://github.com/giantswarm/e2e-harness/blob/b2048d8645c15ac583a270c57b95855426e48958/pkg/framework/guest.go#L105

kopiczko:
I think changelog entry would be useful.

Gopkg.toml
```diff
@@ -33,96 +33,96 @@ required = [
...
 
 [[constraint]]
-  branch = "master"
   name = "github.com/giantswarm/randomkeys"
+  branch = "master"
 
 [[constraint]]
-  branch = "master"
   name = "github.com/giantswarm/statusresource"
+  branch = "fix-status-init"
```
rossf7: TODO. Needs https://github.com/giantswarm/statusresource/pull/45.

sslavic:
e2e tests failed

> The maximum number of VPCs has been reached. (Service: AmazonEC2; Status Code: 400; Error Code: VpcLimitExceeded; Request ID: 7b6fcf99-1aeb-4fc0-b547-b60d59181950)

rossf7:
@sslavic Argh this is the never ending story.

See https://github.com/giantswarm/giantswarm/issues/8141 the CI cleaner bot is not working. Deleting VPCs manually to get under the quota is probably the best bet.

sslavic:
@rossf7 there are `cp-peer-ci-` prefixed CF stacks - maybe better to delete them instead of VPCs directly.

rossf7:
@sslavic Sorry yes exactly right. Best to delete the stacks not the VPCs!

sslavic:
Cleaned up manually previously left-over `cp-peer-ci-` prefixed stacks.
Added teardown logic to have those automatically deleted in the future.

sslavic:
With additional teardown logic, e2eTestPRRecreateCluster started failing, looking into that.

sslavic:
Fixed teardown, e2e tests are green again.

rossf7:
Going with this so we can **finally**  close giantswarm/giantswarm#7675. @sslavic thanks for all the help with this!

@xh3b4sd @tuommaki PTAL when you're back.

PR #2098, opened 2019-12-20
Improve TCCP CF template rendering test

tuommaki:
Add some variables for TCCP CF template golden file rendering test.


service/controller/resource/tccp/template/template_main_iam_policies.go
```diff
@@ -84,7 +84,7 @@ const TemplateMainIAMPolicies = `
           Effect: "Allow"
           Action: "sts:AssumeRole"
           Resource: "*"
-{{- if $v.Route53Enabled -}}
+{{- if $v.Route53Enabled}}
```
tuommaki: This was the crucial bit here

PR #2099, opened 2019-12-23
push-and-use-china-registry

calvix:
towards giantswarm/giantswarm#7435

push and use china image in china region


calvix:
work fine, my only question is should i bump the version ?

kopiczko:
> work fine, my only question is should i bump the version ?

Do you need to release it?

helm/aws-operator/values.yaml
```diff
@@ -7,9 +7,12 @@ flag:
       labelselector:
         version: ""
 image:
-  registry: "quay.io"
   name: "giantswarm/aws-operator"
   tag: "[[ .Version ]]"
+Installation:
+  V1:
+    Registry:
+      Domain: quay.io
```
kopiczko: Is that needed? The default is set here https://github.com/giantswarm/installations/blob/91d748669d9523f53d285f3cafde1812d7165261/default-draughtsman-configmap-values.yaml#L718

kopiczko:
BTW here's how it's done in net-exporter https://github.com/giantswarm/net-exporter/pull/77/files

helm/aws-operator/values.yaml
calvix: Yes, for e2e test this is needed as the e2e installations tools dont use draightsman and dont have its defaults

kopiczko: Understood. Then we should unify the approach with https://github.com/giantswarm/net-exporter/pull/77/files

calvix:
probably don't need specific release as any real change will pick this up and thats fine

kopiczko:
Yes, that's also what I thought.

helm/aws-operator/values.yaml
rossf7: The problem is cert-exporter and net-exporter are deployed by both draughtsman and app CRs. So `image.registry` is needed for app CRs.

aws-operator is deployed by app CR but for VOO so only the draughtsman values are used. So I think `.Values.Installation.V1.Registry.Domain` is correct.

I find this confusing and maybe supporting both structures is better and more future proof?

calvix: Yes, what @rossf7  said. I did that already to few places. I woudl leave it that way.

kopiczko: 👍

PR #2100, opened 2019-12-24
Fix default cloudconfig kubelet extra-args

corest:
In china clusters I've found `--pod-infra-container-image` was missing. Kubelet tried to use gcr repo which is blocked in China. `--pod-infra-container-image` flag added to kubelet extra flags in `Default()` function, but that function is never called actually. 

This PR fixes that.


corest:
I'm lost a bit here as what I see now in aws-operator repo looks like the release process was ignored. There is a tag for 8.0.2 and it is deployed everywhere. But there is master referencing tag 8.0.1-dev. I want to bring my version 8.0.3  and I've expected that master will have 8.0.3-dev after releasing 8.0.2 . But that is not true. 

Here I'm adding `8.3.0` anyway and then will add `8.0.4-dev` after merge. But looking into last PR, where @tuommaki brought back `8.0.1-dev` bundle into master, it looks like `8.0.2` is broken and shouldn't be used anywhere? Why then it is deployed everywhere?

tuommaki:
> Here I'm adding 8.0.3 anyway and then will add 8.0.4-dev after merge. But looking into last PR, where @tuommaki brought back 8.1.0-dev bundle into master, it looks like 8.0.2 is broken and shouldn't be used anywhere? Why then it is deployed everywhere?

Uhm, xmas & new year was and went but just some remarks:
* `8.0.0` and `8.0.1` tags were rubbish.
* `8.0.2` is expected to work. That's the Node Pools release (https://github.com/giantswarm/releases/blob/master/aws.yaml#L2)
* Our thinking was that bug fixes which become patch releases over old versions are patch version bumps and every new feature release would be at least a minor versions. That's why I bumped the dev version to `8.1.0-dev`.

I guess we need to get used to this and agree some common rules here as this is still new for most of us.

PR #2107, opened 2020-01-03
circleci: push to aliyun on tag; fix comments

kopiczko:


.circleci/config.yml
```diff
@@ -95,10 +96,13 @@ workflows:
...
           requires:
             - build
-          # Needed to trigger job only on merge to master.
           filters:
+            # Trigger the job on merge to master.
             branches:
               only: master
+            # Trigger the job also on git tag.
+            tags:
+              only: /^v.*/
```
kopiczko: @giantswarm/team-batman that's may be an issue also in other repositories.

PR #2132, opened 2020-02-03
Create ingress wildcard for internal resolution

paurosello:
If we want to resolver ingress domain from inside the cluster itself, we will need to add a wildcard to the ingress domain in the internal DNS records.

This issue appeared when trying to issue certificates with cert-manager.

More info in https://gigantic.slack.com/archives/C7865GLSY/p1580723650001300


corest:
The customer got into this issue. Why it was not merged? @paurosello

paurosello:
I noticed on Monday too :( maybe we can merge now before actually making a release

PR #2164, opened 2020-03-04
Change Deployment rollout strategy to Recreate

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/9173

We have seen cases where `MultipleOperatorsRunningSameBundleVersion` fires during an operator update. This is because during some time we have 2 pods (the old being terminated and the new being started up) running at the same time.

This PR changes the rollout strategy of Deployment so we don't run into this and make sure that same operator do not run twice.


fiunchinho:
Should we replicate this in all operators then?

kopiczko:
> Should we replicate this in all operators then?

For all flat operators. See the TODO here https://github.com/giantswarm/giantswarm/issues/9173

TheoBrigitte:
I though this was going to be controversial, but all good :)

Gonna replicate this on all operators then.

PR #2207, opened 2020-03-16
legacy 1.15: Generate AMI info

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8436

This code reads https://stable.release.core-os.net/amd64-usr/ to get a list of versions, parses `coreos_production_ami_all.json` for each one, and writes this to a multi-line JSON string so it can be loaded as needed to convert a Container Linux version and region into an AMI for masters and workers.


omegas27:
Would it be difficult to change your script to use flatcar instead or even better support both coreos and flatcar? Maybe you could have your main.go file have a flatcar ami fetcher and and coreos flatcar fetcher. Also, could we use aws api instead? Otherwise, LGTM

tfussell:
> Would it be difficult to change your script to use flatcar instead or even better support both coreos and flatcar? Maybe you could have your main.go file have a flatcar ami fetcher and and coreos flatcar fetcher. Also, could we use aws api instead? Otherwise, LGTM

I looked up `containerlinux`, `coreos`, and known AMIs in the AWS Console AMI search and didn't find the official images. Maybe this will work with the AWS API, but I feel like this works well enough for now. Supporting Flatcar should be quite easy as the versions transition seamlessly. It will just be a matter of looping over the two release URLs and combining them into one top-level map. I will do it in a separate PR because I don't want to hold this up.

tfussell:
> Let's get this merged into the correct branch - is `release-551` accurate?

I think so. I discussed in #dev yesterday whether we should chain PRs like A<-B<-C and merge C to B then B to A or B to A and then C to A so I'm trying the first way here.

PR #2211, opened 2020-03-18
Update public and private intermediate zones

corest:
Towards https://github.com/giantswarm/giantswarm/issues/8900


corest:
closing because of unsupported branch name :/

PR #2212, opened 2020-03-18
Update public and private intermediate zones

corest:
Towards giantswarm/giantswarm#8900


corest:
This will not work :/ https://github.com/giantswarm/giantswarm/issues/8900#issuecomment-601440127

PR #2230, opened 2020-03-26
'remove-encryption-from-master-volume'

calvix:
because of some unforesson issue we did not have properly set encryption for master volumes, and with removing o vault encryptor this is  render volumes  with encryption flag and causes trouble with update.

We will address this issue when we migrate  to ASG masters, as we migrate etcd volume via snapshot to new volume.

### DIFF
--- a/service/controller/resource/tccp/template/template_main_instance.go
+++ b/service/controller/resource/tccp/template/template_main_instance.go
@@ -26,9 +26,6 @@ const TemplateMainInstance = `
   {{ $v.Master.DockerVolume.ResourceName }}:
     Type: AWS::EC2::Volume
     Properties:
-{{ if eq $v.Master.EncrypterBackend "kms" }}
-      Encrypted: true
-{{ end }}
       Size: 50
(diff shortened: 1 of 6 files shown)


calvix:
currently created a cluster with a release version `11.1.2`

```
  DockerVolumeSB8X4B24D8:
    Type: AWS::EC2::Volume
    Properties:

      Size: 50
      VolumeType: gp2
      AvailabilityZone: eu-west-1a
      Tags:
      - Key: Name
        Value: sb8x4-docker
  EtcdVolume:
    Type: AWS::EC2::Volume
    Properties:

      Size: 100
      VolumeType: gp2
      AvailabilityZone: eu-west-1a
      Tags:
      - Key: Name
        Value: sb8x4-etcd
  LogVolume:
    Type: AWS::EC2::Volume
    Properties:

      Size: 100
      VolumeType: gp2
      AvailabilityZone: eu-west-1a
      Tags:
      - Key: Name
        Value: sb8x4-log
```

PR #2239, opened 2020-03-30
Fix ignition files location

wtty-fool:
Did a little investigation - `cache` folder was not right, it should have been `/go/pkg/mod`.

Also (I found it interesting), the last path segment is versioned, i.e `v6@6.0.1` instead of `v6` found in repository. Added some bash magic to copy it properly. Tested it on _gauss_ by creating a cluster and it works.


wtty-fool:
I think we were ready to merge before this PR, so feel free to assign this to me. I'll pester people for reviews.

This issue cropped up while testing, along with https://github.com/giantswarm/k8scloudconfig/pull/684 (the reason aws-operator has been restarted over and over when ignition credentials were missing), but we should be good to merge now.

PR #2245, opened 2020-03-31
legacy: Versions from releases - Part 3 - Implementation

tfussell:
## Checklist


service/controller/key/generate/generate_amis.go
```diff
@@ -0,0 +1,110 @@
...
+			t := z.Token()
+			if t.Data == "a" {
+				versions = append(versions, strings.TrimSuffix(t.Attr[0].Val, "/"))
+			}
+		}
+	}
+
+	versionAMIs := map[string]key.AMIInfoList{}
+	for _, version := range versions {
+		url := fmt.Sprintf("https://stable.release.core-os.net/amd64-usr/%s/coreos_production_ami_all.json", version)
```
MarcelMue, line 67: Just thinking: We will need to change this URL eventually when `core-os` finally goes the way of to dodo.
We should make this configurable through e.g. `installations` because otherwise it will be a very sad day when we need to scramble and patch old operator versions which stopped working because the URL does not serve anymore.

MarcelMue: Okay I just saw that this is fixed in a PR on this branch - there are too many branches :D

PR #2247, opened 2020-04-01
legacy: Generate Flatcar AMIs

tfussell:
## Checklist


tfussell:
The latest Flatcar release is 2345.3.1. I checked that AMIs before this version points to CoreOS stable releases on the AWS Console as expected and 2345.3.1 alone points to the stable version released by Kinvolk. I also ensured that all images are HVM as we don't use PV images as far as I know. Finally, I checked that AMI regions match those in AWS Console as well.

MarcelMue:
@cornelius-keller @mazzy89 This is the change we need to just specify the `containerlinux` version in the release and have a nice little `flatcar` cluster.

mazzy89:
Considering this a pretty important piece would be nice to have unit tests also for a better review.

tfussell:
Okay, there are unit tests for both scraping functions now. Please take a look @mazzy89

tfussell:
I want to get some of these PRs merged. Let me know if you have any issues with this one.

mazzy89:
Thank you @tfussell to add tests. now it is much clearer the input and the output of the script and it is easier for the future generation to change the code. surprised here that we need to deal and parse an HTML. Why?

MarcelMue:
> Thank you @tfussell to add tests. now it is much clearer the input and the output of the script and it is easier for the future generation to change the code. surprised here that we need to deal and parse an HTML. Why?

`coreOS` (which we still have some versions from) only publishes the `hvm` list which we used previously to match `hvm` `AMI`s to coreOS versions as a HTML page.
It's not pretty but AFAIK there is no other way to get the `AMI`s from coreOS.

mazzy89:
I was sure that putting in place unit tests would open the discussions and make it interesting 🤓

CoreOS provides the list of stables releases as a pretty handy JSON so it would be enough to parse that JSON rather than use an HTML file, much easier. 

This is why I asked for unit tests. there was something that it was not convincing me because I have done already this thing in OSS and it was strange that we were parsing an HTML.

the list is in here https://coreos.com/dist/aws/aws-stable.json

tfussell:
I found the JSON for the AMIs for each region for an individual release and that's basically what I'm using in this script, but I couldn't find all releases together anywhere except on the HTML page. That one you linked only includes 2345.3.0, right? (Looking at it on my phone)

mazzy89:
I see @tfussell now. you wanted to have all the releases together to not make any further lookup in any other external resource. so you just pull once the HTML and then take the release.

PR #2251, opened 2020-04-01
legacy: Versions from releases - Part 2 - Vendor k8scc

tfussell:


tfussell:
Vendor only, going with a ping.

PR #2264, opened 2020-04-06
legacy: migrate from dep to Go modules

tfussell:
## Checklist


tfussell:
Tested on `gauss` just now. Cluster came up without any issues.

PR #2275, opened 2020-04-08
Add new permissions needed for latest node pools version

pipo02mix:
## Checklist


pipo02mix:
Tested on gauss successfully (after 2 retries :) )

Screen Shot 2020-04-08 at 15 56 06

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fixed
```
xh3b4sd, line 12:
```suggestion
### Fixed
```

policies/tenant_cluster.json
```diff
@@ -56,57 +56,35 @@
                 "iam:UpdateAssumeRolePolicy",
                 "iam:UpdateRoleDescription"
             ],
-            "Resource": "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-EC2-K8S-Role"
+            "Resource": [
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-EC2-K8S-Role",
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-Route53Manager-Role",
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/gs-*",
```
xh3b4sd, line 62: I understand that this is the desired rule we want to establish now?

pipo02mix: No, this is needed for the current state. When fmt PR is approved it will be the source for the transition. There is a ticket in our roadmap assigned to firecracker team. The ticket description suggests how can be done the transition to the new naming.

But with the latest node pools, we need to add these changes in order to let aws operator create all resources

pipo02mix:
> Where is the role now coming from that we allow via gs-*?

We added a [new node pool role](https://github.com/giantswarm/aws-operator/blob/138f5e7220499d6f516b5833756ffa4cf4b0ff2d/service/controller/resource/tcnp/template/template_main_iam_policies.go#L8) but we did not update this policy. We need to do it, otherwise, a current customer with this policy cannot create clusters

It is independent of the coming naming convention (tracked in different ticket). The PM defines the steps I am doing for more info

PR #2276, opened 2020-04-08
helm: add resource names stem to `_helpers.tpl`

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/9878.

Define a common name stem for resource names in helpers template to
reduce repetition in charts' `values.yaml` file.

## Checklist


helm/aws-operator/values.yaml
```diff
@@ -34,10 +31,10 @@ project:
...
-    name: '{{ .Release.Name | replace "." "-" | trunc 47 }}'
+    name: '{{ include "aws-operator.resourceName" . }}'
     namespace: "giantswarm"
   psp:
-    name: '{{ .Release.Name | replace "." "-" | trunc 47 }}-psp'
+    name: '{{ include "aws-operator.resourceName" . }}-psp'
   pullSecret:
-    name: '{{ .Release.Name | replace "." "-" | trunc 47 }}-pull-secret'
+    name: '{{ include "aws-operator.resourceName" . }}-pull-secret'
     namespace: "giantswarm"
```
kopiczko, line 40: Wouldn't it be better to move all of that to the helper? I didn't want to make it configurable from the outside. I think it makes little sense. Back then it was a compromise between having it here and repeating this everywhere in the resources, because I didn't know about _helpers.tpl pattern.

helm/aws-operator/templates/_helpers.tpl
```diff
@@ -13,6 +13,18 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
```
kopiczko, line 15: We will need to get rid of `tpl` in places where this is used.

helm/aws-operator/values.yaml
mig4: I'm not sure I follow, we can't move the entire `resource` block to the helper because it isn't used as-is, only individual variables are used in other templates.

I suppose we could define variables like `aws-operator.resourceName`, `aws-operator.pspResourceName` and `aws-operator.pullSecretResourceName` in the helper instead and remove the `resource` block from `values.yaml`, but what about the namespace? Shouldn't it be configurable?

kopiczko: I thought defining variables like `resource.default.name` etc. Do you think `aws-operator` prefix is important? I don't think the namespace should be configurable. We don't have the usecase for that and we don't even know if it would work if we changed the namespace for a single resource.

TheoBrigitte: Taking over here.

* moved the entire `resource` block from values to helpers
* changed from `tpl` to `include` in actual templates
* replaced pullSecret namespace with default namespace, because well they are just the same

helm/aws-operator/templates/_helpers.tpl
```diff
@@ -13,6 +13,30 @@ Create chart name and version as used by the chart label.
...
 
+{{/*
+Create a name stem for resource names
+
+When pods for deployments are created they have an additional 16 character
+suffix appended, e.g. "-957c9d6ff-pkzgw". Given that Kubernetes allows 63
+characters for resource names, the stem is truncated to 47 characters to leave
+room for such suffix.
+*/}}
+{{- define "resource.default.name" -}}
```
kopiczko, line 24: Should we prefix with `aws-operator.` for consistency? E.g. `aws-operator.resource.default.name`.

helm/aws-operator/templates/_helpers.tpl
```diff
@@ -13,6 +13,30 @@ Create chart name and version as used by the chart label.
...
+{{- include "resource.default.name" . -}}-psp
+{{- end -}}
+
+{{- define "resource.pullSecret.name" -}}
+{{- include "resource.default.name" . -}}-pull-secret
+{{- end -}}
+
+{{- define "resource.default.namespace" -}}
+giantswarm
+{{- end -}}
```
kopiczko, line 38: Let's put that at the bottom of this file. `labels` & `selectorLabels` should be probably defined in all charts. This part is optional.

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: How about removing all `aws-operator` prefix? So we can then just copy this file over without having to rename those variables every-time.

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: we can also move this to a different file then like `_resource.tpl`

helm/aws-operator/templates/_helpers.tpl
kopiczko: I'm not against that but I feel it would make Mr. @mig4 sad. He based that on upstream charts. E.g:
 https://github.com/helm/charts/blob/c1bd92189296e4e3ae248f27da4fc7c2138b2648/stable/chartmuseum/templates/_helpers.tpl#L19

helm/aws-operator/templates/_helpers.tpl
```diff
@@ -2,34 +2,34 @@
...
-helm.sh/chart: {{ include "aws-operator.chart" . | quote }}
+helm.sh/chart: {{ include "chart" . | quote }}
 {{- end -}}
 
 {{/*
 Selector labels
 */}}
-{{- define "aws-operator.selectorLabels" -}}
-app.kubernetes.io/name: {{ include "aws-operator.name" . | quote }}
+{{- define "selectorLabels" -}}
```
kopiczko, line 32: WDYT about this to make alphabet gods happy (also variable grouping here would make more sense to me)?
 
```suggestion
{{- define "labels.selector" -}}
```

helm/aws-operator/templates/_helpers.tpl
kopiczko: But since it would allow simply copy/pasting stuff I'm leaning towards this solution. I basically don't see any downsides other than not being compatible with upstream.

Maybe for managed apps that is a problem if they defined dependencies? But I feel that helm dependencies is an evil.

TheoBrigitte: I see no issue with not being compatible with upstream. When there's a real necessity to do this we can change this in the future, but I'd like to keep it simple and allow copy/pasting of this helpers.

Again we for helm dependency, we are no there yet and when that's an issue we can change it in the future.

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: This doesn't work as we already use `labels` as a string.

TheoBrigitte: We could do `labels.common` and `labels.selector`

kopiczko: Sounds good to me. I didn't know that a variable can't be a prefix of another one.

helm/aws-operator/templates/_helpers.tpl
kopiczko: Sounds good 👍

helm/aws-operator/templates/_helpers.tpl
kopiczko: Fine with me.

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: done

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: done

helm/aws-operator/templates/_helpers.tpl
TheoBrigitte: Going with no prefix, happy to help when that's an issue in the future.

kopiczko:
This 100% satisfies my OCD :D :shipit:

kopiczko:
*Assuming it's still deployable.

TheoBrigitte:
Tested successfully on `gauss`

```
$ k get po |grep voo-refactor
aws-operator-voo-refactor-chart-resource-names-58b49b8d94-fnr84   1/1     Running     0          51s
```

```
$ helm list | grep voo-refactor
aws-operator-voo-refactor-chart-resource-names  1               Thu Apr 16 15:32:59 2020        DEPLOYED        aws-operator-8.2.3-ce7e74b74562364c84289b02bcfe1ff65a33a51e   8.2.4-dev 
```

PR #2306, opened 2020-04-15
Add hosted zone id to tccpoutputs

calvix:
The move of route53 IAM role to tccpn stack  broke some references so we need to fetch hostedZoneID and InternalHostedZoneID via CF outputs and propagate to control context

Also, I had to add `tccpoutputs` to `control_plane_resource_set` that means adjusting the resource with `ToClusterFunc`


service/controller/controllercontext/status_types.go
```diff
@@ -56,6 +56,12 @@ type ContextStatusTenantClusterAWS struct {
 	Region    string
 }
 
+type ContextStatusTenantClusterDNS struct {
```
calvix, line 59: moved all DNS related values to new struct

PR #2307, opened 2020-04-15
Test replace deepcopytime

tfussell:
## Checklist


tfussell:
This is failing with `CrashLoopBackoff` due to a validation error in the `AWSConfiig` CRD introduced while adding documentation. I will test on `master` instead which doesn't install CRDs (I think).

PR #2309, opened 2020-04-15
legacy: Legacy release revision

rossf7:
Towards giantswarm/giantswarm#10098

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Switch from dep to Go modules.
+- Use release.Revision in Helm chart for Helm 3 support.
 
 ## [5.6.0] - 2020-01-29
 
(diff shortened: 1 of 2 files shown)


tfussell:
I have this change in the `legacy-update-helm` branch already: https://github.com/giantswarm/aws-operator/blob/25a197520cd84eddc785af438a87c1fe8c0492e2/helm/aws-operator/templates/deployment.yaml#L19 This should be merged into `legacy` tomorrow. I don't mind at all if you merge this to `legacy` then I can rebase my branch--just wanted to mention.

tfussell:
P.S. I appreciate the explicit backport into `legacy`. Sometimes it's hard to know which `master` changes should be brought into `legacy`.

rossf7:
Tested on `ginger`.

rossf7:
@tfussell Ah it was this PR :)

The problem is even once the app CR is deployed we need to update the helm release if the values will change. Helm 3 can't template `Release.Time` so the release will fail.

So I'll also need to "reissue" the tarballs for the active releases.

rossf7:
>  Let's use @giantswarm/team-firecracker-pe next time then we do not bother Marian with this.

@xh3b4sd Ack. 👍

PR #2318, opened 2020-04-16
Rm tccpoutputs from cluster

calvix:
this caused issues with updates as the update did not proceed since  the new tccpoutputs were not available and that caused cancellign the reconcilation

 i don't see any reason why `tccpoutputs` should be in the cluster resource as none of the values is used there so removing


calvix:
nope this is wrong , we need it dammit

PR #2323, opened 2020-04-17
legacy: Remove CoreOS images

tfussell:
In Azure and KVM, we decided not to support CoreOS at all in upcoming operator releases so I'm removing the CoreOS AMIs here to match.

## Checklist


service/controller/key/generate/generate_amis.go
```diff
@@ -104,45 +102,32 @@ func scrapeVersionAMIs(source io.Reader) (map[string]string, error) {
 }
 
 func main() {
```
MarcelMue, line 104: Did we add some docs on how we do this? Something maybe in this operators readme would be good.

tfussell: Good call. I added some. Let me know what you think.

service/controller/key/amis.go
```diff
@@ -1,249 +1,147 @@
+// NOTE: This file is generated. Do not edit.
```
MarcelMue, line 1: This seems to still be missing `china` images. Can this easily be extended to include them?

mazzy89: We generate as an artifact of the build to China a JSON file containing the `current` (latest) AMIs ID for China. It is public available

tfussell: Yep! I'll add it in a separate PR once I have the China JSON URL.

PR #2336, opened 2020-04-20
'migrate-master-to-aws-launch-template'

calvix:
Lots of renaming  but not that much real changes


calvix:
tested on ginger and giraffe (china)

PR #2343, opened 2020-04-22
Fix Release.Revision templating

rossf7:
Follow up to remove `$`. As per the docs https://helm.sh/docs/chart_template_guide/builtin_objects

Will go with a ping.


rossf7:
Going with a ping.

PR #2387, opened 2020-05-05
Add release resource

tfussell:
Since release will be accessed in multiple places, it should be set in a separate resource. This is ported from recent work on the `legacy` branch.

### DIFF
--- a/service/controller/resource/release/delete.go
+++ b/service/controller/resource/release/delete.go
@@ -1,4 +1,4 @@
-package accountid
+package release
 
 import (
 	"context"
--- a/service/controller/resource/release/delete.go
+++ b/service/controller/resource/release/delete.go
@@ -1,0 +1,9 @@
+package accountid
+
+import (
+	"context"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	return nil
+}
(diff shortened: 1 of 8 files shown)


service/controller/resource/release/delete.go
```diff
@@ -0,0 +1,9 @@
+package accountid
+
+import (
+	"context"
+)
+
+func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
+	return nil
```
tfussell, line 8: The `accountid` resource adds the value to the context in `EnsureDeleted` as well. That doesn't seem necessary here, but please let me know if I'm misunderstanding.

tfussell:
> I think this is not the best way to go about it. For HA Masters we change the approach. In #2368 I started to implement an "Image" service that does not. My intention with it is to enable it to cache fetched information scoped to the current reconciliation loop. See also [giantswarm/operatorkit#389](https://github.com/giantswarm/operatorkit/pull/389). So in general I think we should stop making use of the controller context and instead start to make resource handlers self sufficient. I will bring this up during the next SIG Operator Sync as well.

Okay, thanks for the info. Should I wait until the caching functionality is ready in operatorkit?

tfussell:
This isn't strictly necessary. I will close this in favor of getting the release in `tcnp` and `tccpn` where it's needed without `controllercontext` until we have a caching system in place.

PR #2401, opened 2020-05-07
Fix error pkg

njuettner:
Updating the error pkg to the latest release


anvddriesch:
Did you close this by accident? :)

PR #2421, opened 2020-05-11
'change-eni-static-ip-settings'

calvix:
towards https://github.com/giantswarm/giantswarm/issues/10474

remove ENI static address reservation and move eth1 routing file render to `aws-attach-etcd-dep` tool


calvix:
This is now fully tested and should be fine to merge.

5 clusters created and 3 updated all green

service/controller/internal/cloudconfig/template/systemd-networkd-eth1-network.go
```diff
@@ -1,18 +1,9 @@
...
-
-[RoutingPolicyRule]
-Table=2
-From={{.MasterENIAddress}}/32
-
-[Route]
-Destination=0.0.0.0/0
-Gateway={{.MasterENIGateway}}
-Table=2
+DHCP=no
```
calvix, line 8: this configuration has moved to this tool https://github.com/giantswarm/aws-attach-etcd-dep/pull/20

Now we just make sure that the interface doesn't use DHCP when it's attached

PR #2445, opened 2020-05-18
add a makefile

njuettner:
Similar PR: https://github.com/giantswarm/cluster-operator/pull/1055

Makes it a bit easier to use certain go tools, etc.

```bash
$ make help
Usage:

  build          builds a local binary
  build-linux    builds binary for linux/amd64
  build-darwin   builds binary for darwin/amd64
  install        install the application
  run            runs go run main.go
  clean          cleans the binary
  lint           runs golangci-lint
  test           runs go test with default values
  build-docker   builds docker image locally
  help           prints this help message
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ build-darwin: clean $(SOURCES)
 .PHONY: install
 ## install: install the application
 install: $(SOURCES)
-	go install .
+	go install $(BUILD_FLAGS) .
 
 .PHONY: run
 ## run: runs go run main.go
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,59 @@
+APPLICATION   ?= aws-operator
+VERSION       ?= $(shell git describe --tags --always --dirty)
+SOURCES       = $(shell find . -name '*.go')
+GOPKGS        = $(shell go list ./...)
+TAG           ?= $(VERSION)
+BUILD_FLAGS   ?= -v
+
+default: build
+
+.PHONY: build
+## build: builds a local binary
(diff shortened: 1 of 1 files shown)


kopiczko:
In my opinion random makefiles in some repos are bad idea. They will get out of sync and will not be trustworthy without looking inside. Having said that I think they are useful and I’m all in for generating them in devctl and happy to discuss that.

Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+.PHONY: build-darwin
+## build-darwin: builds binary for darwin/amd64
+build-darwin: clean $(SOURCES)
+	GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build $(BUILD_FLAGS) .
+
+.PHONY: install
+## install: install the application
+install: $(SOURCES)
+	go install .
```
TheoBrigitte, line 28: How about:
```suggestion
	go install $(BUILD_FLAGS) .
```

njuettner:
@kopiczko 
> In my opinion random makefiles in some repos are bad idea. They will get out of sync and will not be trustworthy without looking inside. Having said that I think they are useful and I’m all in for generating them in devctl and happy to discuss that.

Sure I understand your concern, I'm happy to discuss it (haven't heard about devctl so I'm curious 😄). If there would be a better way that this Makefile doesn't get stale I'm totally open.

njuettner:
I talked with Pawel about that it make sense to implement it in devctl. I will come up with a new approach to make it more maintainable. Closing this PR for now.

PR #2513, opened 2020-06-09
Remove dockershim and dockersock mounts from aws-node

marians:
Towards https://github.com/giantswarm/giantswarm/issues/11457

These mounts are not needed for aws-node v1.6.0 and newer.

## Checklist


calvix:
did you test cluster creation and cluster update from previous versions with this change and then see everything works including ingress?

marians:
testing is up next

marians:
When I tested this on gauss, liveness or readyness checks failed. Not pursuing this.

yulianedyalkova:
The docker socket should not be needed anymore. The dockershim one is being used as a [default](https://github.com/aws/amazon-vpc-cni-k8s/blob/1c1d4b91e8d21d979cad494fe8228c464d1d4d3b/pkg/cri/cri.go#L54-L57) in case `/var/run/cri.sock` is missing. So it makes sense that it failed if we didn't pass the containerd socket.

PR #2528, opened 2020-06-17
adding cache for ELB collector

anvddriesch:
Towards: https://github.com/giantswarm/giantswarm/issues/11097
We make a lot of requests to the aws-api when we collect metrics about the load balancers in our collector. This causes throttling issues for the customer.
This PR adds a cache for the ELB collector to save api requests. It is based on the cache used in our NAT collector. The collection method itself is unchanged.

## Checklist


anvddriesch:
Tested on `giraffe`. It's reporting metrics as it should.

PR #2546, opened 2020-06-24
auto-terminate-bad-nodes

calvix:
Node auto-repair controller

-  Periodic node state checks
- If node is bad state - NotReady - increase not-ready-tick annotation on the node
- If X consecutive checks sees node in NotReady state (tick counter reaches certain threshold) -> terminate the node via AWS ASG api

Can be enabled/disabled by installation, it's going to be experimental feature eas start that we want to enable on certain installations before rolling everywhere

was initially produced as part of the hackathon

I decided to go for a separate controller to have better rsync period since the cluster resync period is 5mins also it makes more sense to have it outside fo the big cluster reconciliation loop


marians:
A high level explanation of what's happening here would be appreciated

main.go
```diff
@@ -154,6 +154,7 @@ func mainE(ctx context.Context) error {
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CrtFile, "", "Certificate file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 
+	daemonCommand.PersistentFlags().Bool(f.Service.NodeAutoRepair, false, "Enable Node Auto repair feature.")
```
marians, line 157: Why is the default `false` here (no auto repair), while it's `true` in the config?

service/controller/key/cluster.go
```diff
@@ -86,6 +87,16 @@ const (
 	ComponentOS = "containerlinux"
 )
 
+const (
+	// Defines how many times node-auto-repair logic needs to see node
+	// in NotReady state before it will terminate it to repair it.
+	NodeNotReadyTickThreshold = 4
```
marians, line 93: Would be nice to have an explanation of a "tick" here. Is it one reconciliation loop?

service/controller/resource/nodeautorepair/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+
+const (
+	gaugeValue float64 = 1
+)
+
+var (
+	nodeAutoRepairTermination = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "aws_operator_node_auto_repair_termination",
+			Help: "Metric representing node termination due node auto repair feature.",
```
marians, line 15:
```suggestion
			Help: "Gauge representing node termination due to node auto repair feature.",
```

BTW, why gauge and not counter? This number can only go up, so I think a counter would fit better.

calvix:
this is WIP for the hackathon, I won't put too much info until its stop changing

service/controller/resource/nodeautorepair/metrics.go
calvix: It will never have other value than 1

calvix: each instance is only terminated once

main.go
calvix: fixed

PR #2557, opened 2020-07-02
Emit metric label on instance lifecycle to count spot/on-demand

marians:
The goal here is to add a label to the `aws_operator_ec2_instance_status` for the lifecycle of the instance. The value can be `"spot"` or `"on-demand"`.

## Checklist

### DIFF
--- a/service/collector/ec2_instances.go
+++ b/service/collector/ec2_instances.go
@@ -34,6 +34,9 @@ const (
 	// labelPrivateDNS will contain the private dns name
 	labelPrivateDNS = "private_dns"
 
+	// labelLifecycle shows the distinction betwen spot and on-demand.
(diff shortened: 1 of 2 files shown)


service/collector/ec2_instances.go
```diff
@@ -34,6 +34,9 @@ const (
 	// labelPrivateDNS will contain the private dns name
 	labelPrivateDNS = "private_dns"
 
+	// labelLifecycle shows the distinction betwen spot and on-demand.
+	labelLifecycle = "lifecycle"
```
njuettner, line 38: does it make sense to rename it to labelInstanceLifecycle to keep it consistent like the other Instance labels?

marians: Yes, indeed. Didn't even notice.

PR #2597, opened 2020-07-15
Fix region prefix in helm chart

wtty-fool:
Fixes region condition error.

Quote from helm docs:
> hasPrefix and hasSuffix
> The hasPrefix and hasSuffix functions test whether a string has a given prefix or suffix:
> 
> hasPrefix "cat" "catch"
> The above returns true because catch has the prefix cat.

### DIFF
--- a/helm/aws-operator/templates/configmap.yaml
+++ b/helm/aws-operator/templates/configmap.yaml
@@ -45,7 +45,7 @@ data:
           clusterDomain: '{{ .Values.Installation.V1.Guest.Kubernetes.ClusterDomain }}'
           networkSetup:
             docker:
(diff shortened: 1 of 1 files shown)


wtty-fool:
Deployed it real quick and checked configmap:
```
~ k get cm -n giantswarm aws-operator-fix-region-codition -o yaml | grep -A2 registry:
    \ registry:\n    domain: 'registry-intl.cn-shanghai.aliyuncs.com'\n    mirrors:
    ''\n  \n  installation:\n    name: 'giraffe'\n    guest:\n      ipam:\n        network:\n
    \         CIDR: '10.1.0.0/16'\n          subnetMaskBits: '24'\n          privateSubnetMaskBits:
```
Looks better :sweat_smile:

PR #2600, opened 2020-07-15
Drop unused integration tests to simplify dependencies

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/11974

We disabled the integration tests a while ago. This also removes the integration directory and allows us to simplify the dependency graph.

The true e2e tests we want to add should not need our `helmclient`. We plan to refactor our integration tests so they also don't use `helmclient` directly.


go.mod
```diff
@@ -7,17 +7,10 @@ require (
...
 	github.com/giantswarm/backoff v0.2.0
 	github.com/giantswarm/certs/v2 v2.0.0
-	github.com/giantswarm/e2e-harness v0.2.0
-	github.com/giantswarm/e2eclients v0.2.0
-	github.com/giantswarm/e2esetup v0.3.0
-	github.com/giantswarm/e2etemplates v0.2.0
-	github.com/giantswarm/e2etests v0.3.1
 	github.com/giantswarm/errors v0.2.3
 	github.com/giantswarm/exporterkit v0.2.0
-	github.com/giantswarm/helmclient v1.0.3
```
rossf7, line 20: 🔥  This means we can get rid of the replace madness. 🎉

rossf7:
Tested on `ginger`.

PR #2602, opened 2020-07-16
fix-route53-for-china

calvix:
towards  https://github.com/giantswarm/giantswarm/issues/11636

## Checklist


calvix:
even thou this is a fix for route53 in china, it can be merged as we just simply replace the static value with the value from a variable, nothing will change as in AWS global this value will be the same and for china, this is not rendered.

But I don't want to have this hanging around for long time.

PR #2603, opened 2020-07-17
event recorder

njuettner:
## Checklist


njuettner:
tested in gauss, lowercase messaged will be fixed 🤞

describe awsclusters
image

describe awsmachinedeployments
image

describe awscontrolplanes
image

njuettner:
HA enabled
image

njuettner:
I integrated the status when a CF is complete, however it will write this status with each loop. Since it's writing with each loop it will also not disappear after eventTTL has expired (1h).

**awscluster**
image

**awscontrolplane**
image

**awsmachinedeployments**
image

Probably need some rubberduck here, any good ideas?

calvix:
that is actually very nice :)

service/controller/resource/tccpn/create.go
```diff
@@ -126,14 +126,23 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 			return nil
 		}
 
+		if *o.Stacks[0].StackStatus == cloudformation.StackStatusCreateComplete {
+			r.event.Emit(ctx, &cr, "CFCreated", fmt.Sprintf("The tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusCreateComplete))
+		}
+		if *o.Stacks[0].StackStatus == cloudformation.StackStatusUpdateComplete {
+			r.event.Emit(ctx, &cr, "CFUpdated", fmt.Sprintf("The tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusUpdateComplete))
+		}
+
```
calvix, line 145: maybe we could emit even other CF states like Error, Deleting or Rollback

calvix:
I don't see a big problem with repeating events.  At least even after 1H I can see the current CF state.

helm/aws-operator/templates/rbac.yaml
```diff
@@ -120,6 +120,14 @@ rules:
...
     verbs:
       - get
+
+  # The aws-operator needs access for events to write those to the belonging objects.
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - "*"
```
xh3b4sd, line 130: Object creation does not mean object deletion, even though both are writing operations. Can we make this more specific?

service/controller/control_plane_drainer.go
```diff
@@ -24,9 +24,11 @@ import (
 	"github.com/giantswarm/aws-operator/service/controller/resource/drainerfinalizer"
 	"github.com/giantswarm/aws-operator/service/controller/resource/drainerinitializer"
 	"github.com/giantswarm/aws-operator/service/internal/asg"
+	"github.com/giantswarm/aws-operator/service/internal/recorder"
 )
 
 type ControlPlaneDrainerConfig struct {
+	Event     recorder.Interface
```
xh3b4sd, line 31: This reads wrong to me. What about a bit more consistency? 

```suggestion
	Event     event.Interface
```

service/controller/resource/s3bucket/resource_test.go
```diff
@@ -4,6 +4,10 @@ import (
 	"testing"
 
 	"github.com/giantswarm/micrologger/microloggertest"
+
+	"github.com/giantswarm/aws-operator/service/internal/unittest"
+
+	"github.com/giantswarm/aws-operator/service/internal/recorder"
```
xh3b4sd, line 10:
```suggestion
	"github.com/giantswarm/aws-operator/service/internal/recorder"
	"github.com/giantswarm/aws-operator/service/internal/unittest"
```

service/controller/resource/tccpn/create.go
```diff
@@ -126,14 +126,23 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
 			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 			return nil
 		} else if *o.Stacks[0].StackStatus == cloudformation.StackStatusUpdateInProgress {
 			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusUpdateInProgress))
+			r.event.Emit(ctx, &cr, "CFUpdate", fmt.Sprintf("the tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusUpdateInProgress))
 			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
 			return nil
 		}
 
+		if *o.Stacks[0].StackStatus == cloudformation.StackStatusCreateComplete {
```
xh3b4sd, line 139: I would like to have one consistent block of these conditions and not handle different conditions differently.

helm/aws-operator/templates/rbac.yaml
njuettner: Yes I agreee I tested something with searching events and I wanted to change it back, but good catch 👍

service/controller/control_plane_drainer.go
njuettner: 👍

service/controller/resource/tccpn/create.go
njuettner: 👍

njuettner:
> * When we introduce something new it would be cool to get more quickly feedback using less code. This PR is too big and adds too much noise.

I understand. I only wanted to see what events appear first.

> * I do not see why we chose some events over others. I think we should start with the most prominent events and then expand the usage. There are already too many "useless" events logged.

It was not my intention to chose some events over others. I just mainly went through the existing log lines and started with the ones where I thought it would actual make sense to see them. I can take another look and drop some.

> * I am wondering why we need two different tools for logs and events. We repeat the log messages twice in the same place, but then not really, since we rewrite the messages slightly. I would like to talk about alternatives, like enabling the micrologger to emit certain events selectively.

The reason is micrologger is it is a bit different than other loggers. It does not satisfy the requirements to use it for the event recorder (https://github.com/kubernetes/client-go/blob/master/tools/record/event.go#L279). Even if we would integrate it it will write it as a plain log line.
I don't know exactly what you mean by enabling the micrologger to emit events, but maybe we can pick this up for a sig-operator session?

service/controller/resource/tccpn/create.go
njuettner: good idea, for error this would automatically handled in operatorkit when we set a description and a kind for the error so it will picked up and emit an error event.

the other states I would probably ask for help since I don't know what we want to add. I would also iterate from a certain point 🙂.

xh3b4sd:
> I can take another look and drop some.

Let's focus on the most important once, which are about the main cluster transitions. A handful should be good enough as a starter. 

> I don't know exactly what you mean by enabling the micrologger to emit events

What about extending micrologger to optionally emit these events? I am not sure if this is a good idea to mix, but as a suggestion. /cc @kopiczko 

```
r.logger.LogAndEmitCtx(ctx, cr, "level", "debug", "message", "requested cloud formation stack update", "reason", "CloudFormationUpdate")

**Comment by njuettner** [comment]
> > I don't know exactly what you mean by enabling the micrologger to emit events
> 
> What about extending micrologger to optionally emit these events? I am not sure if this is a good idea to mix, but as a suggestion. /cc @kopiczko
> 
> ```
> r.logger.LogAndEmitCtx(ctx, cr, "level", "debug", "message", "requested cloud formation stack update", "reason", "CloudFormationUpdate")
> ```

And the initialization for the [event recorder](https://github.com/giantswarm/aws-operator/blob/92d80cc87baa2bd97eb3304a7b89f6090d01d48c/service/internal/recorder/recorder.go#L25) would be optional inside when initiating a new logger with [micrologger.New]( https://github.com/giantswarm/micrologger/blob/master/logger.go#L25)?

**Comment by xh3b4sd** [comment]
When you configure an event recorder you could make use of this extensions.

**Comment by njuettner** [comment]
Let's talk about that, I'm not sure if I got this right.

**Comment by kopiczko** [comment]
I dislike the idea of adding k8s dependencies to micrologger. I also _don't_ dislike having a repeated message in `r.event.Emit` because they are next to each other. And I believe we'll have way less events than logs.

My 2c. You do what you think is right.

**Comment by xh3b4sd** [comment]
Good feedback. Let's keep it separate.

File: service/controller/resource/tccpn/delete.go [hunk 6]
```diff
@@ -33,6 +34,18 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 		_, err = cc.Client.TenantCluster.AWS.CloudFormation.UpdateTerminationProtection(i)
 		if IsDeleteInProgress(err) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "the tenant cluster's control plane nodes cloud formation stack is being deleted")
+			r.event.Emit(ctx, &cr, "CFDelete", fmt.Sprintf("The tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusDeleteInProgress))
```
njuettner, line 37: In case the deletion is in progress we want to add an event

service/controller/resource/tccpn/delete.go
```diff
@@ -33,6 +34,18 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", "keeping finalizers")
+			finalizerskeptcontext.SetKept(ctx)
+
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource")
+
+			return nil
+
+		} else if IsDeleteFailed(err) {
```
njuettner, line 46: I also wanted to catch deletion failed cases

service/controller/resource/tccp/error.go
```diff
@@ -20,6 +21,45 @@ var executionFailedError = &microerror.Error{
 	Kind: "executionFailedError",
 }
 
+// event... is an error type for situations where we want to create an Kubernetes event in operatorkit.
+var eventCFCreateError = &microerror.Error{
```
njuettner, line 25: I introduced eventErrors, those should be used in case we want to emit events via operatorkit. For such events we need to make sure Kind and Desc is set. Kind also starts with upper-case letter to make sure it aligns with other event reasons.

helm/aws-operator/templates/rbac.yaml
```diff
@@ -120,6 +120,16 @@ rules:
       - "/healthz"
     verbs:
       - get
+
+  # The aws-operator needs access for events to write those to the belonging objects.
```
njuettner, line 124: Updated RBAC permissions otherwise we cannot emit events

service/controller/resource/cleanupmachinedeployments/delete.go
```diff
@@ -52,6 +52,7 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
 		}
 
 		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("deleted aws machine deployment %#q for tenant cluster %#q", md.Namespace+"/"+md.Name, key.ClusterID(&cr)))
+		r.event.Emit(ctx, &cr, "MachineDeploymentDeleted", fmt.Sprintf("Deleted aws machine deployment %#q for tenant cluster %#q", md.Namespace+"/"+md.Name, key.ClusterID(&cr)))
```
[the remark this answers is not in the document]
njuettner: This is only needed to align with K8s event convention.

K8s Events:

image

Reason and Message are written with an upper-case letter.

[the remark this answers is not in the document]
xh3b4sd: We could keep the messages aligned and let `Emit` capitalize the message.

service/controller/resource/tccpn/create.go
```diff
@@ -122,16 +123,19 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
-			return nil
-		} else if *o.Stacks[0].StackStatus == cloudformation.StackStatusUpdateInProgress {
-			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("the tenant cluster's control plane nodes cloud formation stack has stack status %#q", cloudformation.StackStatusUpdateInProgress))
+			return microerror.Maskf(eventCFCreateError, "expected successful status, got %#q", *o.Stacks[0].StackStatus)
+		} else if *o.Stacks[0].StackStatus == cloudformation.StackStatusRollbackFailed {
+			return microerror.Maskf(eventCFRollbackError, "expected successful status, got %#q", *o.Stacks[0].StackStatus)
+		} else if *o.Stacks[0].StackStatus == cloudformation.StackStatusUpdateRollbackFailed {
+			return microerror.Maskf(eventCFUpdateRollbackError, "expected successful status, got %#q", *o.Stacks[0].StackStatus)
+
+		} else if stackInProgress(*o.Stacks[0].StackStatus) {
```
njuettner, line 132: simplified it a bit. 

all stack status which are in progress should be emitted via `CFInProgess`
and
all stack status which are completed should be emitted via `CFCompleted`

so that we don't have to catch all status (Create,Update,Rollback...)

service/controller/resource/tccpn/create.go
```diff
@@ -611,3 +617,11 @@ func idFromSubnets(subnets []*ec2.Subnet, name string) string {
 
 	return ""
 }
+
+func stackComplete(status string) bool {
```
njuettner, line 621: Added helper function to check for the status

service/internal/recorder/recorder.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Emit writes only informative events like status of creation or updates.
+// Error events will be handled by operatorkit when using microerror.
+func (r *Recorder) Emit(ctx context.Context, obj pkgruntime.Object, reason, message string) {
+	r.Event(obj, corev1.EventTypeNormal, reason, upper(message))
+}
+
+// upper is a helper function to uppercase first letter of the event message
+func upper(in string) string {
```
njuettner, line 48: Added the helper function, to let Emit handle uppercase first letter.

service/controller/resource/tccpn/create.go
xh3b4sd: We could add these functions in the key or cloud formation package so that we reuse the code where possible.

njuettner: Makes totally sense, I moved it to key since we already use it and it doesn't collide with cloudformation package (external vs internal)

njuettner:
Tested in ginger, looks good (triggered some updates):
We have less events now 🙂 


image


image

paurosello:
> Tested in ginger, looks good (triggered some updates):
> We have less events now 
> 
> image
> 
> image

I love it!

PR #2606, opened 2020-07-23
set etcd mount dir mode to 0700

calvix:
etcd v 4.10+ wont start unless the permission on the etcd data dir(`/var/lib/etcd`) are se to 0700

also removed old non-asg etcd mount file (its no longer used)

## Checklist


calvix:
unfortunately, the solution with specifying DirectoryMode in the systemd unit does not work.

This needs to be fixed by running chmod on the dir so pushing teh change to k8scc instead

PR #2632, opened 2020-07-29
Do not automatically push to aws-app-collection

yulianedyalkova:
## Checklist


yulianedyalkova:
@giantswarm/team-firecracker is it intentional that all PRs get blocked on `hold-push-to-aliyun-pr`?

marians:
That depends on the definition of "blocked". Pushing to Aliyun usually takes a long time, with large variation. That's why we made it on-demand only for PRs. So whoever wants to test a branch in China manually approves the Aliyun push.

Releases should always get pushed to Aliyun.

marians:
@yulianedyalkova Can you please explain why you're removing the push to aws-app-collection?

yulianedyalkova:
It could be that I'm not understanding something, but right now my PR doesn't go green because it needs a manual interaction to reach a CI task that's required.

@marians I should've added that in the description, my bad. Yesterday I did the migration to release-operator on all AWS installations. So we removed all the `aws-operator`s from the `aws-app-collection` and all of them that are referenced by a release will get automatically deployed by `release-operator`.

xh3b4sd:
> right now my PR doesn't go green

The only thing you need to pay attention to is the `go-build` step and then your manual testing you want to do. This is what it is right now, which is also why we put more energy into `awscnfm`. The build does not just "go green" here.

marians:
Ah, get it. The problem here in this PR is not the push to Aliyun, which is an **optional** check.

The reason why you still saw this ...

image

... is the `push-aws-operator-to-app-catalog` check, which is **Required**.

But this very check is what you are removing from the CircleCI config with this PR. So it cannot become green.

This is configred for the repository in [Settings / Branches](https://github.com/giantswarm/aws-operator/settings/branch_protection_rules/5709181).

image

I just changed it, so the `push-aws-operator-to-app-catalog` check is no longer required.

stone-z:
I think the problem was that it's not clear whether or not it's ok to merge with this state:
aws operator does require aliyun Screenshot from 2020-08-04 12-11-10
unless someone already told you to ignore aliyun because it takes too long

PR #2658, opened 2020-08-10
Test for release version when operator version did not change

njuettner:
PR only for testing

## Checklist


njuettner:
We're now introducing the release tag for our ec2 instances.

image

This allows us to roll the nodes because we trigger the update by checking our components when they have changed

image

This changes the launch template by changing the release version and we're able to roll all nodes which are needed

image
image

PR #2659, opened 2020-08-10
Update tenant_cluster.json

paurosello:
Towards https://github.com/giantswarm/giantswarm/issues/12645

## Checklist


paurosello:
This does not actually need a release, but customers should update the IAM policy on their installations to fix this issue manually.

I will inform SE that this needs to be done in all AWS CP.

PR #2662, opened 2020-08-11
make namespace configurable

anvddriesch:
https://github.com/giantswarm/giantswarm/issues/12647
## Checklist

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "aws-operator"
 	source      string = "https://github.com/giantswarm/aws-operator"
-	version            = "8.7.6-anvdd"
+	version            = "8.7.6-dev"
 )
 
 func Description() string {
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "aws-operator"
 	source      string = "https://github.com/giantswarm/aws-operator"
-	version            = "8.7.6-dev"
+	version            = "8.7.6-anvdd"
 )
 
 func Description() string {
--- a/service/controller/resource/secretfinalizer/secret_accessor.go
+++ b/service/controller/resource/secretfinalizer/secret_accessor.go
@@ -20,7 +20,7 @@ func newSecretAccessors(ctx context.Context, cr infrastructurev1alpha2.AWSCluste
 		// certificate.
 		{
 			Name:      fmt.Sprintf("%s-api", key.ClusterID(&cr)),
-			Namespace: "default",
+			Namespace: cr.GetNamespace(),
 		},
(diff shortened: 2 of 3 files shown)


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "aws-operator"
 	source      string = "https://github.com/giantswarm/aws-operator"
-	version            = "8.7.6-dev"
+	version            = "8.7.6-anvdd"
```
anvddriesch, line 8: I think I actually don't have to do this anymore.

anvddriesch:
I went through the `aws-operator` very closely. It seems like the only places where we explicitly still use the default namespace are
- default cluster definitions in unit tests : That should be fine
- eniconfigs : they are for the AZs so I believe they should remain there.
- secrets : made this one configurable as they are for the TC specifically

service/controller/resource/secretfinalizer/secret_accessor.go
```diff
@@ -20,7 +20,7 @@ func newSecretAccessors(ctx context.Context, cr infrastructurev1alpha2.AWSCluste
 		// certificate.
 		{
 			Name:      fmt.Sprintf("%s-api", key.ClusterID(&cr)),
-			Namespace: "default",
+			Namespace: cr.GetNamespace(),
```
njuettner, line 23: From a security perspective, will that be ok? The thing is it might be than possible for someone from the customer side to get access to the cert key as soon as they have access via RBAC to that namespace?

anvddriesch: That is a very good question.
Since right now everything is in the default namespace, wouldn't it at least make the access scope smaller?

stone-z: What would a customer need this for?
I don't see an immediate reason why a customer would need to access this secret. Presumably, there should be individual keypairs/kubeconfigs for each customer user which may have lower access than this, so I would be inclined to keep these outside customer-facing namespaces.

anvddriesch: That makes sense. I was approaching it from the perspective that all resources belonging to a specific cluster should move from the default to a configurable namespace. But this might not make sense for secrets when we create the namespaces by organization.

stone-z: For sure, I agree with separating them into dedicated namespaces. Since presumably the customer will get access to `default` as well, maybe it makes sense to have an additional namespace per-cluster for things like this which Giant Swarm needs in order to manage the cluster, but which the customer doesn't need access to.

anvddriesch: That sounds like a very good separation. We should have a separate story.
As for the configurable namespace for cluster resources I believe there is nothing more to change in `cluster-operator`

PR #2670, opened 2020-08-14
emit change detection events

njuettner:
emit events in case of a change detection occurs on tccp, tccpn and tcnp stacks

## Checklist


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add release version tag for ec2 instances
 - Update Cloudformation Stack when components version differ
+- Emit Kubernetes events in case of change detection for tccp, tccpn and tcpn CF stacks
```
anvddriesch, line 17:
```suggestion
- Emit Kubernetes events in case of change detection for tccp, tccpn and tcnp CF stacks
```

service/internal/changedetection/tccp.go
```diff
@@ -62,6 +70,7 @@ func (t *TCCP) ShouldUpdate(ctx context.Context, cr infrastructurev1alpha2.AWSCl
 			"message", "detected TCCP stack should update",
 			"reason", fmt.Sprintf("operator version changed from %#q to %#q", cc.Status.TenantCluster.OperatorVersion, key.OperatorVersion(&cr)),
 		)
+		t.event.Emit(ctx, &cr, "CFUpdateRequested", fmt.Sprintf("detected TCCP stack should update:operator version changed from %#q to %#q", cc.Status.TenantCluster.OperatorVersion, key.OperatorVersion(&cr)))
```
anvddriesch, line 73:
```suggestion
		t.event.Emit(ctx, &cr, "CFUpdateRequested", fmt.Sprintf("detected TCCP stack should update: operator version changed from %#q to %#q", cc.Status.TenantCluster.OperatorVersion, key.OperatorVersion(&cr)))
```

anvddriesch:
> I only added events when we check `ShouldUpdate` but I saw we have also `ShouldScale`. Does it make sense to add it there as well?

Generally, having scaling events sounds useful. I'm not sure if we need _should scale_ events in addition but it wouldn't hurt.

CHANGELOG.md
njuettner, line 17:
```suggestion
- Emit Kubernetes events in case of change detection for tccp, tccpn and tcnp CF stacks
```

njuettner:
Tested in gauss, example:

control plane nodes
image

node pool
image

PR #2682, opened 2020-08-18
tls-decrypt-exit-on-error

calvix:
## Checklist

towards https://github.com/giantswarm/giantswarm/issues/12808

### DIFF
--- a/service/internal/cloudconfig/template/decrypt_keys_assets_script.go
+++ b/service/internal/cloudconfig/template/decrypt_keys_assets_script.go
@@ -1,6 +1,8 @@
 package template
 
 const DecryptKeysAssetsScript = `#!/bin/bash -e
+set -o errexit
+
 kms_keys_assets_decrypt() {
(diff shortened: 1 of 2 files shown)


service/internal/cloudconfig/template/decrypt_keys_assets_script.go
```diff
@@ -1,6 +1,8 @@
 package template
 
 const DecryptKeysAssetsScript = `#!/bin/bash -e
+set -o errexit
```
njuettner, line 4: would this be the same as if you would set `set -e`

calvix: True, but this is at least little bit more decriptive

njuettner: Thanks, I haven't seen the longer version of it, that's why I was asking 👍

calvix:
tested on gauss, looks good

PR #2690, opened 2020-08-24
Update k8scc to 7.0.5

tfussell:
## Checklist

### DIFF
--- a/service/controller/internal/cloudconfig/cloud_config_test.go
+++ b/service/controller/internal/cloudconfig/cloud_config_test.go
@@ -7,8 +7,8 @@ import (
 	"testing"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
-	"github.com/giantswarm/k8scloudconfig/v6/pkg/ignition"
-	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v6/pkg/template"
+	"github.com/giantswarm/k8scloudconfig/v7/pkg/ignition"
+	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v7/pkg/template"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger/microloggertest"
 	"github.com/giantswarm/randomkeys"
@@ -49,9 +49,20 @@ func Test_Service_CloudConfig_NewMasterTemplate(t *testing.T) {
 			t.Fatalf("expected %#v got %#v", nil, err)
 		}
 
+		versions := k8scloudconfig.Versions{
+			Calico:                       "3.14.1",
+			CRITools:                     "0.1.0",
+			Etcd:                         "3.4.9",
+			Kubernetes:                   "1.16.12",
+			KubernetesAPIHealthz:         "0.1.1",
+			KubernetesNetworkSetupDocker: "0.2.0",
+		}
+		images := k8scloudconfig.BuildImages("quay.io", versions)
 		data := IgnitionTemplateData{
 			CustomObject: tc.CustomObject,
 			ClusterKeys:  tc.ClusterKeys,
+			Images:       images,
+			Versions:     versions,
 		}
 		template, _, err := ccService.NewMasterTemplate(ctx, data)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


service/controller/internal/cloudconfig/master_template.go
```diff
@@ -46,19 +48,27 @@ func (c *CloudConfig) NewMasterTemplate(ctx context.Context, data IgnitionTempla
...
 		params.DisableEncryptionAtREST = true
 		// Ingress controller service remains in k8scloudconfig and will be
 		// removed in a later migration.
 		params.DisableIngressControllerService = false
-		params.EtcdPort = data.CustomObject.Spec.Cluster.Etcd.Port
+		params.EnableAWSCNI = false
+		params.Etcd = k8scloudconfig.Etcd{
+			ClientPort:          key.EtcdPort(data.CustomObject),
+			HighAvailability:    false,
+			InitialCluster:      "etcd0=https://etcd0." + params.BaseDomain + ":2380",
```
MarcelMue, line 61: I am not sure if this will just work. AFAIK this assumes the structure of HA masters with a single master.

service/controller/internal/templates/cloudformation/tccp/record_sets.go
```diff
@@ -67,6 +67,15 @@ const RecordSets = `
       Name: '{{ $v.EtcdDomain }}.'
       HostedZoneId: !Ref 'HostedZone'
       Type: A
+  EtcdNodeRecordSet:
```
MarcelMue, line 70: This is something I would need to check with e.g. vaclav.

service/controller/internal/templates/cloudformation/tccp/record_sets.go
```diff
@@ -67,6 +67,15 @@ const RecordSets = `
...
+    Type: AWS::Route53::RecordSet
+    Properties:
+      ResourceRecords:
+      - !GetAtt {{ $v.MasterInstanceResourceName }}.PrivateIp
+      Name: 'etcd0.{{ $v.ClusterID }}.k8s.{{ $v.BaseDomain }}.'
+      HostedZoneId: !Ref 'InternalHostedZone'
+      Type: A
+      TTL: '60'
```
MarcelMue, line 78: Why a TTL?

service/controller/internal/cloudconfig/cloud_config_test.go
```diff
@@ -49,9 +49,20 @@ func Test_Service_CloudConfig_NewMasterTemplate(t *testing.T) {
...
 		}
 
+		versions := k8scloudconfig.Versions{
+			Calico:                       "3.14.1",
+			CRITools:                     "0.1.0",
+			Etcd:                         "3.4.9",
+			Kubernetes:                   "1.16.12",
+			KubernetesAPIHealthz:         "0.1.1",
+			KubernetesNetworkSetupDocker: "0.2.0",
+		}
```
stone-z, line 59: Just curious - is there interest / do you think it's worth it to get these from releases in legacy?

service/controller/internal/templates/cloudformation/tccp/record_sets.go
```diff
@@ -67,6 +67,15 @@ const RecordSets = `
       Name: '{{ $v.EtcdDomain }}.'
       HostedZoneId: !Ref 'HostedZone'
       Type: A
+  EtcdNodeRecordSet:
+    Type: AWS::Route53::RecordSet
+    Properties:
+      ResourceRecords:
+      - !GetAtt {{ $v.MasterInstanceResourceName }}.PrivateIp
```
stone-z, line 74: This is only needed at creation, IIUC? Could it happen that this IP changes and then the path can't resolve?

service/controller/internal/templates/cloudformation/tccp/record_sets.go
tfussell: It's required if you specify `ResourceRecords`. AWS API rejects it otherwise. I took this from `aws-operator@master`.

service/controller/internal/cloudconfig/cloud_config_test.go
tfussell: These are just test values--we already implemented "versions from releases" in this operator.

service/controller/internal/templates/cloudformation/tccp/record_sets.go
tfussell: We don't run the master in an ASG in `legacy`. It's a statically defined instance in the CloudFormation stack so the only time the IP would change would be if the master is terminated. We have a complex ops-recipe for this scenario, might be worth adding updating the record IP to the recipe if this PR is accepted.

service/controller/internal/cloudconfig/master_template.go
tfussell: True. In k8scloudconfig before HA masters the only etcd node was called `etcd0` so I believe this is backwards-compatible. At least it worked when I tested creation and upgrades. I'm not sure if there are other scenarios where this could fail.

tfussell:
This is too much work for too little gain. I'm closing in favor of the simpler https://github.com/giantswarm/aws-operator/pull/2702

PR #2693, opened 2020-08-25
Fix routing table names for HA

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/12940
Related PR https://github.com/giantswarm/giantnetes-terraform/pull/380

Improves routing table names for:
- [x] tccp
- [x] tcnp

## Checklist


xh3b4sd:
Just one thing, when we change the templates and its resources we need to make sure that upgrades work fine and that there is as little workload disruption as possible. Cloud Formation might be bitchy about renaming certain resources because it treats it like a replacement. So it destroys the world and creates it from the ashes.

wtty-fool:
@xh3b4sd I've tested the update and it looked fine to my untrained eye. 
Since AWS uses `logical-id` to update CF resources, it did not mind the change of names.

Here are the updated route tables:
image

And here's proof they were updated in the process:
image

Let me know if my reasoning is off or I need to do more digging to ensure safe updates!

xh3b4sd (approved):
Considering the upgrades are fine it should be good to merge. Just for awareness I would like @calvix to acknowledge this change.

wtty-fool:
FYI the final shape of RouteTables section:
image

service/controller/key/common.go
```diff
@@ -72,6 +72,11 @@ func AWSTags(getter LabelsGetter, installationName string) map[string]string {
 	return tags
 }
 
+func AvailabilityZoneRegionSuffix(az string) string {
+	elements := strings.Split(az, "-")
+	return elements[len(elements)-1]
```
calvix, line 77: this could use little bit of comment with an example, I had to read it several times and know what values is going in to figure out what it will return

calvix (approved):

calvix:
LGTM if this was tested also with multiple node pools and updates

wtty-fool:
It works with multiple node pools and updates :) 
However @corest pointed out we'd have to list all subnets in `RouteTables` flag now. We are currently working on improving this before merging this PR.

CHANGELOG.md
```diff
@@ -11,12 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 - Don't panic when AWSControlPlane CR AZs are nil.
+- Add suffix to Route Tables to get rid of naming collision.
 
 ### Changed
 
 - Updated workflows for automatic release.
 - Updated backward incompatible Kubernetes dependencies to v1.18.5.
 - Removed migration code to ensure the Control Plane CRs for existing Node Pool clusters.
+- Add Route Table lookup using tags, so `RouteTables` flag can be phased out in the future.
```
kopiczko, line 21: There is a `### Deprecated` section. We can add a note about `RouteTables` flag there.

Till we don't break the compatibility I think we can add this entry to `### Added` and once we remove `RouteTables` it will get added to `### Removed` and that will be the breaking change.

Headers.

service/controller/resource/cproutetables/resource.go
```diff
@@ -37,15 +44,18 @@ func New(config Config) (*Resource, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
 
-	if len(config.Names) == 0 {
-		return nil, microerror.Maskf(invalidConfigError, "%T.Names must not be empty", config)
+	// TODO: Get installation from flags
```
kopiczko, line 47: Is it a TODO still?

service/controller/resource/cproutetables/resource.go
```diff
@@ -66,43 +76,44 @@ func (r *Resource) addRouteTablesToContext(ctx context.Context) error {
 		return microerror.Mask(err)
 	}
 
-	// We check if we have all route tables cached for the configured route table
-	// names. If we find all information, we return them.
-	{
-		r.logger.LogCtx(ctx, "level", "debug", "message", "finding cached route tables")
```
kopiczko, line 72: Shouldn't this log message stay?

service/controller/resource/cproutetables/resource.go
```diff
@@ -66,43 +76,44 @@ func (r *Resource) addRouteTablesToContext(ctx context.Context) error {
...
 
 			r.routeTables = append(r.routeTables, rt)
 		}
-
-		r.logger.LogCtx(ctx, "level", "debug", "message", "cached route tables")
 	}
+	r.logger.LogCtx(ctx, "level", "debug", "message", "cached route tables")
 
 	cc.Status.ControlPlane.RouteTables = r.routeTables
+	r.expectedTableCount = len(r.routeTables)
```
kopiczko, line 111: Does it make sense to have that `r.expectedTableCount`? I feel like it's only testing if `len` returns the same value for the same slice.

wtty-fool:
I updated RouteTable caching. The idea is as follows:
**Short term**
When there are `names` supplied via `RouteTables` flag (which is what we do currently), we use them to look up RouteTables just like before. The new part is `lookupByTag`, which works when no `names` are given. It uses tags already existing in TC RouteTables to find all matching RouteTables. This way `aws-operator` works both on 1CP and 2CP out of the box.

**Long term**
We add a missing tag (`giantswarm.io/route-table-type`) to control planes https://github.com/giantswarm/giantnetes-terraform/pull/380. When all CPs are eventually rolled, and all route tables are tagged, we can drop `RouteTableNames` parameter from installations repo https://github.com/giantswarm/installations/blob/master/ginger/draughtsman-configmap-values.yaml#L39 and `routeTables` from configmap https://github.com/giantswarm/aws-operator/blob/master/helm/aws-operator/templates/configmap.yaml#L28

service/controller/resource/cproutetables/resource.go
wtty-fool: RouteTables don't change over time. If we get rid of `r.expectedTableCount`, we are hitting AWS API every time to get the same list of tables. This way we ask once and perform lookup again if something changes.

kopiczko: I'm asking if we can replace:

```
if len(r.routeTables) == r.expectedTableCount && r.expectedTableCount > 0 {
```

with:

```
if len(r.routeTables) > 0 {
```

and get rid of `r.expectedTableCount`.

wtty-fool: Smart, I'll do that.

kopiczko (approved):
LGTM

wtty-fool:
@xh3b4sd Pinging you in case you want to take another look before I merge :)

xh3b4sd (approved):
I think that in case your collaboration ensures that the cluster transitions work fine this is good to go. Thanks a lot everyone.

PR #2700, opened 2020-08-27
Assume role for CP account

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/12945

This is to unify access to AWS account to CP and TC. The reason is following:

- We want to be able to create a CP using aws-operator.
- When running aws-operator in newly created TC we can assume role used to create the TC and we don't need to configure anything else.
- To make it compatible with the current setup we want to assume role of the same account.

## Checklist


xh3b4sd (approved):

calvix:
KIAM needs to be deployed to AWS CP than and we should add the aws-operator role to the nodes that will run  kiam server

kopiczko:
> KIAM needs to be deployed to AWS CP than and we should add the aws-operator role to the nodes that will run kiam server
> 
> that needs to be done all AWS CPs

For the record we talked with @calvix. He thought we want to not use credentials, but we do. We only want to create some level of comatibility but no new features.

corest (approved):

policies/tenant_cluster.json
```diff
@@ -58,9 +58,10 @@
             ],
             "Resource": [
                 "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-EC2-K8S-Role",
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-IAMManager-Role",
                 "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-Route53Manager-Role",
-                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/gs-*",
-                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-IAMManager-Role"
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/*-vpc-peer-access",
+                "arn:aws:iam::<CUSTOMER_ACCOUNT_ID>:role/gs-*"
```
kopiczko, line 64: I added only `/*-vpc-peer-access` here. The rest is sorting. aws-operator is creating *-vpc-peer-access role.

kopiczko:
Tested in EU & CN including upgrades from 12.3.0.

PR #2701, opened 2020-08-27
Events for tccpi and tccpf CF stack failures

njuettner:
Since tccpi and tccpf stacks are created in controlplane account, I think it would make sense to show failures on the AWSCluster CR in case we have some issues. It would help us and the customer to see what's going on easily.

This will write K8s events in case something is wrong.

I'll create a separate PR for tcnpf stacks to make this PR a bit smaller.

## Checklist


njuettner:
tested in gauss

xh3b4sd (approved):

PR #2760, opened 2020-09-23
update flatcar AMI for China

njuettner:
Flatcar AMI for China was not present at the moment, this needs to be in v12.2 before we release it.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


paurosello (approved):

paurosello:
Would this mean we need to release 11.5.6 also?

calvix (approved):

njuettner:
Yes, this means we have to fix it for 11.5.6. The reason is https://github.com/giantswarm/flatcar-china/releases is a bit behind, I didn't saw the whole picture. So it means Adidas is currently broken in China they cannot update to this release there.

PR #2800, opened 2020-10-13
Fix cpvpc && routetable resources

corest:
While deploying aws-operator in tenant-cluster, we get into the issue where current lookup process for VPC/routing tables doesn't work with tenant cluster tags.

With this change lookup is done by name instead of inconsistent tags. So that it works in CP and TC

## Checklist

### DIFF
--- a/service/controller/resource/cpvpc/resource.go
+++ b/service/controller/resource/cpvpc/resource.go
@@ -100,35 +100,26 @@ func (r *Resource) lookup(ctx context.Context, client EC2, installationName stri
 	var vpcCIDR string
 	var vpcID string
 	{
-		i := &ec2.DescribeVpcsInput{}
-
-		o, err := client.DescribeVpcs(i)
-		if err != nil {
-			return "", "", microerror.Mask(err)
-		}
-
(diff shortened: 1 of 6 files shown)


corest:
Tested in TC - all good

wtty-fool (approved):

service/controller/resource/cpvpc/resource.go
```diff
@@ -105,15 +105,15 @@ func (r *Resource) lookup(ctx context.Context, client EC2, installationName stri
...
 					Values: []*string{
 						aws.String(installationName),
 					},
 				},
 				{
-					Name: aws.String(fmt.Sprintf("tag:%s", key.TagClusterType)),
+					Name: aws.String(fmt.Sprintf("tag:%s", key.TagCluster)),
 					Values: []*string{
-						aws.String(key.TagClusterTypeControlPlane),
+						aws.String(installationName),
```
kopiczko, line 116: Do we need the previous condition when the value is the same?

kopiczko (approved):

PR #2881, opened 2020-12-03
enable vpa for aws-operator

njuettner:
By enabling the VPA auto recommender, we would have the benefit of setting the right requests to our aws-operator. It will handle OOM's and adjust the right settings to those pods and also reduces the requests in some aws-operators which don't have to much to do.

## Checklist


njuettner:
Tested for some of the crashing operator due to high memory consumption (operators have to fetch a lot of instances in a lot of clusters), VPA fixes OOM's and set's the right requests for those pods. 

It also helps us to reduce CPU and memory settings when certain operators don't have much to do.

Example how one recommendation would look like (see target, this is what the vpa would set for this operator):

```
Name:         aws-operator-9-2-0
Namespace:    giantswarm
Labels:       <none>
Annotations:  <none>
API Version:  autoscaling.k8s.io/v1
Kind:         VerticalPodAutoscaler
Spec:
  Target Ref:
    API Version:  apps/v1
    Kind:         Deployment
    Name:         aws-operator-9-2-0
  Update Policy:
    Update Mode:  Off
Status:
  Conditions:
    Last Transition Time:  2020-12-03T15:20:47Z
    Status:                True
    Type:                  RecommendationProvided
  Recommendation:
    Container Recommendations:
      Container Name:  aws-operator
      Lower Bound:
        Cpu:     15m
        Memory:  442693543
      Target:
        Cpu:     109m
        Memory:  628694953
      Uncapped Target:
        Cpu:     109m
        Memory:  628694953
      Upper Bound:
        Cpu:     205m
        Memory:  864505777
Events:          <none>
```

calvix (approved):

anvddriesch (approved):

calvix:
should we merge this and include in release?  think it might make sense as well

njuettner:
@calvix Yes please do 👍 , I fixed the changelog issue

PR #2882, opened 2020-12-04
Using new micrologger 0.4.0

tomahawk28:
Introducing new micrologger functions so make logs simpler.


kopiczko (approved):
I skimmed over. Looks sane.

tomahawk28:
Tested in `ginger`

PR #2927, opened 2021-01-15
add-retry-logic-to-decrypt-units

calvix:
## Checklist

towards https://github.com/giantswarm/giantswarm/issues/14836

### DIFF
--- a/service/internal/cloudconfig/template/decrypt_keys_assets_script.go
+++ b/service/internal/cloudconfig/template/decrypt_keys_assets_script.go
@@ -14,7 +14,7 @@ done
 echo "Successfully fetched docker image ${AWS_CLI_IMAGE}."
 
 
-docker run --net=host -v /etc/kubernetes/encryption:/etc/kubernetes/encryption \
+while ! docker run --net=host -v /etc/kubernetes/encryption:/etc/kubernetes/encryption \
         --entrypoint=/bin/sh \
         ${AWS_CLI_IMAGE} \
         -ec \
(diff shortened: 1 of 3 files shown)


service/internal/cloudconfig/template/decrypt_keys_assets_script.go
```diff
@@ -14,23 +14,28 @@ done
 echo "Successfully fetched docker image ${AWS_CLI_IMAGE}."
 
 
-docker run --net=host -v /etc/kubernetes/encryption:/etc/kubernetes/encryption \
+while ! docker run --net=host -v /etc/kubernetes/encryption:/etc/kubernetes/encryption \
```
calvix, line 17: whole decrypt logic is retired until it succeed

service/internal/cloudconfig/template/decrypt_keys_assets_script.go
```diff
@@ -14,23 +14,28 @@ done
...
       aws \
         --region {{.AWSRegion}} kms decrypt \
         --ciphertext-blob fileb://$encKey \
         --output text \
-        --query Plaintext \
-      | base64 -d > $f
-      mv -f $f ${encKey%.enc}
+        --query Plaintext > $f
+      base64 -d $f > $f2
+      mv -f $f2 ${encKey%.enc}
```
calvix, line 33: splitting `aws` and `base64`  connected with `pipe` into two command  to avoid error hiding
( if `aws` failed than and `base64` succeeded than  result of this command was `0` thus not failing)

calvix: so first run aws command and output into tem file 1
than use base64 and output into temfile 2

service/internal/cloudconfig/template/decrypt_tls_assets_script.go
```diff
@@ -14,23 +14,29 @@ done
 echo "Successfully fetched docker image ${AWS_CLI_IMAGE}."
 
 
-docker run --net=host -v /etc/kubernetes/ssl:/etc/kubernetes/ssl \
+while ! docker run --net=host -v /etc/kubernetes/ssl:/etc/kubernetes/ssl \
```
calvix, line 17: completely same changes as for the decrypt key assets

calvix:
works nice, now the decryption will retry until succeeded, the previous error is now cautch

xh3b4sd (approved):

njuettner (approved):
lgtm

paurosello:
should systemd already restart a failed system unit? I somehow thought the unit was not configured correctly or the command was not returning an error value and that was the reason it was not retried

calvix:
> should systemd already restart a failed system unit? I somehow thought the unit was not configured correctly or the command was not returning an error value and that was the reason it was not retried

systems units type `oneshot` do not support retry logic in systems definition so we need to implement it another way

https://unix.stackexchange.com/questions/272313/how-to-assure-a-oneshot-service-gets-retried-if-it-fails-the-first-time

PR #2933, opened 2021-01-18
detach role policies upon node pool deletion

xh3b4sd:
## Checklist

## Context 

Fixes https://github.com/giantswarm/giantswarm/issues/14691. I tested the cleanup manually in a little go script. There should never be more than a handful attached policies on a given role.

### DIFF
--- a/service/controller/key/machine_deployment.go
+++ b/service/controller/key/machine_deployment.go
@@ -152,6 +152,10 @@ func MachineDeploymentWorkerCountRatio(workers int, ratio float32) string {
 	return strconv.Itoa(rounded)
 }
 
+func NodeRole(cr infrastructurev1alpha2.AWSMachineDeployment) string {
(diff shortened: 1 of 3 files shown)


njuettner (approved):

calvix:
how about polices attached to tccpn role? I think we should clean them too

calvix:
Also, I miss code where you  add this resource to any operator kit controller so this code is now unused

xh3b4sd:
I split my PRs. I added and wired the resource in https://github.com/giantswarm/aws-operator/pull/2932. The PM was only about the node role of the node pools. I can add the same logic for the control plane.

calvix:
> I split my PRs. I added and wired the resource in #2932. The PM was only about the node role of the node pools. I can add the same logic for the control plane.

OK I missed that previous PR, so the code is good for NP but yeah I would also like to see it for the control plane as well as I think it makes a lot of sense

calvix (approved):
LGTM for NP, lets tackle the COntrol plane in another PR

PR #2939, opened 2021-01-21
Add new flatcar AMIs

MarcelMue:
## Checklist


calvix:
sure but AFAIk new flatcar as new systems which causes issues with was CNI right?

calvix (approved):

stone-z (approved):

MarcelMue:
> sure but AFAIk new flatcar as new systems which causes issues with was CNI right?

True, but we managed to solve these issues for Azure and mostly KVM. So I want to also make it work for AWS.
Testing first though, obviously.

njuettner (approved):

PR #2973, opened 2021-02-10
clean-vpc-peering-when-deleting-cluster

calvix:
towards https://github.com/giantswarm/giantswarm/issues/15647

## Checklist


service/controller/resource/cleanupvpcpeerings/delete.go
```diff
@@ -0,0 +1,123 @@
...
+
+				vpcPeeringConnectionIDs = append(vpcPeeringConnectionIDs, s.VpcPeeringConnectionId)
+			}
+		}
+		// fetch all vpc peering connection via accepter vpc filter
+		{
+			i := &ec2.DescribeVpcPeeringConnectionsInput{
+				Filters: []*ec2.Filter{
+					{
+						Name: aws.String("accepter-vpc-info.vpc-id"),
```
njuettner, line 67: does it make sense do have it in a small helper function? it seems it's duplicate code except the name is different.

njuettner (commented):
just a nit, not sure if that would make sense what I wrote. let me know

xh3b4sd (approved):
I am happy when Nick is happy.

service/controller/resource/cleanupvpcpeerings/delete.go
calvix: I have adjusted the code, can you check now?

njuettner (approved):
🚀

PR #2992, opened 2021-03-01
Update architect-orb to 2.1.0

ubergesundheit:
Update architect-orb to 2.1.0 which will use giantswarm.github.io as catalog base url.

Towards giantswarm/giantswarm#15898


ubergesundheit:
Merging with ping @giantswarm/team-firecracker

njuettner (approved):

PR #2993, opened 2021-03-04
Remove dockershim mount

yulianedyalkova:
## Checklist


calvix:
Are you certain that the `aws-cni` support running with `containerd` instead of docker shim?

I would like to see some test and results :p

yulianedyalkova:
> Are you certain that the `aws-cni` support running with `containerd` instead of docker shim?
> 
> I would like to see some test and results :p

You're too fast, I was just upgrading the description of the PR 😆 I tested the change on `ginger` and managed to successfully create a cluster. All `aws-node` pods were running happily.

calvix:
What about new pods that need an IP? did you tested that as well? ( it wait until the cluster is fully created and creates a custom pod and sees if it gets an IP assigned and also if it has connectivity. Did you test that as well? :P

With CNI we need to be careful :p

yulianedyalkova:
I saw all the exporter pods running and they were reporting no errors so I assumed that's maybe enough? 

I also checked the code and that gives me some level of confidence - the only mention of the `dockershim` socket is the one linked and `containerd` has a `cri` plugin so the socket should definitely be usable.

calvix:
when I tried this on a fresh CGS cluster with docker (without contained), I was not able to create a new pod, the pods got stuck as there was no IP assigned to them

when I added both docker sock and cri sock it, pods started  without issue

whites11 (approved):

T-Kukawka (approved):

PR #3019, opened 2021-04-13
fix-etcd-init-failure

calvix:
towards https://github.com/giantswarm/giantswarm/issues/15814
## Checklist

### DIFF
--- a/service/controller/key/aws_control_plane.go
+++ b/service/controller/key/aws_control_plane.go
@@ -35,9 +35,6 @@ func ControlPlaneENIResourceName(id int) string {
 }
 
 func ControlPlaneEtcdNodeName(id int) string {
-	if id == 0 {
-		return "etcd"
-	}
 	return fmt.Sprintf("etcd%d", id)
 }
 
--- a/service/internal/cloudconfig/tccpn.go
+++ b/service/internal/cloudconfig/tccpn.go
@@ -354,6 +354,10 @@ func (t *TCCPN) newTemplate(ctx context.Context, obj interface{}, mapping hamast
 			HighAvailability:    multiMasterEnabled,
 			NodeName:            key.ControlPlaneEtcdNodeName(mapping.ID),
 		}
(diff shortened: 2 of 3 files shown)


service/controller/key/aws_control_plane.go
```diff
@@ -35,9 +35,6 @@ func ControlPlaneENIResourceName(id int) string {
 }
 
 func ControlPlaneEtcdNodeName(id int) string {
-	if id == 0 {
-		return "etcd"
-	}
```
calvix, line 40: dropping this means we no longer use ELB  dns (`etcd.clusterid.k8s.basedomain.gigantic.io`) for DNS initial state, but instead node DNS `etcd0.clusterid.k8s.basedomain.gigantic.io`

service/internal/cloudconfig/tccpn.go
```diff
@@ -354,6 +354,10 @@ func (t *TCCPN) newTemplate(ctx context.Context, obj interface{}, mapping hamast
 			HighAvailability:    multiMasterEnabled,
 			NodeName:            key.ControlPlaneEtcdNodeName(mapping.ID),
 		}
+		// we need to explicitly set InitialCluster for single master, since k8scc qhas different config logic which does nto work for AWS
+		if !multiMasterEnabled {
+			params.Etcd.InitialCluster = fmt.Sprintf("%s=https://%s.%s:2380", key.ControlPlaneEtcdNodeName(mapping.ID), key.ControlPlaneEtcdNodeName(mapping.ID), key.TenantClusterBaseDomain(cl))
```
anvddriesch, line 359: Just for my understanding, what happens when this value is not explicitly set? (In the multi master case)

anvddriesch (approved):

service/internal/cloudconfig/tccpn.go
calvix: k8scc is checking the value and if its empty i will fill it with its internal logic, but resulting in issue inc ase of single master in AWS

PR #3052, opened 2021-05-10
disable etcd3-attach-dependencies.service

ericgraf:
do not merge

debugging for new flatcar release


calvix:
flatcar was sorted, just cleaning up

PR #2979, opened 2021-02-15
Enable AWS CSI migration

njuettner:
## Checklist

- [x] Update changelog in CHANGELOG.md.

PR #3053, opened 2021-05-10
adjust etcd3-attach-deps for new flatcar  and add k8s 1.20 support from k8scc

calvix:
## Checklist

to avoid bug https://github.com/systemd/systemd/issues/18108, etcd3-attach deps will use `networkctl reload` instead of `systemctl restart systemd-networkd`

This si needed in order to run latest flatcar

### DIFF
--- a/service/internal/cloudconfig/template/etcd3_attach_dep_service.go
+++ b/service/internal/cloudconfig/template/etcd3_attach_dep_service.go
@@ -28,6 +28,7 @@ ExecStart=/bin/bash -c "docker run --rm -i \
       --volume-tag-key=Name \
       --volume-tag-value={{ .MasterEtcdVolumeName }}"
 ExecStartPost=/usr/bin/systemctl daemon-reload
+ExecStartPost=/usr/bin/ip set link eth1 down
 ExecStartPost=/usr/bin/systemctl restart systemd-networkd
 
 [Install]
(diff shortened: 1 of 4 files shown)


anvddriesch (approved):
Looks fine to me if it is tested.

service/internal/cloudconfig/template/etcd3_attach_dep_service.go
```diff
@@ -27,8 +27,9 @@ ExecStart=/bin/bash -c "docker run --rm -i \
       --volume-device-label=etcd \
       --volume-tag-key=Name \
       --volume-tag-value={{ .MasterEtcdVolumeName }}"
-ExecStartPost=/usr/bin/systemctl daemon-reload
-ExecStartPost=/usr/bin/systemctl restart systemd-networkd
+# use 'networkctl reload'' instead restarting systemd-networkd to avoid bug in systemd
```
anvddriesch, line 30:
```suggestion
# use 'networkctl reload' instead of restarting systemd-networkd to avoid bug in systemd 
```

njuettner:
There's a PR https://github.com/giantswarm/k8scloudconfig/pull/879 and https://github.com/giantswarm/aws-operator/pull/2979 which also needs to be in k8s 1.20 this is for EBS.

calvix:
> There's a PR [giantswarm/k8scloudconfig#879](https://github.com/giantswarm/k8scloudconfig/pull/879) and #2979 which also needs to be in k8s 1.20, it is for EBS integration via `csi-driver`. Not sure how we want to do this since I'm off for the next days.

I think I will merge this now and the CSI changes can be merged separately

PR #3067, opened 2021-05-20
Avoid TCCPN stack failure by checking if a control-plane tag exists

anvddriesch:
Towards: https://github.com/giantswarm/giantswarm/issues/17073
We want to check if the tag actually exists before adding it into the tccpn stack template and risking failure in case it does not.

## Checklist


calvix (approved):
lgtm

PR #3105, opened 2021-06-23
update flatcar amis to the latest stable releases

njuettner:
## Checklist


njuettner:
Also ensured the AMI's in China are present.

calvix (approved):
lgtm but I would be hesitant with using latest flatcar in the 15.0.0 release as we encountered a bug  with clusters that have many services

njuettner:
Sure, we can still merge it right? Even using a new operator release wouldn't do anything as long we don't change the flatcar version inside the release, right?

calvix:
yes

PR #3177, opened 2021-09-29
s3-propagate-tags

calvix:
This is `poor man` s3 bucket tag propagation.
The creation of S3 bucket is almost the same we just append the cloud tags from cluster CR

But for the update, we did not have any logic to determine the changes, to make it super simple in each update loop we simply update the tags.

I was thinking about checking if the tags are there before updating but that would be an extra AWS API call and in the end even more expensive than just blindly set the tags to desired value on every loop.

I just wanted to get rid of this ASAP  as it's just one customer who needs this.

towards https://github.com/giantswarm/giantswarm/issues/18887

### DIFF
--- a/service/controller/resource/s3bucket/resource.go
+++ b/service/controller/resource/s3bucket/resource.go
@@ -1,10 +1,15 @@
 package s3bucket
 
 import (
+	"context"
 	"github.com/aws/aws-sdk-go/service/s3"
 	infrastructurev1alpha3 "github.com/giantswarm/apiextensions/v3/pkg/apis/infrastructure/v1alpha3"
+	"github.com/giantswarm/aws-operator/pkg/label"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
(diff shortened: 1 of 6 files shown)


service/controller/resource/s3bucket/resource.go
```diff
@@ -97,11 +106,48 @@ func containsBucketState(bucketStateName string, bucketStateList []BucketState)
...
+		client.MatchingLabels{label.Cluster: key.ClusterID(&customObject)},
+	)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	if len(list.Items) != 1 {
+		return nil, microerror.Maskf(executionFailedError, "expected 1 CR got %d", len(list.Items))
+	}
+
+	allTags := addCloudTags(tags, list.Items[0].GetLabels())
```
calvix, line 125: here we take the usual `tags`  and append the cloud tags to the map

calvix:
thoughts, ideas?

anvddriesch (commented):
LGTM if tested

anvddriesch (approved):

PR #3235, opened 2022-01-20
use-encryption-provider-operator

calvix:
changes needed in order to fully work with `encryption-provider-operator`


njuettner (approved):
lgtm

PR #3242, opened 2022-01-31
force cgroup v1

calvix:
Add feature to force cgroups v1 for AWS to specific node pools

### DIFF
--- a/service/internal/cloudconfig/tcnp.go
+++ b/service/internal/cloudconfig/tcnp.go
@@ -3,13 +3,13 @@ package cloudconfig
 import (
 	"context"
 	"fmt"
-	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	"net"
 	"sync"
 
 	infrastructurev1alpha3 "github.com/giantswarm/apiextensions/v3/pkg/apis/infrastructure/v1alpha3"
 	"github.com/giantswarm/certs/v3/pkg/certs"
 	k8scloudconfig "github.com/giantswarm/k8scloudconfig/v11/pkg/template"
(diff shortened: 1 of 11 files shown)


service/internal/cloudconfig/tcnp.go
```diff
@@ -92,6 +94,36 @@ func (t *TCNP) NewTemplates(ctx context.Context, obj interface{}) ([]string, err
...
+		md = list.Items[0]
+	}
+
+	var forceCGroupsV1 bool
+	{
+		if _, ok := md.Annotations[annotation.NodeForceCGroupsV1]; ok {
+			forceCGroupsV1 = true
+		} else {
+			forceCGroupsV1 = false
+		}
```
njuettner, line 124: We could drop those lines, but only if you prefer it

njuettner (approved):

service/internal/cloudconfig/tcnp.go
```diff
@@ -92,6 +94,36 @@ func (t *TCNP) NewTemplates(ctx context.Context, obj interface{}) ([]string, err
...
+	}
```
AverageMarcus, line 125:
```suggestion
	_, forceCGroupsV1 := md.Annotations[annotation.NodeForceCGroupsV1]
```

AverageMarcus: Minor suggestion to save lines. I don't mind either way.

calvix: I was thinking about this, but it might be harder to read, but no strong feelings

PR #3278, opened 2022-03-15
Switch gp2 to gp3

njuettner:
## Checklist


calvix (changes requested):
I don't think we can simply do this,  AFAIk gp2 cannot be seamlessly upgraded to gp3, you need to migrate to the new volumes. This would break CF as it would detect change that cannot be applied.

Fell free to test the update and prove me wrong

calvix (approved):

whites11 (approved):
if this works it's awesome!

paurosello:
I love it!

calvix:
what is the situation in china? is gp3 also available in china?

njuettner:
gp3 is available in both regions in CN

service/controller/resource/tccpn/template/template_main_etcd_volume.go
```diff
@@ -16,6 +16,12 @@ const TemplateMainEtcdVolume = `
       - Key: Name
         Value: {{ .Name }}
       VolumeType: gp3
+      {{- if and (ge .Iops 3000) (le .Iops 16000) }}
```
whites11, line 19: how are those bounds defined? do they change by disk size and or instance type?

njuettner: those bounds are set by volume type. according to the docs we are bound to those numbers if we use gp3, see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-blockdevicemapping-ebs.html

whites11 (approved):

PR #3340, opened 2022-05-25
bump Etcd attach deps

whites11:
Towards https://github.com/giantswarm/giantswarm/issues/22199

Bump to latest version of `aws-attach-etcd-dep`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump `k8scc` to use `systemd` cgroup driver on masters and cgroups v2 worker nodes.
+- Bump `aws-attach-etcd-dep` to 0.4.0.
 
 ## [11.11.0] - 2022-05-16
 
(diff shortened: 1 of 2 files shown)


whites11:
jump a bump, will merge without review

whites11:
The error is in the AWS API call and this is about retrying that. Why don't you think it can fix it?

calvix:
I might be wrong since I don't have the full info on this issue but It works like this:
Before the attach phase. the app is trying to detach the volume from the other instance and then it's waiting until it sees that it is detached - ie in an `available` state ( via backoff retry logic - 240 retries, wait before retry 15 sec) if this didn't worked for some reason than the attachment will fail with an error that is described in the issue, and 5 extra retries of attaching will most probably not do anything since it already retried for 240 times

It probably won't hurt anything as long as the code executes properly, but it might not fix the desired issue.

whites11:
> and 5 extra retries of attaching will most probably not do anything since it already retried for 240 times

Nope. Attaching the disk involves 2 API calls.
One was actually retried 240 times as you mentioned, the other one (the one failing in the logs above) was tried only once.
My PR makes that second one retry 5 times.

PR #3388, opened 2022-08-02
Cidr

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/23061

This PR makes aws-operator use an annotation in `AWSCluster` CR as CIDR for cilium.
That is because we need a different CIDR compared to what AWS-CNI is using.

For new clusters there is no need for this and the annotation is missing.

Presence of annotation in the places where it's needed is mandated to aws-admission-controller

## Checklist


njuettner (approved):
lgtm

PR #3392, opened 2022-08-04
Use cloudfront for IRSA

njuettner:
Issue: https://github.com/giantswarm/roadmap/issues/1255

This will potentially break existing clusters when IRSA is already enabled. 

If `irsa-operator` can create Cloudfront before the upgrade, we only need to delete the OIDC configuration file in S3 bucket manually (it will be created again with the correct OIDC config). Afterwards everything should work again.

- Change API flag for IRSA service account issuer to Cloudfront Endpoint
- Change Route53Manager IAM Role to Cloudfront Endpoint

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Use Cloudfront Domain for IRSA.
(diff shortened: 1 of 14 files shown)


whites11 (approved):
Code looks good.
I'm puzzled about the need of manual intervention to fix clusters.
Can't we do something to avoid that? We have hundreds of clusters in AWS

njuettner:
The OIDC configuration only needs deleted where IRSA is currently activated (17 clusters atm)

PR #3393, opened 2022-08-08
Cilium annotation cluster

whites11:
Use the Cluster CR rather than the AWSCluster CR to get the cilium CIDR during v17 to v18 upgrades

## Checklist


njuettner (approved):
Would that be temporary looking on the Cluster CR for CNI annotation or only for the migration process? 
Would that be the same on CAPI, CNI seems to be related to infrastructure CR's?

whites11:
> Would that be temporary looking on the Cluster CR for CNI annotation or only for the migration process? Would that be the same on CAPI, CNI seems to be related to infrastructure CR's?

This is only needed during migration. For the following release we can remove a lot of code from aws-operator and looking up of the Cluster CR can be removed everywhere.

CAPI is not involved at all.

njuettner:
Nice, thank you ❤️. Feel free to merge 🙂

PR #3421, opened 2022-09-27
Add permission for autoscaling healthcheck

njuettner:
https://github.com/giantswarm/giantswarm/issues/23752

## Checklist

### DIFF
--- a/service/controller/resource/tccpn/template/template_main_iam_policies.go
+++ b/service/controller/resource/tccpn/template/template_main_iam_policies.go
@@ -46,7 +46,6 @@ const TemplateMainIAMPolicies = `
             Resource: "*"
           - Effect: "Allow"
             Action:
-              - "autoscaling:CompleteLifecycleAction"
               - "autoscaling:DescribeAutoScalingGroups"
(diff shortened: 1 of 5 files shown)


service/internal/cloudconfig/template/instance_health_check.go
```diff
@@ -0,0 +1,136 @@
+package template
+
+const MasterInstanceLifeCycleCompletionService = `
+[Unit]
+Description=master-instance-lifecyle-completion job
+After=docker.service etcd3.service k8s-kubelet.service k8s-setup-network-env.service
+Requires=docker.service etcd3.service k8s-kubelet.service k8s-setup-network-env.service
+[Service]
+ExecStartPre=/bin/bash -c "while [ ! -f /opt/bin/master-instance-healthcheck ]; do /opt/bin/master-instance-healtcheck && sleep 1; done"
```
whites11, line 9: here you rely on the exit status of the script to know if you can move on with the unit execution or not.
If I'm not reading it wrong, it seems to me the script always exits with status code 0, even when there is an error.
If I'm right this check will never work properly

PR #3423, opened 2022-10-10
Cilium kube proxy migration hacks

whites11:
This PR introduces another bunch of hacks needed to smoothly migrate between aws-cni and cilium with kube proxy replacement.

More specifically, this PR:
- manages an annotation in the Cluster CR in order to enable or disable cilium's kube-proxy feature during upgrade. Timing is very important so aws-operator has the controller role in this process.
- adds a sidecar container to aws-node pods that adjusts the node's routing tables in order for aws-cni pods to be able to connect to cilium pods during migration

Sorry this is a big PR, but there is no meaningful way to split it.

## Checklist

### DIFF
--- a/service/controller/key/cilium.go
+++ b/service/controller/key/cilium.go
@@ -1,15 +1,12 @@
 package key
 
 import (
+	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	apiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-const (
-	CiliumForceDisableKubeProxyAnnotation = "cilium-force-disable-kube-proxy-replacement"
-)
-
 func ForceDisableCiliumKubeProxyReplacement(cluster apiv1beta1.Cluster) bool {
-	v, found := cluster.Annotations[CiliumForceDisableKubeProxyAnnotation]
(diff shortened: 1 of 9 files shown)


T-Kukawka:
i will leave this review for Nick

service/controller/key/cilium.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/giantswarm/k8smetadata/pkg/annotation"
+	apiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+func ForceDisableCiliumKubeProxyReplacement(cluster apiv1beta1.Cluster) bool {
+	v, found := cluster.Annotations[annotation.CiliumForceDisableKubeProxyAnnotation]
+
+	return found && v == "true"
```
njuettner, line 11: Do we wanna check the value or would it be enough to just look if the annotation is set?

njuettner (approved):
lgtm

service/controller/key/cilium.go
whites11: I would be confused if I see the annotation set to "false" but the behaviour still happening.
I know elsewhere we only check the presence of the annotation, but I think it's best to be explicit and check the value as well.

njuettner: Makes sense 👍🏻

PR #3428, opened 2022-10-19
Enable IRSA by default in release v19.0.0

njuettner:
Issue: https://github.com/giantswarm/roadmap/issues/1095

## Checklist


service/controller/key/common.go
```diff
@@ -180,6 +183,11 @@ func IsDeleted(getter DeletionTimestampGetter) bool {
 	return getter.GetDeletionTimestamp() != nil
 }
 
+func IsV19Release(releaseVersion *semver.Version) bool {
```
whites11, line 186: How about checking the major version is v19 or higher? I'm afraid comparison of patch versions might be tricky and unreliable

njuettner: Agree, I slightly changed it to `return releaseVersion.Major >= v19.Major`. Is this better for you?

whites11: Yes perfect

whites11 (approved):
Lgtmiiwfy

PR #3429, opened 2022-10-21
Add cluster domain as sts audience

whites11:
Add cluster domain as sts audience

## Checklist


njuettner (approved):

whites11:
@paurosello if we don't need another STS audience when we change the cloudfront domain then this should be enough, right?

paurosello:
yeah, finishing tests but looks perfect

PR #3435, opened 2022-11-03
Use custom KMS key for encryption on your Amazon EBS volumes

njuettner:
## Checklist

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "aws-operator"
 	source      string = "https://github.com/giantswarm/aws-operator"
-	version            = "14.0.1-iamrole"
+	version            = "14.0.1-dev"
 )
 
 func Description() string {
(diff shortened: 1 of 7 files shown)


paurosello:
I think we need to allow access to all kms keys with "*" as they could swap the KMS keys anytime on AWS and then the cluster will use multiple, each EBS can be encrypted with a different key.

Also we can remove the conditional as it will always be true, we generate a KMS key by default

njuettner:
That sounds good Pau, let's do that 👍🏻

njuettner (commented):
looks good to me, unfortunately I can't approve my own PR :D

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "aws-operator"
 	source      string = "https://github.com/giantswarm/aws-operator"
-	version            = "14.0.1-dev"
+	version            = "14.0.1-iamrole"
```
whites11, line 8: this doesn't have to be merged, please

whites11 (approved):
LGTM

pkg/project/project.go
paurosello: yeah, reverting back now

paurosello:
tested on gauss and works

PR #3440, opened 2022-11-14
Cloudfront alias

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/1546

we added a custom domain to the cloudfront distribution we use for irsa, but we didn't edit the IAM role that allows assuming the route53 manager role when authenticated through IRSA with the new domain name.
This PR fixes this problem

## Checklist

### DIFF
--- a/service/controller/resource/tccpn/create.go
+++ b/service/controller/resource/tccpn/create.go
@@ -442,6 +442,7 @@ func (r *Resource) newIAMPolicies(ctx context.Context, cr infrastructurev1alpha3
 	}
 
 	var cloudfrontDomain string
+	var cloudfrontAliasDomain string
 	var cloudfrontEnabled bool
 	{
 		// ignore China, Cloudfront does not work.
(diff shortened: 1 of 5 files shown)


service/controller/resource/tccpn/create.go
```diff
@@ -458,16 +459,8 @@ func (r *Resource) newIAMPolicies(ctx context.Context, cr infrastructurev1alpha3
...
-					cloudfrontDomain = cm.Data["domainAlias"]
-					if cloudfrontDomain == "" {
-						// Default to cloudfront domain if there is no alias.
-						cloudfrontDomain = cm.Data["domain"]
-						if cloudfrontDomain == "" {
-							return nil, microerror.Maskf(emptyDataError, "domain value in irsa cloudfront configmap for cluster must not be empty")
-						}
-					}
-
+					cloudfrontDomain = cm.Data["domain"]
```
njuettner, line 462: Not sure if this is right, in any case if the irsa-operator doesn't create the config map properly for handling the cloudfront part and the aws-operator is doing a lookup to get the values from non-existing keys, this will panic.

njuettner: The aws-operator could for instance reconcile the cluster before the irsa-operator notices the cluster, as an example.

whites11: fair enough

njuettner (approved):

PR #3441, opened 2022-11-15
Release v11.16.x efs (v17)

silenteh:
This PR addresses the issue that, when a customer wants to use the EFS csi driver app, additional permissions are needed in the ec2 IAM role used in the node pools, to connect to the AWS EFS Service.

The PR adds the additional policies, needed for the EFS csi driver app to create new EFS end points and mount them as volume in pods.

The additional permissions needed and added by this PR to the IAM role of the ec2 nodes are:

    EFS write CreateAccessPoints
    EFS write DeleteAccessPoints
    EFS list DescribeAccessPoints
    EFS list DescribeFileSystems
    EFS read DescribeMountTarget
    EC2 list DescribeAvailabilityZones

Unfortunately the EFS csi driver at the moment does not support IRSA, otherwise this could have potentially be solved in a different way.

## Checklist

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -2,10 +2,29 @@
 # Waiting for more direct dependencies to move away.
 CVE-2022-29153 until=2023-06-25
 CVE-2022-24687 until=2023-06-25
+CVE-2021-41803 until=2023-06-25
+CVE-2021-41803 until=2023-06-25
 
 # Pending github.com/kataras/iris release > 12.1.8
-CVE-2021-23772 until=2022-08-25
+CVE-2021-23772 until=2023-08-25
+
+### github.com/nats-io/nats-server
+# Stack-based Buffer Overflow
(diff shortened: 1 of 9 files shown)


njuettner:
There is a `release-major.minor.x` branch which is used when we want to patch some things for older releases.
I would recommend to merge this into the `release-10.16.x` branch. this will also create new release for you, see https://github.com/giantswarm/aws-operator/blob/master/.github/workflows/zz_generated.create_release_pr.yaml#L5-L24

njuettner:
So what you need to do is

1. Merge this PR into branch `release#v10.16.x`
2. Checkout the updated branch `release#v10.16.x` and verify you changes are present
3. Create a new branch from `release#v10.16.x`, called `release-v16.0.x#release#v16.0.1`.
4. This will create a new PR with a new release version 16.0.1 for the aws-operator.

whites11 (approved):

PR #3470, opened 2023-01-16
Add annotations to configure all disks in nodepools

paurosello:
## Goal of this PR
- Allow disk size configuration of logging volume. New default value is 15Gb.
- Allow different values for docker and containerd volume.

## Checklist

### DIFF
--- a/service/controller/key/machine_deployment.go
+++ b/service/controller/key/machine_deployment.go
@@ -20,6 +20,27 @@ func MachineDeploymentDockerVolumeSizeGB(cr infrastructurev1alpha3.AWSMachineDep
 	return strconv.Itoa(cr.Spec.NodePool.Machine.DockerVolumeSizeGB)
 }
 
+func MachineDeploymentContainerdVolumeSizeGB(cr infrastructurev1alpha3.AWSMachineDeployment) string {
+	result, ok := cr.ObjectMeta.Annotations[annotation.AWSContainerdVolumeSize]
+	if !ok {
+		//Default back to current value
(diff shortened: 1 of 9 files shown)


service/controller/key/machine_deployment.go
```diff
@@ -20,6 +20,27 @@ func MachineDeploymentDockerVolumeSizeGB(cr infrastructurev1alpha3.AWSMachineDep
 	return strconv.Itoa(cr.Spec.NodePool.Machine.DockerVolumeSizeGB)
 }
 
+func MachineDeploymentContainerdVolumeSizeGB(cr infrastructurev1alpha3.AWSMachineDeployment) string {
+	result, ok := cr.ObjectMeta.Annotations[annotation.AWSContainerdVolumeSize]
+	if !ok {
```
whites11, line 25: why don't we try to parse the value and default if it's invalid like we do for logging volume? This way we could try to create a volume that is "blah" gigs big just by setting a wrong annotation

paurosello:
Working on test cluster

Screenshot 2023-01-16 at 15 14 33

Screenshot 2023-01-16 at 15 13 31

Screenshot 2023-01-16 at 15 15 43

service/controller/key/machine_deployment.go
paurosello: We could try parsing anyway I guess and still return a string number... does that sound better?

paurosello: just did the change, can you check now?

whites11 (approved):
great job man

PR #3474, opened 2023-01-23
Remove IP Limit when prefix delegation is enabled

paurosello:
Set IP limit to 110 if prefix delegation is enabled on nodepools.

## Checklist


service/internal/cloudconfig/tcnp.go
```diff
@@ -184,6 +184,24 @@ func (t *TCNP) NewTemplates(ctx context.Context, obj interface{}) ([]string, err
 		externalSNAT = *key.ExternalSNAT(cl)
 	}
 
+	hasCilium, err := key.HasCilium(&cl)
```
whites11, line 187: you need to check the error here

service/internal/cloudconfig/tcnp.go
```diff
@@ -192,7 +210,6 @@ func (t *TCNP) NewTemplates(ctx context.Context, obj interface{}) ([]string, err
 
 		g8sConfig := cmaClusterToG8sConfig(t.config, cl, key.KubeletLabelsTCNP(&cr))
 
-		hasCilium, err := key.HasCilium(&cl)
 		if err != nil {
```
whites11, line 213: this check should be moved at line 188

whites11 (approved):

PR #3507, opened 2023-03-31
Set nftables env for aws-cni

njuettner:
For AWS Release v18.3.0 https://github.com/giantswarm/releases/pull/1100

This needs to be released with aws-operator 14.8.x

The aws-operator 14.9.x is already used in v19.0.0-alpha

## Checklist


njuettner:
Tests looking good, AFAIK

```
$ ks logs aws-node-khjs4 aws-node
time="2023-03-31T14:13:42Z" level=info msg="Updating iptables mode to nft"
Installed /host/opt/cni/bin/aws-cni
Installed /host/opt/cni/bin/egress-v4-cni
time="2023-03-31T14:13:42Z" level=info msg="Starting IPAM daemon... "
time="2023-03-31T14:13:42Z" level=info msg="Checking for IPAM connectivity... "
time="2023-03-31T14:13:43Z" level=info msg="Copying config file... "
time="2023-03-31T14:13:43Z" level=info msg="Successfully copied CNI plugin binary and config file."
```

whites11:
why are we doing this? We already have (or used to have) aws-cni releases featuring aws-cni using nftables.

See for example https://github.com/giantswarm/releases/blob/3f49a60beb5277b874d948f26053f244312b9b59/aws/v17.4.3/release.yaml#L56

njuettner:
With aws-cni release v1.12 and higher, upstream decided to support both cases nftables and iptables-legacy. They also updated their images so we cannot build on top anymore and we don't need to do that. 

I just started v18.3.0 release to fix the vulnerabilities in aws-cni, see https://github.com/giantswarm/releases/pull/1100. All tests are passing.

whites11 (approved):

PR #3521, opened 2023-04-18
Use `alpine` as image for aws-cni's `routes-fixer`.

whites11:
Since latest update of aws-cni, the image does not have the tools needed to run `routes-fixer` any more.
Thus we need to run alpine

## Checklist

### DIFF
--- a/service/controller/resource/prepareawscniformigration/create.go
+++ b/service/controller/resource/prepareawscniformigration/create.go
@@ -2,6 +2,7 @@ package prepareawscniformigration
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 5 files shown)


service/controller/resource/prepareawscniformigration/create.go
```diff
@@ -109,8 +110,8 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		r.logger.Debugf(ctx, "Daemonset %q doesn't have routes-fixer container", dsName)
 		ds.Spec.Template.Spec.Containers = append(ds.Spec.Template.Spec.Containers, corev1.Container{
 			Name:    "routes-fixer",
-			Image:   ds.Spec.Template.Spec.Containers[0].Image,
-			Command: []string{"/usr/bin/bash"},
+			Image:   fmt.Sprintf("%s/giantswarm/alpine:3.17.3", r.registryDomain),
```
njuettner, line 113: Not sure if we really want to hard set the image version of alpine here. 
Is there any other way to get the version somewhere?

whites11: would you rather have this as an helm chart value?

njuettner (approved):
Not blocking here, we will hopefully get rid off it anyway

service/controller/resource/prepareawscniformigration/create.go
njuettner: Do we need to bump it regularly, if so helm chart value might make sense?

njuettner: No worries, let's merge it.

PR #3522, opened 2023-04-18
Change route53 role Trust Policy

paurosello:
Allow multiple externalDNS app deployments to assume this role.

Towards: https://github.com/giantswarm/giantswarm/issues/26598

## Checklist

### DIFF
--- a/service/controller/resource/tccpn/template/template_main_iam_policies.go
+++ b/service/controller/resource/tccpn/template/template_main_iam_policies.go
@@ -384,7 +384,7 @@ const TemplateMainIAMPolicies = `
             Action: "sts:AssumeRoleWithWebIdentity"
             Condition:
               StringLike:
-                "{{ .IAMPolicies.CloudfrontDomain }}:sub": "system:serviceaccount:kube-system:*"
+                "{{ .IAMPolicies.CloudfrontDomain }}:sub": "system:serviceaccount:*:*"
           {{- end }}
           {{- if ne .IAMPolicies.CloudfrontAliasDomain "" }}
           - Effect: "Allow"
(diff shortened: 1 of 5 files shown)


service/controller/resource/tccpn/template/template_main_iam_policies.go
```diff
@@ -383,17 +383,17 @@ const TemplateMainIAMPolicies = `
               Federated: "arn:{{ .IAMPolicies.RegionARN }}:iam::{{ .IAMPolicies.AccountID }}:oidc-provider/{{ .IAMPolicies.CloudfrontDomain }}"
             Action: "sts:AssumeRoleWithWebIdentity"
             Condition:
-              StringEquals:
-                "{{ .IAMPolicies.CloudfrontDomain }}:sub": "system:serviceaccount:kube-system:external-dns"
+              StringLike:
+                "{{ .IAMPolicies.CloudfrontDomain }}:sub": "system:serviceaccount:kube-system:*"
```
njuettner, line 387: Do we want to also allow different namespaces? If so we also must replace kube-system with *

njuettner (approved):

service/controller/resource/tccpn/template/template_main_iam_policies.go
paurosello: I think leaving this in kube-system is the right approach... as it offer a bit of additional security.

@T-Kukawka also worth mentionting for release 19 upgrades

Will try to find out if customers install in other NS

paurosello: @whites11 as the SA depend on the app name, we can't do that :(

we can probably get away with external-dns* as the SA name on any namespace

paurosello: Screenshot 2023-04-18 at 11 33 43
customer example

PR #3540, opened 2023-04-28
Keep aws cni redux

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/2407

Before upgrading to v18, we have the aws-cni resources:

before-upgrade
before-upgrade-resources

During migration, an annotation `aws-operator.giantswarm.io/legacy-aws-cni-pod-cidr` is set and the aws cni resources are kept (except one NAT routing rule but that's fine)

after-upgrade-with-annotation
after-upgrade-with-annotation-resources

you can cleanup aws cni resources by manually removing the `aws-operator.giantswarm.io/legacy-aws-cni-pod-cidr` annotation

after-upgrade-without-annotation
after-upgrade-without-annotation-resources

## Checklist

### DIFF
--- a/service/controller/resource/tccp/create.go
+++ b/service/controller/resource/tccp/create.go
@@ -242,8 +242,14 @@ func (r *Resource) newParamsMain(ctx context.Context, cl apiv1beta1.Cluster, cr 
 			return nil, microerror.Mask(err)
 		}
 
+		enableAWSCni := key.IsAWSCNINeeded(cl)
+		// If we have the aws-operator.giantswarm.io/legacy-aws-cni-pod-cidr, we still need to keep AWS cni subnets around.
+		if key.LegacyAWSCniCIDRBlock(cr) != "" {
+			enableAWSCni = true
+		}
+
 		params = &template.ParamsMain{
-			EnableAWSCNI:    key.IsAWSCNINeeded(cl),
+			EnableAWSCNI:    enableAWSCni,
 			InternetGateway: internetGateway,
 			LoadBalancers:   loadBalancers,
 			NATGateway:      natGateway,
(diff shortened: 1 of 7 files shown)


service/controller/resource/tccp/create.go
```diff
@@ -242,8 +242,14 @@ func (r *Resource) newParamsMain(ctx context.Context, cl apiv1beta1.Cluster, cr
 			return nil, microerror.Mask(err)
 		}
 
+		enableAWSCni := key.IsAWSCNINeeded(cl)
+		// If we have the aws-operator.giantswarm.io/legacy-aws-cni-pod-cidr, we still need to keep AWS cni subnets around.
+		if key.LegacyAWSCniCIDRBlock(cr) != "" {
+			enableAWSCni = true
```
njuettner, line 248:
```suggestion
			keepAWSCni = true
```

service/controller/resource/tccp/create.go
```diff
@@ -242,8 +242,14 @@ func (r *Resource) newParamsMain(ctx context.Context, cl apiv1beta1.Cluster, cr
...
+		}
+
 		params = &template.ParamsMain{
-			EnableAWSCNI:    key.IsAWSCNINeeded(cl),
+			EnableAWSCNI:    enableAWSCni,
```
njuettner, line 252:
```suggestion
			KeepAWSCNI:    enableAWSCni,
```

service/controller/resource/tccp/create.go
njuettner: I'd rather change it to keep as mentioned in your comment

njuettner (approved):

service/controller/resource/tccp/create.go
whites11: this is not exactly true. The flag is actually used to enable aws CNI on v18, and incidentally reused to keep aws cni resources on v19. I think the name is correct (and it would require changing all cloudfront templates otherwise)

njuettner: Ok 👍🏻

PR #3564, opened 2023-06-14
Check if all nodes are rolled before deleting AWS CNI resources when …

whites11:
…upgrading from v18 to v19.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Check if all nodes are rolled before deleting AWS CNI resources when upgrading from v18 to v19.
+
 ## [14.17.1] - 2023-05-11
 
 ### Added
--- a/service/controller/resource/awscnicleaner/create.go
+++ b/service/controller/resource/awscnicleaner/create.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	operatorlabel "github.com/giantswarm/aws-operator/v14/pkg/label"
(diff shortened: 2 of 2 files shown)


service/controller/resource/restrictawsnodedaemonset/create.go
```diff
@@ -134,3 +114,39 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
...
+	found := false
+
+	for _, requirement := range requirements {
+		// Check if Key or Operator are not the ones we're looking for.
+		if requirement.Key != label.OperatorVersion || requirement.Operator != "NotIn" {
+			// This requirement is acting on a different label than the one we care, keep it as-is.
+			ret = append(ret, requirement)
+		} else {
+			// This requirement is using the same key and operator as the one we expected to set, so let's check if the value matches.
+			if len(requirement.Values) != 1 || requirement.Values[0] != project.Version() {
```
fiunchinho, line 130: would it make more sense to have `len(requirement.Values) > 0`?

whites11: that is not what we want. When the `key` and `operator` fields match, we want the value to be just one and with a specific value.
What this If is doing, is recognize when there is something off, for example:

- more than one value => `len(..)` != 1`
- wrong value

in those 2 cases we want to replace everything with our only desired value.
does it make sense?

fiunchinho: Got it, thanks for the explanation. Maybe worth adding this explanation as comment to the function?

whites11: :done:

fiunchinho (approved):

PR #3593, opened 2023-08-23
Allow specific flatcar release for machine pools

njuettner:
This PR allows customer to set an annotation on `AWSMachineDeployment` CR's to change the Flatcar Release Version.

For now it only allows to set `alpha.giantswarm.io/flatcar-release-version: "3689.0.0"` or a higher.

- when setting the annotation the TCNP Stack for the specific node pool is rolled and replaces the OS image

- when removing the annotation the node pool is updated and switches back to the default OS image which is coming from the AWS release 

- when upgrading the cluster to a new AWS release, the node pool uses the specific flatcar release from the annotation as long as you don't change by either setting it to a higher version or removing the annotation.

Tested:

- [x] gaia
- [x] giraffe

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow flatcar alpha releases for machine pools.
+
 ## [14.19.2] - 2023-08-03
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Allow flatcar alpha releases for machine pools.
+- Allow newer flatcar releases for node pools as provided by AWS release.
 
 ## [14.19.2] - 2023-08-03
 
--- a/service/controller/key/common.go
+++ b/service/controller/key/common.go
@@ -68,18 +68,18 @@ const sslOnlyBucketPolicyTemplate = `{
 }`
 
 // AMI returns the EC2 AMI for the configured region and given version.
-func AMI(region string, release releasev1alpha1.Release, flatcarAlphaReleaseVersion string) (string, error) {
+func AMI(region string, release releasev1alpha1.Release, flatcarReleaseVersion string) (string, error) {
 
 	var osVersion string
 	var err error
 
-	if flatcarAlphaReleaseVersion == "" {
+	if flatcarReleaseVersion == "" {
 		osVersion, err = OSVersion(release)
 		if err != nil {
 			return "", microerror.Mask(err)
 		}
 	} else {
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow flatcar alpha releases for machine pools.
```
whites11, line 12: generically, with this change we allow to change flatcar to any (newer) release, not just alpha. Shall we rephrase the changelog to explain that?

service/controller/key/common.go
```diff
@@ -188,6 +196,11 @@ func IsV19Release(releaseVersion *semver.Version) bool {
 	return releaseVersion.Major >= v19.Major
 }
 
+func IsMinimumFlatcarVersion(flatcarVersion *semver.Version) bool {
+	minRelease, _ := semver.New("3689.0.0")
```
whites11, line 200: if we hardcode this, it means we have to remember to bump it every time we release aws operator in the future.
How about taking the value from the release CR and only allowing anything newer than that?

service/controller/key/zz_generated.ami.go
```diff
@@ -1166,7 +1166,32 @@ var amiJSON = []byte(`{
     "us-east-2": "ami-033f192100c9980a1",
     "us-west-1": "ami-07a8a027c32ef1115",
     "us-west-2": "ami-08d7c3221e6f57e1f"
-   }
+   },
+  "3689.0.0": {
```
whites11, line 1170: this is a problem if you ask me. Why having an annotation to be able to use any flatcar version if we anyway need to add any new version into this file and make a new release?
I believe if we want the annotation in place we should get rid of this file and get AMI ID discovery somehow automated or at least prevent the need of a new release

service/controller/key/common.go
njuettner: That make sense, I changed it 👍🏻

CHANGELOG.md
njuettner: Changed it slightly 👍🏻

service/controller/key/zz_generated.ami.go
njuettner: I agree, AMI ID discovery might be a good solution. 

It might be a bit more challenging in China since we're using private AMI's.

We might need to also touch the permissions for `aws-operator again, IIRC we don't have enough permissions right now to do the discovery.

Do you want me to add an issue for this idea?

whites11: at least for tracking purposes, yes please

njuettner: ok will do

whites11 (approved):

service/controller/key/zz_generated.ami.go
njuettner: Issue: https://github.com/giantswarm/giantswarm/issues/27990

PR #3645, opened 2023-11-09
Pss

whites11:
## Checklist


paurosello (approved):
LGTM

whites11:
> Don't think this will actually remove the restricted PSP, only it will not be created on new clusters, is that the expectation?

Pretty much,yes. When on 1.25 those resources will disappear anyway so pointless to remove them imho

PR #3718, opened 2024-02-29
Handle karpenter nodes in node-termination-handler.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/30092

node-problem-detector based node termination for disk full does not work with karpenter nodes.
this PR fixes that

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [15.0.1] - 2024-01-31
-
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,11 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [15.0.1] - 2024-01-31
```
whites11, line 10: 15.0.1 is a patch release to v15 released after 16.0.0 was out. It shouldn't have been in the changelog to begin with

T-Kukawka (approved):

